PR #1, opened 2015-05-21
Update argument parser and two tests to use utils.getPackageDir

jhoblitt:


jhoblitt:
Fixed up and repushed.

PR #3, opened 2015-09-03
Replace --trace with --logdest in ArgumentParser

PaulPrice:
Enhanced --logdest to support name=value arguments (like --trace)
and eliminated --trace.
Updated the unit test to test this modified functionality.

### DIFF
--- a/python/lsst/pipe/base/argumentParser.py
+++ b/python/lsst/pipe/base/argumentParser.py
@@ -217,7 +217,7 @@ class ArgumentParser(argparse.ArgumentParser):
             usage = usage,
             fromfile_prefix_chars = '@',
             epilog = """Notes:
-* --config, --configfile, --id, --trace and @file may appear multiple times;
+* --config, --configfile, --id, --loglevel and @file may appear multiple times;
     all values are used, in order left to right
 * @file reads command-line options from the specified file:
     * data may be distributed among multiple lines (e.g. one option per line)
@@ -239,9 +239,10 @@ class ArgumentParser(argparse.ArgumentParser):
             help="config override(s), e.g. -c foo=newfoo bar.baz=3", metavar="NAME=VALUE")
         self.add_argument("-C", "--configfile", dest="configfile", nargs="*", action=ConfigFileAction,
             help="config override file(s)")
-        self.add_argument("-L", "--loglevel", help="logging level")
-        self.add_argument("-T", "--trace", nargs="*", action=TraceLevelAction,
-            help="trace level for component", metavar="COMPONENT=LEVEL")
+        self.add_argument("-L", "--loglevel", nargs="*", action=LogLevelAction,
+            help="logging level; supported levels are [debug|warn|info|fatal] or an integer; "
+                "trace level is negative log level, e.g. use level -3 for trace level 3",
+            metavar="LEVEL|COMPONENT=LEVEL")
         self.add_argument("--debug", action="store_true", help="enable debugging output?")
         self.add_argument("--doraise", action="store_true",
             help="raise an exception on error (else log a message and continue)?")
@@ -328,8 +329,8 @@ class ArgumentParser(argparse.ArgumentParser):
           the value of which is a DataIdArgument that includes public elements 'idList' and 'refList'
         - log: a pex_logging log
         - an entry for each command-line argument, with the following exceptions:
-          - config is Config, not an override
-          - configfile, id, logdest, loglevel are all missing
+          - config is the supplied config, suitably updated
+          - configfile, id and loglevel are all missing
         - obsPkg: name of obs_ package for this camera
         """
         if args == None:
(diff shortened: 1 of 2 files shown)


python/lsst/pipe/base/argumentParser.py
```diff
@@ -239,9 +239,10 @@ def __init__(self, name, usage = "%(prog)s input [options]", **kwargs):
             help="config override(s), e.g. -c foo=newfoo bar.baz=3", metavar="NAME=VALUE")
         self.add_argument("-C", "--configfile", dest="configfile", nargs="*", action=ConfigFileAction,
             help="config override file(s)")
-        self.add_argument("-L", "--loglevel", help="logging level")
-        self.add_argument("-T", "--trace", nargs="*", action=TraceLevelAction,
-            help="trace level for component", metavar="COMPONENT=LEVEL")
+        self.add_argument("-L", "--loglevel", nargs="*", action=LogLevelAction,
+            help="logging level for component; supported levels are [debug|warn|info|fatal] or an integer; "
+                "use -n for trace level n",
+            metavar="LEVEL|COMPONENT=LEVEL")
```
PaulPrice: It was not immediately obvious to me that the reference to `-n` meant `--loglevel -n` --- it looks like another argument name.

python/lsst/pipe/base/argumentParser.py
```diff
@@ -732,19 +724,28 @@ def __call__(self, parser, namespace, values, option_string):
         @param[in] parser       argument parser (instance of ArgumentParser)
         @param[in] namespace    parsed command (an instance of argparse.Namespace); ignored
         @param[in] values       a list of trace levels;
-            each item must be of the form component_name=level
+            each item must be of the form component_name=level or level
```
PaulPrice: Put quotes around the form, to distinguish them.  And I think it would help to put 'level' first.

    each item must be of the form 'level' or 'component_name=level'

python/lsst/pipe/base/argumentParser.py
```diff
@@ -732,19 +724,28 @@ def __call__(self, parser, namespace, values, option_string):
...
+            where level is an integer or a keyword
         @param[in] option_string    option value specified by the user (a str)
         """
+        permittedLevelList = ('DEBUG', 'INFO', 'WARN', 'FATAL')
```
PaulPrice: Don't need `permittedLevelList`.  If you don't want the below error to say "set(...)", then cast it to a list then.

python/lsst/pipe/base/argumentParser.py
```diff
@@ -732,19 +724,28 @@ def __call__(self, parser, namespace, values, option_string):
...
-            pexLog.Trace.setVerbosity(component, level)
-
+                levelStr, component = component, ""
+            logLevelUpr = levelStr.upper()
+            if logLevelUpr in permittedLevelSet:
+                logLevel = getattr(namespace.log, logLevelUpr)
+            else:
+                try:
+                    logLevel = int(levelStr)
+                except Exception:
```
PaulPrice: except ValueError

And include the exception message in the error?

python/lsst/pipe/base/argumentParser.py
```diff
@@ -732,19 +724,28 @@ def __call__(self, parser, namespace, values, option_string):
...
             component, sep, levelStr = componentLevel.partition("=")
             if not levelStr:
-                parser.error("%s level %s must be in form component=level" % (option_string, componentLevel))
-            try:
-                level = int(levelStr)
-            except Exception:
-                parser.error("cannot parse %r as an integer level for %s" % (levelStr, component))
-            pexLog.Trace.setVerbosity(component, level)
-
+                levelStr, component = component, ""
```
PaulPrice: Setting `component = None` might be a bit clearer than `""`.

tests/testArgumentParser.py
```diff
@@ -268,18 +268,26 @@ def testLogLevel(self):
...
                 args = [DataPath, "--loglevel", logLevel],
             )
             self.assertEqual(namespace.log.getThreshold(), intLevel)
+            self.assertFalse(hasattr(namespace, "loglevel"))
 
             namespace = self.ap.parse_args(
                 config = self.config,
-                args = [DataPath, "--loglevel", str(intLevel)],
+                args = [DataPath, "--loglevel", str(intLevel),
+                    "foo.bar=%s" % (str(intLevel + 27),),
```
PaulPrice: Don't write magic numbers (27, 5000) in more than one place.

python/lsst/pipe/base/argumentParser.py
r-owen: Casting a set to a list loses its order, and the order is useful in the error message. I agree it is annoying to save two such similar variable, but I didn't think of a simple, fast alternative. Two other solutiosn I did think of:
- Cast to a set for each test, or search the list instead of the set. Neither of these slows parsing very much, but it irks me to use linear search.
-  Save the info as a dict of name: severity, then sort by severity when printing the message. In my opinion this adds too much code.

python/lsst/pipe/base/argumentParser.py
r-owen: I do not see the value of including the error message. In my opinion the reported error says all that the user needs to know, and I worry about adding needless clutter. What am I missing?

python/lsst/pipe/base/argumentParser.py
r-owen: Good suggestion. I'll do it.

tests/testArgumentParser.py
r-owen: OK. And I should try a negative value, as well.

python/lsst/pipe/base/argumentParser.py
r-owen: Good point.

PR #4, opened 2015-10-08
Review for DM-3811(HSC backport: Include documentation strings for config parameters when they are dumped)

jdswinbank:


tests/testArgumentParser.py
```diff
@@ -196,12 +210,28 @@ def testShow(self):
...
+        # Test show with glob for multi-line doc string
+        with capture() as out:
+            self.ap.parse_args(self.config, [DataPath, "--show", "config=*choiceItem", "run"])
+        stdout = out[0]
+        stdoutList = stdout.rstrip().split("\n")
+        self.assertGreater(len(stdoutList), 2) # at least 2 docstring lines (1st line is always a \n)
+                                               # and 1 config parameter
+        answer = [ans for ans in stdoutList if not ans.startswith("#")] # get rid of comment lines
+        self.assertEqual(len(answer), 2)       # only 1 newline plus 1 config item matches
+        self.assertIn("ONE", answer[1])
```
jdswinbank: You could refactor the above to reduce quite a lot of duplication (there's a repeating `with capture; assertion; list comprehension; assertion; assertion` pattern, which smells a bit dubious).

tests/testArgumentParser.py
```diff
@@ -64,7 +64,9 @@ class SampleConfig(pexConfig.Config):
     floatItem = pexConfig.Field(doc="sample float field", dtype=float, default=3.1)
     strItem = pexConfig.Field(doc="sample str field", dtype=str, default="strDefault")
     subItem = pexConfig.ConfigField(doc="sample subfield", dtype=SubConfig)
-    
+    choiceItem = pexConfig.ChoiceField(doc = "sample ChoiceField (# multiline doc)", dtype = str,
```
jdswinbank: Momentary confusion as this isn't a multi-line string. Of course, it is when the allowed values are appended. Is it worth making it explicitly multi-line (ie, include a `\n`) just to drive home the point (and stop me from double-taking next time...)?

python/lsst/pipe/base/argumentParser.py
```diff
@@ -564,8 +564,10 @@ def __init__(self, pattern):
 
                     def write(self, str):
                         str = str.rstrip()
-                        if str and fnmatch.fnmatch(str, self._pattern):
-                            print str
+                        # Strip off doc string line(s) and cut off at "=" for string matching
+                        matchStr = str.split("\n")[-1].split("=")[0]
+                        if fnmatch.fnmatch(matchStr, self._pattern):
+                            print "\n" + str
```
jdswinbank: Print as function?

python/lsst/pipe/base/argumentParser.py
```diff
@@ -564,8 +564,10 @@ def __init__(self, pattern):
 
                     def write(self, str):
                         str = str.rstrip()
-                        if str and fnmatch.fnmatch(str, self._pattern):
-                            print str
+                        # Strip off doc string line(s) and cut off at "=" for string matching
+                        matchStr = str.split("\n")[-1].split("=")[0]
```
jdswinbank: You could make this even shorter by including the `rstrip` directly -- `matchStr = str.rstrip().split(...)`, etc. Whether that's gaining or losing you readability is arguable, I suppose.

jdswinbank:
Could you please update the commit message for 004a5d3 to refer to the specific commit on `pex_config` which necessitated the change to the tests here?

python/lsst/pipe/base/argumentParser.py
laurenam: But I don't think that will leave `str` modified (i.e rstrip-ed), and I believe that is still desirable for the printing of `str`, so I think I will leave this one as is.

PaulPrice: Not your fault, but we shouldn't be using `str` as a variable name --- it clobbers the class name.

PaulPrice: Not your fault, but we shouldn't be using `str` as a variable name --- it clobbers the class name.

laurenam: Thanks...remedied.

PR #6, opened 2015-12-16
Initial prototype of super task infrastructures -- tickets/DM-4509

timj:
This integrates the prototype super task into pipe_base. There should be no functional change to pipe_base itself as this code only adds a single attribute to `task.py`.

We are proposing to integrate this before it is completely finished in order to ease migration and testing without having to work on branches of branches.


TallJimbo:
Would this be a good time to consider a name other than `SuperTask`?  While it clearly has some emotional baggage for some people, I think we're big enough to consider reusing `Stage`, for instance.

frossie:
I don't mind another name but folks would have to propose it and if we're going to pick a name from scratch I would vote at least picking something that is reasonably standard in off-the-shelf workflows. Taverna uses Components, but really most of them call them Tasks. I will say "stage" seems confusing to me, as in the verb context it means something else.

mgckind:
I agree with Frossie, I don't mind changing the name., but "stage" might be confusing. Note though, that SuperTasks are indeed  a subclass of Task with some extra additions, but the core is all in Task. Whether we want to keep the suffix Task or not is also open to discussion (like ComponentTask). Same goes for the Activator and the CmdLineActivator

python/lsst/pipe/base/activator.py
```diff
@@ -0,0 +1,449 @@
+"""
+activator: This module contains the activator class which are used
+to execute (Super)Task in different context/environments
```
r-owen: Please fix the grammar "activator class which are...". I suspect you want "activator classes".

python/lsst/pipe/base/activator.py
```diff
@@ -0,0 +1,449 @@
...
+@contextlib.contextmanager
+def profile(filename, log=None):
+    """!Context manager for profiling with cProfile
+
+    @param filename     filename to which to write profile (profiling disabled if None or empty)
+    @param log          log object for logging the profile operations
+
+    If profiling is enabled, the context manager returns the cProfile.Profile object (otherwise
+    it returns None), which allows additional control over profiling.  You can obtain this using
+    the "as" clause, e.g.:
```
r-owen: I don't see any test for "if profiling not enabled then return None". Is this accurate? (And why return None instead of a disabled Profile()?)

python/lsst/pipe/base/activator.py
```diff
@@ -0,0 +1,449 @@
...
+    profile.enable()
+    yield profile
+    profile.disable()
+    profile.dump_stats(filename)
+    if log is not None:
+        log.info("cProfile stats written to %s" % filename)
+
+
+class TaskNameError(Exception):
+    """
```
r-owen: If you don't want a TaskNameError to cause an exception the consider inheriting from TaskError

I'm skeptical that a new specialized exception is required, but from your comments, so are you.

python/lsst/pipe/base/activator.py
```diff
@@ -0,0 +1,449 @@
...
+class Activator(object):
+    """
+    Hook for other activators
+    """
+
+    def __init__(self):
+        """
+        This might grow into a common initialization, right now
+        is overwritten
+        """
```
r-owen: overridden instead of overwritten?

python/lsst/pipe/base/activator.py
```diff
@@ -0,0 +1,449 @@
...
+        This is the required method to execute the activator
+        """
+        pass
+
+
+class CmdLineActivator(Activator):
+    """
+    Activator for the command line: This Class takes the higher level SuperTask as input and
+    executes in the command line, similarly to CmdLineTask but this is a generic way to run (
```
r-owen: Please start with a brief summary, then provide more info.

python/lsst/pipe/base/activator.py
```diff
@@ -0,0 +1,449 @@
...
+        It creates another instance of the SuperTask passed to the activator, it returns the
+        SuperTask instance. Similar to MakeTask in CmdlineTask
+        """
+        return self.SuperTask.__class__(config=self.config, log=self.log, activator='cmdLine')
+
+    def precall(self):
+        """
+        A hook similar to the CmdLineTask one, is True by default and need to copy it from it,
+        which is TBD.
+
```
r-owen: I have no idea what that means. I think it returns True by default (rather than "is True") but the rest of it is opaque to me.

python/lsst/pipe/base/activator.py
```diff
@@ -0,0 +1,449 @@
...
+        :param super_taskname: The name of the SuperTask
+        :return: the Task and Config instances of the supertask
+        """
+        class_task_instance = None
+        class_config_instance = None
+        super_task_module = None
+
+        for _, package in TASK_PACKAGES.iteritems():  # Loop over all packages
+            for _, module_name, _ in pkgutil.iter_modules(package.__path__):  # Loop over modules
```
r-owen: If you are going to discard the keys then use `for package in TASK_PACKAGES.itervalues():`

python/lsst/pipe/base/activator.py
mgckind: Fixed, thanks

gpdf:
I'm not, um, super happy with "SuperTask" but I agree with Matias that the inclusion of "Task" in the name is a useful reminder of the design, in which SuperTask is indeed a type of (derived from) Task.

There's some content here beyond pure naming:

I've considered an alternate design in which the new behaviors of SuperTask - in particular the execute() method - come from a new class independent of Task (which one could then imagine naming "Stage").  The SuperTask concept would then be that of a class which implements the interfaces of both a Stage and a Task, and usable in either way (via run() or via activate()).  That has some attractions, but it loses the other point, which is that the class-that-has-execute() ("Stage") also participates in the configuration mechanism.

That could be addressed via a "diamond inheritance" approach: Task inherits from ConfigurableComponent, Stage inherits from ConfigurableComponent, SuperTask inherits from both.  (In C++ one would want that to be virtual inheritance so that there is only one instance of the ConfigurableComponent in a SuperTask.)  But this feels unPythonic (especially because ConfigurableComponent is not just an interface) and contrived.

Note that it's specifically SuperTask - i.e., a class that has both execute() and run() - that is meant to be the successor to CmdLineTask.

python/lsst/pipe/base/activator.py
mgckind: I copied this directly from CmdLineTask.py, same is true for other functions. Didn't want to modified to mimic the CmdLineTask closely.

python/lsst/pipe/base/activator.py
mgckind: yes, I think  we might not need this after all. I though about creating a new Exception to deal with name conflicts in general (that differs from NameError). I'll add a note to keep  this in mind

python/lsst/pipe/base/activator.py
mgckind: yes, thanks!

python/lsst/pipe/base/activator.py
mgckind: Noted, thanks

python/lsst/pipe/base/activator.py
mgckind: I had issues when I copied some of these methods  from CmdLineTask.py, so I just left this as a hook as I need to bring more methods back to the activator. I'll add a note on this as it has not been fully imported

python/lsst/pipe/base/activator.py
mgckind: Changed, thanks for the suggestion

timj:
We are closing this PR as it has been noted that the code is too experimental to be released into the approved code base. We will therefore begin work on a "supertask" branch and merge our tickets on to there until we are ready for external review.

PR #11, opened 2016-03-11
DM-4236

r-owen:
This prevents unintentionally writing in the top-level data repo
by leaving off "--output" or "--rerun" or $PIPE_OUTPUT_ROOT. If
none of these is set, we error out. This behavior can be disabled
by setting ArgumentParser.requireOutput = False.

In the process, cleaned up the rerun handling, so that the actual
rerun name is stored in namespace.rerun instead of a directory.

Added a test demonstrating various ways of specifying the output
directory.

### DIFF
--- a/python/lsst/pipe/base/argumentParser.py
+++ b/python/lsst/pipe/base/argumentParser.py
@@ -461,6 +461,12 @@ class ArgumentParser(argparse.ArgumentParser):
 
         obeyShowArgument(namespace.show, namespace.config, exit=False)
 
+        # No environment variable or --output or --rerun specified.
+        if self.requireOutput and namespace.output is None and namespace.rerun is None:
+            self.error("no output directory specified.\n"
+                       "An output directory must be specified with the --output or --rerun\n"
+                       "command-line arguments.\n")
+
         namespace.butler = dafPersist.Butler(
             root = namespace.input,
             calibRoot = namespace.calib,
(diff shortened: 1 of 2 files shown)


python/lsst/pipe/base/argumentParser.py
```diff
@@ -494,6 +496,12 @@ def parse_args(self, config, args=None, log=None, override=None):
         namespace.config.validate()
         namespace.config.freeze()
 
+        # No environment variable or --output or --rerun specified.
+        if self.requireOutput and namespace.output is None and namespace.rerun is None:
+            self.error("no output directory specified.\n"
+                       "An output directory must be specified with the --output or --rerun\n"
+                       "command-line arguments or the PIPE_OUTPUT_ROOT environment variable.\n")
+
```
r-owen: I suggest removing the bit about PIPE_OUTPUT_ROOT (but see associated request near line 515)

python/lsst/pipe/base/argumentParser.py
```diff
@@ -513,8 +520,6 @@ def _parseDirectories(self, namespace):
         # be run as there may be an environment variable set for the output
         else:
             namespace.output = _fixPath(DEFAULT_OUTPUT_NAME, namespace.rawOutput)
```
r-owen: I suggest removing this else clause. $PIPE_OUTPUT_ROOT was always intended to be a parent dir, never the actual output dir. This bug or misfeature has been present for a long time, but it should be squashed.

python/lsst/pipe/base/argumentParser.py
```diff
@@ -280,6 +280,8 @@ class ArgumentParser(argparse.ArgumentParser):
       before I do this checking. Constructing a butler is slow, so I only want do it once,
       after parsing the command line, so as to catch syntax errors quickly.
     """
+    requireOutput = True  # Require an output directory to be specified?
+
```
r-owen: Could you please say a word or two about why this is a class variable instead of a constructor argument? I would have expected the latter.

PR #14, opened 2016-05-06
Fixes for NFS

timj:


tests/testArgumentParser.py
```diff
@@ -289,13 +290,15 @@ def testAtFile(self):
...
-        )
-        self.assertTrue(os.path.isfile(logFile))
-        os.remove(logFile)
+        logFile = tempfile.NamedTemporaryFile()
+        try:
+            self.ap.parse_args(
+                config = self.config,
+                args = [DataPath, "--logdest", logFile.name],
+            )
+            self.assertTrue(os.path.isfile(logFile.name))
```
timj: I think the problem here is that this test would pass even without calling the previous line so the test isn't testing that the logging is working properly.

tests/testArgumentParser.py
```diff
@@ -289,13 +290,15 @@ def testAtFile(self):
     def testLogDest(self):
         """Test --logdest
         """
-        logFile = "tests_argumentParser_testLog_temp.txt"
-        self.ap.parse_args(
-            config = self.config,
-            args = [DataPath, "--logdest", logFile],
-        )
-        self.assertTrue(os.path.isfile(logFile))
-        os.remove(logFile)
```
timj: Thanks for finding the issue. This is removing the log file but not closing the file, hence the `.nfs` file still sitting around. I'm not entirely sure why this is not seen as a leak in the non-NFS runs. It might be that this is a file descriptor leak but `psutil` does not see it because it is not associated with a file name any more. On NFS it is an actual file so turns up as a leak.

timj: I can confirm that `psutil.Process().open_files()` does not return the log file but `/usr/sbin/lsof` does list the log file as being open still.

timj: If I change the `utils` leak checker to use `lsof` then this reports a leak as expected. I'm not entirely sure how to proceed. The new patch is not correct because it would work even if the logger failed to do anything. The old code really does leak a file descriptor and the leak checker in `psutil` is not really helping. Do I migrate `utils` to use `lsof` in preference to `psutil`?

tests/testArgumentParser.py
timj: `parse_args` does return the logger object so in theory we could ask that object to free up the resource but I don't think there is the opposite of the `pex_logging` `addDestination` method.

hchiang2: Thanks for catching this!

tests/testArgumentParser.py
```diff
@@ -289,13 +290,17 @@ def testAtFile(self):
...
+        logFile = tempfile.NamedTemporaryFile()
+        try:
+            namespace = self.ap.parse_args(
+                config = self.config,
+                args = [DataPath, "--logdest", logFile.name],
+            )
+            namespace.log.info('test logging to the specified file')
+            self.assertTrue(os.path.isfile(logFile.name))
+            self.assertGreater(os.path.getsize(logFile.name), 0)
```
timj: This is a better test since it does at least try to show that logging is also working. The problem is that if I update the leak detector as described above to use `lsof` instead of `psutil` then this still does leak a file descriptor. I think this is because the logger is global and I think that is correct because the logger will open the log file and this test code has a file descriptor open on the log file; the test file closes the log file which causes it to be unlinked but the logger still has an open file descriptor (which I think you will be able to demonstrate by writing to the log after the `finally` although I don't know if the logger raises an error if it fails to write the log message). @ktlim any thoughts? Do we use `psutil` on the basis that it's "mostly right"? Do I switch to lsof as it's less portable but seemingly more accurate? Do we fix the logger to allow log destination to be removed?

tests/testArgumentParser.py
hchiang2: I just pushed a corrected version that actually test something.  Running Jenkins with this patch and without the utils patch now.

tests/testArgumentParser.py
hchiang2: As I understand, this feature with logdest may be removed when we migrate to use lsst::log. 
Is that something to consider?

hchiang2: Just curious, would some tests in pex_logging also leak file descriptors if utils use `lsof` instead of `psutil`?

timj: Won't make a difference because `pex_logging` doesn't seem to use the memory tester so it's not being checked at all. It only uses `unittest` in 2 out of 9 test files.

hchiang2:
Superseded by PR#13 with the same content.

PR #16, opened 2016-06-06
Tickets/dm 6074

r-owen:


doc/main.dox
```diff
@@ -172,13 +172,27 @@ is to use a camera-specific variant of a subtask. Examples include:
 - lsst.obs.sdss.selectSdssImages.SelectSdssImagesTask: an version of the task that selects images
     for co-addition of SDSS stripe 82 images
 
-Here is an example of retargeting a subtask in a config override file; this retargets
-lsst.pipe.tasks.exampleCmdLineTask.ExampleCmdLineTask `stats` with a simpler version:
+How you retarget a subtask depends if the subtask is specified as an
```
pschella: depends on whether?

as a?

doc/main.dox
```diff
@@ -172,13 +172,27 @@ is to use a camera-specific variant of a subtask. Examples include:
...
+\ref lsst.pex.config.configurableField.ConfigurableField "lsst.pex.config.ConfigurableField"
+(the most common case, and the source of the term "retarget") or as an
```
pschella: as a?

doc/main.dox
```diff
@@ -172,13 +172,27 @@ is to use a camera-specific variant of a subtask. Examples include:
...
+If a subtask is specified as an
+\ref lsst.pex.config.registry.RegistryField "lsst.pex.config.RegistryField"
+then the subtask is retargeted by setting the `name` property of the field
+to the name of the desired subtask, for example:
+
+<pre>
+    import ...  # the package in which subtask "bar" is defined, so it can register itself
+    config.aRegistrySubtask.name = "bar"
+</pre>
+
```
pschella: I think it is clearer if you use the exact same example as above with the same names and modify that to the "RegistryField" case.

python/lsst/pipe/base/struct.py
```diff
@@ -1,30 +1,30 @@
 from __future__ import absolute_import, division
-# 
+#
 # LSST Data Management System
 # Copyright 2008, 2009, 2010, 2011 LSST Corporation.
```
pschella: Adjust copyright to LSST template?

tests/testCmdLineTask.py
```diff
@@ -150,7 +150,8 @@ def testNoBackupConfig(self):
         TestTask.parseAndRun(args=[DataPath, "--output", self.outPath, "--id", "visit=3", "filter=r",
                                    "--config", "floatField=-99.9", "--clobber-config", "--no-backup-config"])
         # Ensure backup config file was NOT created
-        self.assertFalse(os.path.exists(os.path.join(self.outPath, "config", TestTask._DefaultName + ".py~1")))
+        self.assertFalse(
+            os.path.exists(os.path.join(self.outPath, "config",TestTask._DefaultName + ".py~1")))
```
pschella: Are we sure TestTask._DefaultName never has trailing whitespace?

tests/testTask.py
```diff
@@ -114,12 +119,14 @@ def testBasics(self):
             for multiplicand in (0.9, -45.0):
                 config = AddMultTask.ConfigClass()
                 config.add.addend = addend
-                config.mult.multiplicand = multiplicand
+                config.mult["stdMult"].multiplicand = multiplicand
```
pschella: Wow, this is confusing syntax. But probably can't be helped.

doc/main.dox
r-owen: I will chnage "if" to whether".

I believe "an" is correct before "LSST" since the L is pronounced "ell".

tests/testCmdLineTask.py
r-owen: For this test? Certainly. In general, we don't enforce it, but it should usually be so.

doc/main.dox
r-owen: If by "as above" you mean the same as used in the "how to write a task" documentation then I agree and will update it accordingly. I will also include an example of overriding configuration parameters.

PR #17, opened 2016-06-14
Add documentation for the --rerun option

r-owen:
The --rerun option brought over from HSC was lacking good
documentation. This ticket provides said documentation by sumamrizing
the relevent contents of RFC-95: Filesystem Organization & Argument
Parser Features.

### DIFF
--- a/doc/main.dox
+++ b/doc/main.dox
@@ -119,6 +119,44 @@ Thus the following example specifies all sensors for four combinations of visit 
 plus all sensors for one raft of two other visits for calibrated lsstSim data:
 - `--id visit=54123^55523 raft=1,1^2,1 --id visit=623459^293423 raft=0,0`
 
+\subsection pipeBase_argumentParser_rerun The --rerun Option
+
+The `--rerun` option is a convenient shortcut for specifying the output directory of a command line task and
+serves the same purpose as the `--output` argument. However, the value supplied to `--rerun` is relative to
+the "rerun" directory of the root input data repository (i.e. follow the "_parent" chain all the way back,
+then add "rerun" to that). `--rerun` saves the user from typing in the input repository twice. For example:
+
+`processCcd.py $root/hsc --output $root/hsc/rerun/vpk24/DM-4443`
+
+which will run processesCcd on the (HSC) data in "$root/hsc" and put the output in 
+"$root/hsc/rerun/vpk24/DM-4443" can be shortened by using `--rerun` to
+
+`processCcd.py $root/hsc ----rerun vpk24/DM-4443`
+
+This is useful because outputs from the pipeline usually go within the camera-level root repository into a
+subdirectory of a "rerun" directory present in the camera-level root repository. These output sub-directories
+should usually be named using JIRA issue numbers and dates, again with a distinction between public and
+private reruns. For example:
+
+`$root/hsc/rerun/DM-3456_20150607`   # a public rerun useful for many people
+
+`$root/hsc/rerun/jbosch/DM-7890/fixed-psf`   # one of several private reruns used to debug a single issue
+
+`$root/lsstSim/krughoff/DM-5678/rerun/jbosch/DM-5679`   # simulated by Simon, processed by Jim
+
+As with any other output repository, reruns can be chained, but keep in mind that the name of the parent rerun
+is not generally recorded in the name of the child. Chaining is made easy by supporting a two-value 
+form separated by a colon, which specifies both a parent rerun and the new output directory. For example:
+
+`processCcd.py $root/hsc --rerun price/DM-5680 ...`   # Paul processes some CCDs
+
+`makeCoaddTempExp.py $root/hsc --rerun price/DM-5680:jbosch/DM-5681 ...`   # Jim uses Paul's CCD processing to
+start a coadd.
+
+The `--rerun` can also specify a path, part of which can be a symbolic link. For example, if 
+"$root/hsc/rerun/rhl" -> "/other/disk/rhl/rerun", then `--rerun rhl/tmp/foo` creates "/other/disk/rhl/tmp/foo"
+where note that if the path is absolute, it is taken to be the path.
+
 \subsection pipeBase_argumentParser_argumentFiles Argument Files
 
 You may specify long or repetitive command-line arguments in text files and reference those files


doc/main.dox
```diff
@@ -119,6 +119,43 @@ Thus the following example specifies all sensors for four combinations of visit
...
+is not generally recorded in the name of the child. Chaining is made easy by supporting a two-value 
+form separated by a colon, which specifies both a parent rerun and the new output directory. For example:
+
+`processCcd.py $root/hsc --rerun price/DM-5680 ...`   # Paul processes some CCDs
+
+`makeCoaddTempExp.py $root/hsc --rerun price/DM-5680:jbosch/DM-5681 ...`   # Jim uses Paul's CCD processing to
+start a coadd.
+
+The `--rerun` can also specify a path, e.g. `--rerun rhl/tmp/foo` creates "$root/hsc/rerun/rhl/tmp/foo". A
+part of this path can be a symbolic link (e.g. "$root/hsc/rerun/rhl" -> "/other/disk/rhl/rerun").
```
r-owen: This is unclear. I assume you mean that if part of the specified path exists as a symlink then all will work as expected. The text is probably fine as is, but I suggest you make the example more explicit. e.g. with the symlink you gave --rerun rhl/tmp/foo creates /other/disk/rhl/tmp/foo (if I got that right).

doc/main.dox
```diff
@@ -119,6 +119,43 @@ Thus the following example specifies all sensors for four combinations of visit
...
+As with any other output repository, reruns can be chained, but keep in mind that the name of the parent rerun
+is not generally recorded in the name of the child. Chaining is made easy by supporting a two-value 
+form separated by a colon, which specifies both a parent rerun and the new output directory. For example:
+
+`processCcd.py $root/hsc --rerun price/DM-5680 ...`   # Paul processes some CCDs
+
+`makeCoaddTempExp.py $root/hsc --rerun price/DM-5680:jbosch/DM-5681 ...`   # Jim uses Paul's CCD processing to
+start a coadd.
+
+The `--rerun` can also specify a path, e.g. `--rerun rhl/tmp/foo` creates "$root/hsc/rerun/rhl/tmp/foo". A
```
r-owen: Please reword "The --rerun can also specify a path". How about "The value of --rerun can be a relative path". (Which raises a question: what happens if it is an absolute path?).

doc/main.dox
```diff
@@ -119,6 +119,43 @@ Thus the following example specifies all sensors for four combinations of visit
...
+then add "rerun" to that). `--rerun` saves the user from typing in the input repository twice. For example:
+
+`processCcd.py $root/hsc --output $root/hsc/rerun/vpk24/DM-4443`
+
+which will run processesCcd on the (HSC) data in "$root/hsc" and put the output in 
+"$root/hsc/rerun/vpk24/DM-4443" can be shortened by using `--rerun` to
+
+`processCcd.py $root/hsc ----rerun vpk24/DM-4443`
+
+This is possible because as of the adoption of RFC-95, outputs from the pipeline usually go within the
```
r-owen: I suggest rewording this. RFC-95 is not useful information here (in my opinion, and if it is important, it should be a URL). I suggest this approach: "This is useful because outputs from pipelines usually go within...". If you want to provide more information you could include a link to RFC-95 later, but please first present what users need to know, then treat that link as "for more information see...".

doc/main.dox
```diff
@@ -119,6 +119,43 @@ Thus the following example specifies all sensors for four combinations of visit
...
+`processCcd.py $root/hsc --output $root/hsc/rerun/vpk24/DM-4443`
+
+which will run processesCcd on the (HSC) data in "$root/hsc" and put the output in 
+"$root/hsc/rerun/vpk24/DM-4443" can be shortened by using `--rerun` to
+
+`processCcd.py $root/hsc ----rerun vpk24/DM-4443`
+
+This is possible because as of the adoption of RFC-95, outputs from the pipeline usually go within the 
+camera-level root repository into a subdirectory of a "rerun" directory present in the camera-level root
+repository. These output sub-directories should usually be named using JIRA issue numbers and dates, again
```
r-owen: Please remove "These output sub-directories should usually be named...". This is an LSST policy that may not apply to any other camera team. I agree this information is useful for LSST, but this is not the right document for it. That said, you could still provide the examples. That will give folks some idea of why this is useful.

doc/main.dox
jdswinbank: I'd actually be against linking to the RFC at all, unless you really have to: the stack documentation should capture all the useful information, leaving the RFC as purely an historical artefact.

doc/main.dox
```diff
@@ -119,6 +119,43 @@ Thus the following example specifies all sensors for four combinations of visit
...
+with a distinction between public and private reruns. For example:
+
+`$root/hsc/rerun/DM-3456_20150607`   # a public rerun useful for many people
```
r-owen: I suggest removing "DM-3456_". A general run need not be associated with any ticket, and it makes the example more general.

doc/main.dox
RobertLuptonTheGood: The example is clear!  It's a path relative to $root.  If absolute, it's taken to be the path

doc/main.dox
```diff
@@ -119,6 +119,43 @@ Thus the following example specifies all sensors for four combinations of visit
...
+
+`$root/hsc/rerun/jbosch/DM-7890/fixed-psf`   # one of several private reruns used to debug a single issue
+
+`$root/lsstSim/krughoff/DM-5678/rerun/jbosch/DM-5679`   # simulated by Simon, processed by Jim
```
r-owen: Again, please remove "DM-5678/". Also, please do not expect users to know that "krughoff" is Simon and "jbosch" is Jim. Please reword accordingly.

doc/main.dox
```diff
@@ -119,6 +119,43 @@ Thus the following example specifies all sensors for four combinations of visit
...
+
+As with any other output repository, reruns can be chained, but keep in mind that the name of the parent rerun
+is not generally recorded in the name of the child. Chaining is made easy by supporting a two-value 
+form separated by a colon, which specifies both a parent rerun and the new output directory. For example:
```
r-owen: This starts out very unclear. "As with any other output repository, reruns can be chained"? I think you mean that you can read data from one rerun and write it to a new rerun, but if so, please be clearer. Also, I have no idea what you mean by "but keep in mind that the name of the parent rerun is not generally recorded in the name of the child". I suggest that you ditch those first two sentences entirely and start out by explaining that --rerun path1:path2 will read from the rerun specified by path1 and write to the rerun specified by path2. Then if you need to say more about it (e.g. that bit about names of parent reruns) you can provide it as extra details.

doc/main.dox
```diff
@@ -119,6 +119,43 @@ Thus the following example specifies all sensors for four combinations of visit
 plus all sensors for one raft of two other visits for calibrated lsstSim data:
 - `--id visit=54123^55523 raft=1,1^2,1 --id visit=623459^293423 raft=0,0`
 
+\subsection pipeBase_argumentParser_rerun The --rerun Option
+
+The `--rerun` option is a convenient shortcut for specifying the output directory of a command line task and
+serves the same purpose as the `--output` argument. However, the value supplied to `--rerun` is relative to
+the "rerun" directory of the root input data repository (i.e. follow the "_parent" chain all the way back,
+then add "rerun" to that). `--rerun` saves the user from typing in the input repository twice. For example:
```
r-owen: This is pretty good, but I think it could be clearer. How about starting out with something like: "The --rerun option is an alternate way to specify the output (and optionally the input) repositories for a command line task. Unlike `--output`, the value supplied to `--rerun` is relative to..." (where the rest is as you wrote it).

I think it would be clearer and is less likely to trouble the user with the nagging question "what if I specify --output and --rerun", since it's pretty clear that will be an error, or at least a bad idea.

doc/main.dox
r-owen: It may be clear to you, but it was not to me. It takes just a few words to say that if the path is absolute that the rerun dir is ignored (though that begs the question of why not just use --output)

doc/main.dox
r-owen: I agree.

PR #19, opened 2016-08-07
DM-6999: Logging framework migration

hchiang2:


python/lsst/pipe/base/task.py
```diff
@@ -122,9 +122,10 @@ def __init__(self, config=None, name=None, parentTask=None, log=None):
             if config is None:
                 config = self.ConfigClass()
             self._taskDict = dict()
-            if log is None:
-                log = pexLog.getDefaultLog()
-            self.log = pexLog.Log(log, self._fullName)
+            if log is not None and isinstance(log, Log):
```
r-owen: Why is the `isinstance` test? This was not present before and I'm very uncomfortable having it now. Is this a way of trying to avoid pex_logging? If so, is the problem serious enough to need avoiding? If so, please warn or raise if the wrong type of log is passed in. I am uncomfortable with code that silently ignores issues, as it leads to mysterious and silent failures.

python/lsst/pipe/base/task.py
```diff
@@ -122,9 +122,10 @@ def __init__(self, config=None, name=None, parentTask=None, log=None):
...
+                self.log = Log.getLogger(log.getName() + '.' + self._fullName)
+            else:
+                self.log = Log.getLogger(self._fullName)
```
timj: Would it be clearer as:
```python
name = self._fullName
if log is not None and hasattr(log, "getName"):
    name = log.getName() + '.' + name
self.log = Log.getLogger(name)
```
this makes it clearer that the logic is worrying about the name rather than dealing with something more complex.

r-owen: pex_logging logs had getName as well, so if this is an attempt to avoid pex_logging then that won't work.

But I'd still rather avoid the check altogether, if we can, and certainly warn users if the check fails.

timj: No idea, but I'd still be in favor of reorganizing the logic to make it clear this is entirely about obtaining a name.

tests/testArgumentParser.py
```diff
@@ -298,46 +297,35 @@ def testAtFile(self):
...
-            self.assertEqual(namespace.log.getThresholdFor("foo.bar"), fooBarLevel)
-            self.assertEqual(namespace.log.getThresholdFor("baz"), bazLevel)
+            self.assertEqual(namespace.log.getLevel(), intLevel)
+            self.assertEqual(lsstLog.Log.getLogger("foo.bar").getLevel(), intLevel)
+            self.assertEqual(lsstLog.Log.getLogger("baz").getLevel(), getattr(lsstLog.Log, bazLevel))
+
+        self.assertRaises(SystemExit, self.ap.parse_args,
+                          config=self.config,
+                          args=[DataPath, "--loglevel", "1234"],
+                          )
```
timj: Please use `with self.asserRaises` (and in other places in this file). It makes it much clearer what is being tested.

python/lsst/pipe/base/task.py
hchiang2: `isinstance` is not needed here and we can just remove the check.  What you said makes more sense; even if the check stays it should warn or raise if the wrong type of log is passed in.

python/lsst/pipe/base/task.py
hchiang2: I've removed the check, and rewrote it to make the logic more clear.

python/lsst/pipe/base/task.py
```diff
@@ -111,7 +111,7 @@ def __init__(self, config=None, name=None, parentTask=None, log=None):
             if config is None:
                 config = getattr(parentTask.config, name)
             self._taskDict = parentTask._taskDict
-            self.log = Log.getLogger(parentTask.log.getName() + '.' + name)
+            loggerName = parentTask.log.getName() + '.' + name
```
r-owen: Please don't change the original logic unless you have a compelling reason to do so (and if you do have such a reason, please explain it in the commit message). This changes the API and I think the old one sufficed, though it was not properly documented (thus my small commit which updated the documentation for the log argument, which your change then rendered incorrect).

I encourage you to compare your code to the master branch. I think you'll find that almost no changes are required other than documenting the new type of the log argument and tweaking the way the default log is obtained.

hchiang2: I don't think the logic or API is changed, and your commit is still correct.  Do I misunderstand @timj that separating the logger name operation from getting the logger
would make this clearer? 

If so, I'll remove this commit, and simply go back to remove the unnecessary `isinstance` check.

timj: Your change agreed with my comment to get rid of duplication of all the `Log.getLogger` calls and allow the logic to focus on the real issue here which was choosing the name to pass into `getLogger`. I don't see how your change changes the API at all.

r-owen: @hchiang2 Does the new `log` package have the concept of child logs that share the same stream as a parent log, like the older `pex_logging` package? If so, then you have lost that inheritance with your changes and should revisit the code. If the new `log` package has no such inheritance then your changes are safe.

The documentation for the `log` argument is weak, which is my fault, but I'd ask you to see if you can clean it up here (or I'll take a pass at it if you prefer). "The full task name is used..." is not informative. Also, if your changes are correct then the documentation would be improved by stating that the `log` argument is only used for its name.

hchiang2: Unlike `pex_logging`, the child logs do not have to be chained explicitly to the parent log for the logger hierarchy to work using the new `log`.    Now I see where the confusion comes from.  Ensuring only logger names wouldn't work with `pex_logging` but `log` solves that for us.

tests/testArgumentParser.py
hchiang2: Corrected in a new commit

python/lsst/pipe/base/task.py
```diff
@@ -127,7 +127,7 @@ def __init__(self, config=None, name=None, parentTask=None, log=None):
                 config = self.ConfigClass()
             self._taskDict = dict()
             loggerName = self._fullName
-            if log is not None:
+            if log is not None and log.getName():
```
AstroVPK: Good that this is fixed! This was causing a fairly obtuse error!

timj: `getName()` can return an empty string?

AstroVPK: Apparently yes. Try building pipe_base without this commit - testCmdLineTask fails with a fairly unhelpful error message.

hchiang2: Yes, an empty string for the root logger.  

Maybe we should improve the error message when we ask for a logger that doesn't make sense...

AstroVPK: Sure, I think that's a good idea.

hchiang2: [DM-7479](https://jira.lsstcorp.org/browse/DM-7479) is added.

PR #21, opened 2016-08-19
DM-7245: Port to Python 3

timj:


tests/testTask.py
```diff
@@ -1,3 +1,4 @@
+from past.builtins import basestring
```
timj: Might want to move this to the rest of the imports below.

tests/testCmdLineTask.py
```diff
@@ -28,14 +30,14 @@
 import lsst.utils
 import lsst.pex.logging as pexLog
 import lsst.pipe.base as pipeBase
-from lsst.obs.test import TestConfig
+import lsst.obs.test
```
timj: @PaulPrice you need to do this to stop pytest auto-discovering a class called `Test*` in the module namespace and deciding that it's a test class.

PR #24, opened 2016-10-01
Log the arguments when running a task

r-owen:
Also log the command if run from the command line

### DIFF
--- a/python/lsst/pipe/base/cmdLineTask.py
+++ b/python/lsst/pipe/base/cmdLineTask.py
@@ -528,8 +528,11 @@ class CmdLineTask(Task):
             except Exception as exc:
                 raise type(exc)("Unable to read stored config file %s (%s); consider using --clobber-config" %
                                 (configName, exc))
-            output = lambda msg: self.log.fatal("Comparing configuration: " + msg)
-            if not self.config.compare(oldConfig, shortcut=False, output=output):
+
+            def logConfigMismatch(msg):
+                self.log.fatal("Comparing configuration: " + msg)
+
+            if not self.config.compare(oldConfig, shortcut=False, output=logConfigMismatch):
                 raise TaskError(
                     ("Config does not match existing task config %r on disk; tasks configurations " +
                      "must be consistent within the same output repo (override with --clobber-config)") %
--- a/python/lsst/pipe/base/cmdLineTask.py
+++ b/python/lsst/pipe/base/cmdLineTask.py
@@ -28,6 +28,7 @@ import contextlib
 from builtins import str
 from builtins import object
 
+import lsst.utils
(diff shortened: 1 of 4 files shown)


python/lsst/pipe/base/cmdLineTask.py
```diff
@@ -464,10 +465,19 @@ def parseAndRun(cls, args=None, config=None, log=None, doReturnResults=False):
...
+            args = sys.argv[1:]
+        else:
+            commandAsStr = "{}{}".format(lsst.utils.get_caller_name(skip=1), tuple(args))
+
         argumentParser = cls._makeArgumentParser()
         if config is None:
             config = cls.ConfigClass()
         parsedCmd = argumentParser.parse_args(config=config, args=args, log=log, override=cls.applyOverrides)
+        # print this message after parsing the command so the log is fully configured
+        parsedCmd.log.info("Running: {}".format(commandAsStr))
```
hsinfang: This is okay but maybe better `parsedCmd.log.info("Running: %s", commandAsStr)` as `log` does the % formatting (and only the % formatting at this moment...)

python/lsst/pipe/base/cmdLineTask.py
```diff
@@ -518,8 +528,11 @@ def writeConfig(self, butler, clobber=False, doBackup=True):
             except Exception as exc:
                 raise type(exc)("Unable to read stored config file %s (%s); consider using --clobber-config" %
                                 (configName, exc))
-            output = lambda msg: self.log.fatal("Comparing configuration: " + msg)
-            if not self.config.compare(oldConfig, shortcut=False, output=output):
+
+            def logConfigMismatch(msg):
+                self.log.fatal("Comparing configuration: " + msg)
```
hsinfang: Similarly, maybe delay the formatting to `log`?

python/lsst/pipe/base/cmdLineTask.py
r-owen: My understanding is that delayed formatting is useful primarily when a log message may not be displayed. I would expect info and fatal to always be displayed. Still...it ought to be a harmless change and I guess it is good log style.

hsinfang: True, practically it only makes some differences in debugging messages.

hsinfang: It's the same %s formatting, just done after checking the log level.

r-owen: A quick test shows that %s does work on arbitrary objects, as does %r. I just pushed a new commit that changes logging to use deferred formatting in all instances I found in `pipe_base` (which were all in `cmdLineTask.py` and `argumentParser.py`)

PR #25, opened 2016-10-10
Add support for --show history=config.foo.bar

r-owen:
The "config." is optional

### DIFF
--- a/python/lsst/pipe/base/argumentParser.py
+++ b/python/lsst/pipe/base/argumentParser.py
@@ -39,6 +39,7 @@ from builtins import object
 
 import lsst.utils
 import lsst.pex.config as pexConfig
+import lsst.pex.config.history
 import lsst.log as lsstLog
 import lsst.daf.persistence as dafPersist
 from future.utils import with_metaclass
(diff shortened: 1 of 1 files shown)


python/lsst/pipe/base/argumentParser.py
```diff
@@ -715,9 +716,10 @@ def obeyShowArgument(showOpts, config=None, exit=False):
         return
 
     for what in showOpts:
-        mat = re.search(r"^config(?:=(.+))?", what)
-        if mat:
-            pattern = mat.group(1)
+        matConfig = re.search(r"^config(?:=(?:config.)?(.+))?", what)
+        matHistory = None if matConfig else re.search(r"^history(?:=(?:config.)?(.+))?", what)
```
r-owen: I find this too confusing. The search string "^history(?:=(?:config.)?(.+))?" seems far trickier than is required, and making matHistory use the results of matConfig seems needlessly complicated. It would be far clearer to keep the two independent of each other, even if it made the code a bit more verbose.

The number of options is large enough now that it may be worth a different approach, such as first testing all options that have no associated data, then using something like this for the remaining possibilities:
```
if what == "tasks":
   ....
elif what == "data":
    ....
...  # test all items that cannot have associated data first, then...
else:
    item, data = what.split("=", 1)
    if item == "config":
        configMatch = ...
    elif item == "history":
        historyMatch = ...
```

python/lsst/pipe/base/argumentParser.py
```diff
@@ -750,6 +752,32 @@ def write(self, showStr):
...
+            for i, cpt in enumerate(cpath):
+                try:
+                    hconfig = getattr(hconfig, cpt)
+                except AttributeError:
+                    print("Error: configuration %s has no subconfig %s" %
+                          (".".join(["config"] + cpath[:i]), cpt), file=sys.stderr)
+
+                    sys.exit(1)                
+
+            import lsst.pex.config.history as pch
```
r-owen: Please import this at the top along with all other imports

PR #27, opened 2016-11-11
DM-8255: Allow using existing rerun folders

hsinfang:
Currently, tasks cannot run if the rerun folder exists but is empty.

### DIFF
--- a/tests/testArgumentParser.py
+++ b/tests/testArgumentParser.py
@@ -23,6 +23,8 @@
 from __future__ import print_function
 import itertools
 import os
+import shutil
+import tempfile
 import unittest
 import contextlib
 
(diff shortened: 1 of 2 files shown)


tests/testArgumentParser.py
```diff
@@ -506,6 +507,15 @@ def testOutputs(self):
...
         self.assertEqual(args.rerun, ["foo", "bar"])
 
+        # Specified by rerun with an existing empty rerun folder
+        path = tempfile.mkdtemp(dir=os.path.join(DataPath, "rerun"))
+        rerun = os.path.basename(path)
+        args = parser.parse_args(config=self.config, args=[DataPath, "--rerun", rerun])
+        self.assertEqual(args.input, DataPath)
+        self.assertEqual(args.output, os.path.join(DataPath, "rerun", rerun))
+        self.assertEqual(args.rerun, [rerun])
+        shutil.rmtree(path)
```
PaulPrice: The `rmtree` should be in a `finally` clause.

hsinfang:
As discussed in [DM-8255](https://jira.lsstcorp.org/browse/DM-8255`), this makes it behave more similarly to `--output`. The directory of `--output` can exist without a `_parent` link.

tests/testArgumentParser.py
hsinfang: Fixed

PR #31, opened 2017-04-14
DM-10229: Do not write to upstream repository.

jdswinbank:
Reorganize the test case such that it doesn't attempt to create reruns inside
DataPath (which is part of obs_test).


jdswinbank:
@RobertLuptonTheGood — I've added some more comments in the code itself, and an extra sentence to the commit message. Does that scratch your itch? If so, could you please mark this as "reviewed" on JIRA?

PR #33, opened 2017-05-25
ArgumentParser: add calibRoot to output mapperArgs

PaulPrice:
Otherwise, the output repo doesn't know about the calib storage and
writes calibs in the output storage rather than the calib storage.


n8pease:
looks good.

PR #37, opened 2017-07-20
DM-11253: User documentation for command-line tasks (and Sphinx set up)

jonathansick:
This PR adds a Sphinx documentation configuration to `pipe_base`, and also adds user documentation for the command-line task infrastructure.

Key URLs:

- Homepage: https://pipelines.lsst.io/v/DM-11253/index.html
- `pipe_base` package homepage: https://pipelines.lsst.io/v/DM-11253/packages/pipe_base/index.html
- `lsst.pipe.base` module homepage: https://pipelines.lsst.io/v/DM-11253/modules/lsst.pipe.base/index.html

The [Using command-line tasks](https://pipelines.lsst.io/v/DM-11253/modules/lsst.pipe.base/index.html#using-command-line-tasks) topics are new material. The overview and developer topics are material ported from `main.dox`.

I've also converted all the docstrings in `lsst.pipe.base` to Numpydoc to give us a preview of what that is like for a `pipe_base`-scale package.

### DIFF
--- a/doc/lsst.pipe.base/command-line-task-argument-reference.rst
+++ b/doc/lsst.pipe.base/command-line-task-argument-reference.rst
@@ -134,9 +134,13 @@ Other named arguments are optional.
 
    The ``-L``/``--loglevel`` argument can appear multiple times.
 
+   For more information, see :ref:`command-line-task-logging-howto`.
+
 .. option:: --longlog
 
    **Enable the verbose logging format.**
+
+   See :ref:`command-line-task-logging-howto-longlog` for more information.
 
 .. option:: --debug
 
--- a/doc/lsst.pipe.base/command-line-task-argument-reference.rst
+++ b/doc/lsst.pipe.base/command-line-task-argument-reference.rst
@@ -70,11 +70,15 @@ Other named arguments are optional.
 
    The ``-c``/``--config`` argument can appear multiple times.
 
+   See :ref:`command-line-task-config-howto-config` for more information.
+
 .. option:: -C <configfile>, --configfile <configfile>
 
    **Task configuration override file(s).**
 
    The ``-C``/``--configfile`` argument can appear multiple times.
+
+   See :ref:`command-line-task-config-howto-configfile` for more information.
 
 .. option:: --clobber-config
 
--- a/doc/lsst.pipe.base/command-line-task-argument-reference.rst
+++ b/doc/lsst.pipe.base/command-line-task-argument-reference.rst
@@ -1,0 +1,349 @@
+.. _command-line-task-argument-reference:
+
+####################################
+Command-line task argument reference
+####################################
+
+This page describes the command-line arguments and environment variables common to command-line tasks.
+
+Signature and syntax
+====================
+
+The basic call signature of a command-line task is:
+
+.. code-block:: text
+
+   task.py REPOPATH [@file [@file2 ...]] [--output OUTPUTREPO | --rerun RERUN] [named arguments]
+
+See :ref:`command-line-argument-files` for details on ``@file`` syntax.
+
+For named arguments that take multiple values do not use a ``=`` after the argument name.
+For example, ``--configfile foo.py bar.py``, **not** ``--configfile=foo bar``.
+
+Status code
+===========
+
+A command-line task returns a status code of ``0`` if all data IDs were successfully processed.
+If the command-line task failed to process one or more data IDs, the status code is equal to the number of data IDs that failed.
+See also: :option:`--noExit`.
+
+Positional arguments
+====================
+
+.. option:: REPOPATH
+
+   **Input Butler data repository URI or path.**
+
+   The input Butler data repository is always the first argument to a command-line task.
+   This argument is required for all command-line task runs, except when printing help (:option:`--help`).
+
+   In general, this is a URI that depends on the Butler backend.
+   For example, ``swift://host/path`` for a Swift backend or ``file://path`` for a POSIX backend.
+
+   .. TODO DM-11671 Link to Butler docs on URIs.
+
+   For POSIX backends, this may also be an absolute file path or a path relative to the current working directory.
+   
+   If the :envvar:`PIPE_INPUT_ROOT` environment variable is set, then the ``REPOPATH`` is relative to that.
+   See :ref:`command-line-task-envvar-examples`.
+
+   For background, see :doc:`command-line-task-data-repo-howto`.
+
+   See also :option:`--rerun` argument to specify input and output reruns within this Butler repository.
+
+Named arguments
+===============
+
+An output data repository must be specified with either :option:`--output` **or** :option:`--rerun`.
+Other named arguments are optional.
+
+.. option:: --calib <calib_repo>
+
+   **Calibration data repository URI path.**
+
+   The path may be absolute, relative to the current working directory, or relative to :envvar:`PIPE_CALIB_ROOT` (when set).
+   See :ref:`command-line-task-envvar-examples`.
+
+.. option:: -c <name=val>, --config <name=val>
+
+   **Task configuration overrides.**
+
+   The ``-c``/``--config`` argument can appear multiple times.
+
+.. option:: -C <configfile>, --configfile <configfile>
+
+   **Task configuration override file(s).**
+
+   The ``-C``/``--configfile`` argument can appear multiple times.
+
+.. option:: --clobber-config
+
+   **Backup and overwrite existing config files.**
+
+   Normally a command line task checks existing config files in a Butler repository to ensure that the current configurations are consistent with previous pipeline executions.
+   This argument disables this check, which may be useful for development.
+
+   This argument is safe with :option:`-j` multiprocessing, but not necessarily with other forms of parallel execution.
+
+.. option:: --clobber-output
+
+   **Remove and re-create the output repository if it already exists.**
+
+   This argument is safe with :option:`-j` multiprocessing, but not necessarily with other forms of parallel execution.
+
+.. option:: --clobber-versions
+
+   **Backup and then overwrite existing package version provenance.**
+
+   Normally a command line task checks that the Science Pipelines package versions are the same as for previous executions that wrote to an output repository or rerun.
+   This argument disables this check, which may be useful for development.
+
+   This argument is safe with :option:`-j` multiprocessing, but not necessarily with other forms of parallel execution.
+
+.. option:: -h, --help
+
+   **Print help.**
+
+   The help is equivalent to this documentation page, describing command line arguments.
+   This help does not describe the command line task's specific functionality.
+
+.. option:: --id [[<dataid>] ...]
+
+   **Butler data IDs.**
+
+   Specify data IDs to process using data ID syntax.
+   For example, ``--id visit=12345 ccd=1,2^0,3``.
+
+   An ``--id`` argument without values indicates that **all** data available in the input repository will be processed.
+
+   For many-to-one processing tasks the ``--id`` argument specifies **output** data IDs, while :option:`--selectId` is used for **input** data IDs.
+
+   The ``--id`` argument can appear multiple times.
+
+.. option:: -L <level|component=level> [level|component=level...], --loglevel <level|component=level> [level|component=level...]
+
+   **Log level.**
+
+   Supported levels are: ``trace``, ``debug``, ``info``, ``warn``, ``error``, or ``fatal``.
+
+   Log levels can be set globally (``-L debug``) or for a specific named logger (``-L pipe.base=debug``).
+
+   Specify multiple arguments to control the global and named logging levels simultaneousy (``-L warn pipe.base=debug``).
+
+   The ``-L``/``--loglevel`` argument can appear multiple times.
+
+.. option:: --longlog
+
+   **Enable the verbose logging format.**
+
+.. option:: --debug
+
+   **Enable debugging mode.**
+
+   .. TODO DM-11675 cross-link to debug framework docs in lsst.base module.
+
+.. option:: --doraise
+
+   **Raise an exception on error.**
+
+   This mode causes the task to exit early if it encounters an error, rather than logging the error and continuing.
+
+.. option:: --no-backup-config
+
+   **Disable copying config to file~N backup.**
+
+.. option:: --no-versions
+
+   **Disable package version consistency validation.**
+
+   This mode permits data processing even if outputs exist in the output data repository or rerun from a different version of Science Pipelines packages.
+
+   This mode is useful for development should not be used in production processing.
+
+   See also :option:`--clobber-versions`.
+
+.. option:: --output <output_repo>
+
+   **Output data repository path.**
+
+   The output data repository will be created if it does not exist.
+
+   The path may be absolute, relative to the current working directory, or relative to :envvar:`PIPE_CALIB_ROOT` (when set).
+   See :ref:`command-line-task-envvar-examples`.
+
+   ``--output`` may not be used with the :option:`--rerun` argument.
+   See :option:`--rerun` for writing outputs to a rerun directory *inside* the input data repository.
+
+.. option:: -j <processes>, --processes <processes>
+
+   **Number of processes to use.**
+
+   When processes is larger than 1 the task uses the Python `multiprocessing` module to parallelize processing of multiple datasets across multiple processors.
+
+   See also :option:`--timeout`.
+
+.. option:: --profile <profile>
+
+   **Dump cProfile statistics to the named file.**
+
+   See the cProfile_ documentation.
+
+.. option:: --rerun <[input:]output>
+
+   **Specify output (and input) rerun (and optionally the input rerun as well).**
+
+   Reruns are named data repositories inside the :file:`rerun` directory of the root data repository (:option:`REPOPATH`).
+
+   ``--rerun output`` is equivalent to ``--output REPOPATH/rerun/output``.
+
+   An input rerun can also, optionally, be specified.
+   ``--rerun input:output`` sets the input repository path to ``ROOT/rerun/input`` the output repository path to ``REPOPATH/rerun/output``.
+
+   If an argument to `--rerun` starts with a `/`, it will be interpreted as an absolute path rather than as being relative to the root input data repository.
+
+   The arguments supplied to `--rerun` may refer to symbolic links to directories.
+   Data will be read or written from the links' targets.
+
+.. option:: --show <config|data|tasks|run>
+
+   **Print metadata without processing.**
+
+   Permitted values are:
+
+   - ``config``: show configuration state.
+
+   - ``data``: show data IDs resolved by the :option:`--id` argument
+
+   - ``tasks``: show sub-tasks run by the command-line task.
+
+   Multiple values can be shown at once.
+   For example, ``--show config data``.
+
+   Normally the command-line task will exit before processing any data.
+   If you want to *also* run the task after showing metadata, append the ``run`` value.
+   For example, ``--show config data run``.
+
+.. option:: --selectId
+
+   **Input data IDs for many-to-one tasks.**
+
+   For many-to-one processing tasks, such as coaddition, the :option:`--selectId` argument is used to specify input data IDs, while :option:`--id` is used to specify *output* data IDs.
+   The syntax for :option:`--selectId` is identical to that of :option:`--id`.
+
+.. option:: -t timeout, --timeout timeout
+
+   **Multiprocessing timeout (in seconds).***
+
+   See also :option:`-j`.
+
+.. option:: --noExit
+
+   **(Advanced) prevent the command-line task from exiting directly to the shell with a non-zero status code if there are one or more processing failures.**
+
+   If there are failures, by default, a command-line task exits to the directly shell with a status code equal to the number of data IDs that it failed to process.
+   This means that the command-line task does not return to the :ref:`run script <creating-a-command-line-task-run-script>` that originally called the `lsst.pipe.base.CmdLineTask.parseAndRun` method if there is an error.
+   Some command-line tasks (such as the MPI-enabled scripts in ``pipe_drivers``) need `lsst.pipe.base.CmdLineTask.parseAndRun` to always return to the run script.
+   In that case, use this ``--noExit`` argument.
+
+   When ``--noExit`` is used, the command-line task will not exit to the shell from `lsst.pipe.base.CmdLineTask.parseAndRun` if failures are encountered.
+   Instead, it will return normally to the run script that called `~lsst.pipe.base.CmdLineTask.parseAndRun`.
+   In this case, it is up to the run script to set an appropriate shell status code.
+
+   See also :option:`--doraise`.
+
+.. _command-line-argument-files:
+
+Argument files
+==============
+
+Arguments can be written to a plain text file and referenced with an ``@filepath`` command-line argument.
+The contents of argument files are identical to what you'd write on the command line, with these rules:
+
+- Text can be split across multiple lines.
+  For example, you can have one argument per line.
+
+- Do not use ``\`` as a continuation character.
+
+- Include comments with a ``#`` character.
+  Content on a line after the ``#`` character is ignored.
+
+- Blank lines and lines starting with ``#`` are ignored.
+
+You can mix argument files with other command-line arguments (including additional :option:`--id` and :option:`--config` arguments).
+
+You can include multiple ``@filepath`` references in the same command.
+
+Example
+-------
+
+For example, the file :file:`foo.txt` contains:
+
+.. code-block:: text
+
+   --id visit=54123^55523 raft=1,1^2,1 # data ID
+   --config someParam=someValue --configfile configOverrideFilePath
+
+You can then reference it with ``@foo.txt``, along with additional command-line arguments:
+
+.. code-block:: bash
+
+   task.py repo @foo.txt --config anotherParam=anotherValue --output outputPath
+
+.. _command-line-task-envvar:
+
+Environment variables
+=====================
+
+The :envvar:`PIPE_INPUT_ROOT`, :envvar:`PIPE_CALIB_ROOT`, and :envvar:`PIPE_OUTPUT_ROOT` environment variables let you more easily specify Butler data repositories.
+
+Each environment variable is used as a root directory for relative paths provided on the command line.
+If you set an absolute path on the command line, the environment variable is ignored.
+:ref:`see examples <command-line-task-envvar-examples>`.
+
+.. The default value for each of these environment variables is the current working directory.
+
+.. envvar:: PIPE_INPUT_ROOT
+
+   Root directory for the input Butler data repository argument, :option:`REPOPATH`.
+
+.. envvar:: PIPE_CALIB_ROOT
+
+   Root directory for the calibration Butler data repository argument (--calib).
+
+.. envvar:: PIPE_OUTPUT_ROOT
+
+   Root directory for the output Butler data repository argument (--output).
+
+.. _command-line-task-envvar-examples:
+
+Path environment variable examples
+----------------------------------
+
+These examples feature :envvar:`PIPE_INPUT_ROOT` to help specify the input data repository along with :option:`REPOPATH`, which is the first positional argument of any command.
+
+1. The data repository path is :file:`$PIPE_INPUT_ROOT/DATA` (or :file:`DATA` if :envvar:`PIPE_INPUT_ROOT` is undefined):
+   
+   .. code-block:: bash
+
+      processCcd.py DATA [...]
+
+2. The data repository path is :file:`$PIPE_INPUT_ROOT` (or current working directory if :envvar:`PIPE_INPUT_ROOT` is undefined):
+
+   .. code-block:: bash
+
+      processCcd.py . [...]
+
+3. The data repository path is an absolute path:
+   
+   .. code-block:: bash
+
+      processccd.py /DATA/a [...]
+
+   :envvar:`PIPE_INPUT_ROOT` is ignored in this case:
+
+The same behavior applies to the named arguments:
+
+- :option:`--calib` with :envvar:`PIPE_CALIB_ROOT`.
+- :option:`--output` with :envvar:`PIPE_OUTPUT_ROOT`.
+
+.. _cProfile: https://docs.python.org/library/profile.html
--- a/doc/lsst.pipe.base/command-line-task-argument-reference.rst
+++ b/doc/lsst.pipe.base/command-line-task-argument-reference.rst
@@ -80,7 +80,7 @@ Other named arguments are optional.
 
    **Backup and overwrite existing config files.**
 
-   Normally a command line task checks existing config files in a Butler repository to ensure that the current configurations are consistent with previous pipeline executions.
+   Normally a command-line task checks existing config files in a Butler repository to ensure that the current configurations are consistent with previous pipeline executions.
    This argument disables this check, which may be useful for development.
 
    This argument is safe with :option:`-j` multiprocessing, but not necessarily with other forms of parallel execution.
@@ -95,7 +95,7 @@ Other named arguments are optional.
 
    **Backup and then overwrite existing package version provenance.**
 
-   Normally a command line task checks that the Science Pipelines package versions are the same as for previous executions that wrote to an output repository or rerun.
+   Normally a command-line task checks that the Science Pipelines package versions are the same as for previous executions that wrote to an output repository or rerun.
    This argument disables this check, which may be useful for development.
 
    This argument is safe with :option:`-j` multiprocessing, but not necessarily with other forms of parallel execution.
@@ -104,8 +104,8 @@ Other named arguments are optional.
 
    **Print help.**
 
-   The help is equivalent to this documentation page, describing command line arguments.
-   This help does not describe the command line task's specific functionality.
+   The help is equivalent to this documentation page, describing command-line arguments.
+   This help does not describe the command-line task's specific functionality.
 
 .. option:: --id [[<dataid>] ...]
 
@@ -164,7 +164,7 @@ Other named arguments are optional.
 
 .. option:: --output <output_repo>
 
-   **Output data repository path.**
+   **Output data repository URI or path.**
 
    The output data repository will be created if it does not exist.
 
@@ -172,7 +172,8 @@ Other named arguments are optional.
    See :ref:`command-line-task-envvar-examples`.
 
    ``--output`` may not be used with the :option:`--rerun` argument.
-   See :option:`--rerun` for writing outputs to a rerun directory *inside* the input data repository.
+
+   See :doc:`command-line-task-data-repo-howto` for background.
 
 .. option:: -j <processes>, --processes <processes>
 
@@ -190,19 +191,20 @@ Other named arguments are optional.
 
 .. option:: --rerun <[input:]output>
 
-   **Specify output (and input) rerun (and optionally the input rerun as well).**
+   **Specify output rerun (and optionally the input rerun as well).**
 
-   Reruns are named data repositories inside the :file:`rerun` directory of the root data repository (:option:`REPOPATH`).
-
+   Reruns are data repositories relative to the root repository, :option:`REPOPATH`.
    ``--rerun output`` is equivalent to ``--output REPOPATH/rerun/output``.
 
    An input rerun can also, optionally, be specified.
-   ``--rerun input:output`` sets the input repository path to ``ROOT/rerun/input`` the output repository path to ``REPOPATH/rerun/output``.
+   ``--rerun input:output`` sets the input repository path to ``REPOPATH/rerun/input`` the output repository path to ``REPOPATH/rerun/output``.
 
    If an argument to `--rerun` starts with a `/`, it will be interpreted as an absolute path rather than as being relative to the root input data repository.
 
    The arguments supplied to `--rerun` may refer to symbolic links to directories.
    Data will be read or written from the links' targets.
+
+   See :doc:`command-line-task-data-repo-howto` for more information.
 
 .. option:: --show <config|data|tasks|run>
 
--- a/doc/lsst.pipe.base/command-line-task-argument-reference.rst
+++ b/doc/lsst.pipe.base/command-line-task-argument-reference.rst
@@ -195,6 +195,8 @@ Other named arguments are optional.
 
    See also :option:`--timeout`.
 
+   See :ref:`command-line-task-parallel-howto` for more information.
+
 .. option:: --profile <profile>
 
    **Dump cProfile statistics to the named file.**
--- a/doc/lsst.pipe.base/command-line-task-argument-reference.rst
+++ b/doc/lsst.pipe.base/command-line-task-argument-reference.rst
@@ -113,12 +113,14 @@ Other named arguments are optional.
 
    Specify data IDs to process using data ID syntax.
    For example, ``--id visit=12345 ccd=1,2^0,3``.
+   For more information, see :ref:`command-line-task-dataid-howto`.
 
-   An ``--id`` argument without values indicates that **all** data available in the input repository will be processed.
+   An ``--id`` argument without values indicates that **all** data available in the input repository will be processed (see :ref:`command-line-task-dataid-howto-wildcard`).
 
    For many-to-one processing tasks the ``--id`` argument specifies **output** data IDs, while :option:`--selectId` is used for **input** data IDs.
 
    The ``--id`` argument can appear multiple times.
+   See :ref:`command-line-task-dataid-howto-multi-arg`.
 
 .. option:: -L <level|component=level> [level|component=level...], --loglevel <level|component=level> [level|component=level...]
 
@@ -231,6 +233,8 @@ Other named arguments are optional.
 
    For many-to-one processing tasks, such as coaddition, the :option:`--selectId` argument is used to specify input data IDs, while :option:`--id` is used to specify *output* data IDs.
    The syntax for :option:`--selectId` is identical to that of :option:`--id`.
+
+   For more information about dataId selection syntax, see :ref:`command-line-task-dataid-howto`.
 
 .. option:: -t timeout, --timeout timeout
 
--- a/doc/lsst.pipe.base/command-line-task-argument-reference.rst
+++ b/doc/lsst.pipe.base/command-line-task-argument-reference.rst
@@ -85,6 +85,8 @@ Other named arguments are optional.
 
    This argument is safe with :option:`-j` multiprocessing, but not necessarily with other forms of parallel execution.
 
+   See :ref:`command-line-task-prov-howto-config` for more information.
+
 .. option:: --clobber-output
 
    **Remove and re-create the output repository if it already exists.**
@@ -99,6 +101,8 @@ Other named arguments are optional.
    This argument disables this check, which may be useful for development.
 
    This argument is safe with :option:`-j` multiprocessing, but not necessarily with other forms of parallel execution.
+
+   See :ref:`command-line-task-prov-howto-versions` for more information.
 
 .. option:: -h, --help
 
@@ -167,6 +171,8 @@ Other named arguments are optional.
    This mode is useful for development should not be used in production processing.
 
    See also :option:`--clobber-versions`.
+
+   See :ref:`command-line-task-prov-howto-versions` for more information.
 
 .. option:: --output <output_repo>
 
--- a/doc/lsst.pipe.base/command-line-task-config-howto.rst
+++ b/doc/lsst.pipe.base/command-line-task-config-howto.rst
@@ -1,0 +1,151 @@
+.. _command-line-task-config-howto:
+
+##############################
+Configuring command-line tasks
+##############################
+
+:ref:`Command-line tasks <using-command-line-tasks>` are highly configurable.
+This page describes how to review configurations with :option:`--show config <--show>`,  and change configurations on the command line with :option:`--config` or :option:`--configfile`.
+
+One special category of configuration is subtask retargeting.
+Retargeting uses concepts discussed on this page, but with additional considerations described in :doc:`command-line-task-retargeting-howto`.
+
+.. _command-line-task-config-howto-show:
+
+How to show a command-line task's current configuration
+=======================================================
+
+The :option:`--show config <--show>` argument allows you to review the current configuration details for command-line tasks. 
+
+.. tip::
+
+   :option:`--show config <--show>` puts the command-line task in a dry-run mode where no data processed.
+   Once you're done reviewing configurations, remove the :option:`--show` argument from the command-line task invocation to actually process data.
+
+   Alternatively, add ``run`` to the :option:`--show` argument to show configurations *and* process data: :option:`--show config run <--show>`.
+
+.. _command-line-task-config-howto-show-all:
+
+Viewing all configurations
+--------------------------
+
+Use :option:`--show config <--show>` to see all current configurations for a command-line task.
+For example:
+
+.. code-block:: bash
+
+   processCcd.py REPOPATH --output output --show config
+
+Note that both the input and output Butler data repositories repository must be set to show camera-specific task configuration defaults.
+Instead of an :option:`--output` argument you can specify a :option:`--rerun`, see :doc:`command-line-task-data-repo-howto`.
+
+.. _command-line-task-config-howto-show-subset:
+
+Viewing a subset of configurations
+----------------------------------
+
+To filter the :option:`--show config <--show>` output, include a search term with wildcard matching (``*``) characters.
+For example, this will show any configuration that includes the string ``calibration``:
+
+.. code-block:: bash
+
+   processCcd.py REPOPATH --output output --show config="*calibration*"
+
+.. _command-line-task-config-howto-config:
+
+How to set configurations with command-line arguments
+=====================================================
+
+Command-line tasks can be configured through a combination of two mechanisms: arguments on the command line (:option:`--config`) or through configuration files (:option:`--configfile`).
+In general, simple configurations can be made through the command line, while complex configurations and :ref:`subtask retargeting <command-line-task-retargeting-howto>` must done through configuration files (see :ref:`command-line-task-config-howto-configfile`).
+
+To change a configuration value on the command line, pass that configuration name and value to the :option:`--config` argument.
+For example, to set a configuration named ``skyMap.projection`` to a value ``"TAN"``:
+
+.. code-block:: bash
+
+   task.py REPOPATH --output output --config skyMap.projection="TAN"
+
+You can provide multiple :option:`--config` arguments on the same command line or set multiple configurations with a single :option:`--config` argument:
+
+.. code-block:: bash
+
+   task.py REPOPATH --output output --config config1="value1" config2="value2"
+
+Only simple configuration values can be set through :option:`--config` arguments, such as:
+
+- **String values**. For example: ``--config configName="value"``.
+- **Scalar numbers**. For example: ``--config configName=2.5``.
+- **Lists of integers**. For example: ``--config intList=2,4,-87``.
+- **List of floating point numbers**. For example: ``--config floatList=3.14,-5.6e7``.
+- **Boolean values**. For example: ``--config configName=True configName2=False``.
+
+Specific types of configurations you **cannot** perform with the :option:`--config` argument are:
+
+- You cannot :doc:`retarget a subtask <command-line-task-retargeting-howto>` specified by a `lsst.pex.config.ConfigurableField` (which is the most common case).
+- For items in registries, you can only specify values for the active (current) item.
+  See :ref:`command-line-task-retargeting-howto-registry-active-config`.
+- You cannot specify values for lists of strings.
+- You cannot specify a subset of a list.
+  You must specify all values at once.
+
+For these more complex configuration types you must use configuration files, which are evaluated as Python code.
+
+.. _command-line-task-config-howto-configfile:
+
+How to use configuration files
+==============================
+
+You can also provide configurations to a command-line task through a *configuration file*.
+In fact, configuration files are Python modules; anything you can do in Python you can do in a configuration file.
+
+Configuration files give you full access to the configuration API, allowing you to import and :doc:`retarget subtasks <command-line-task-retargeting-howto>`, and set configurations with complex types.
+These configurations can only be done through configuration files, not through command-line arguments.
+
+Use a configuration file by providing its file path through a :option:`-C`/:option:`--configfile` argument:
+
+.. code-block:: bash
+
+   task.py REPOPATH --output output --configfile taskConfig.py
+
+Multiple configuration files can be provided through the same :option:`--configfile` argument and the :option:`--configfile` argument itself can be repeated.
+
+In a configuration file, configurations are attributes of a ``config`` object.
+If on the command line you set a configuration with a ``--config skyMap.projection="TAN"`` argument, in a configuration file the equivalent statement is:
+
+.. code-block:: python
+
+   config.skyMap.projection = "TAN"
+
+``config`` is the root configuration object for the command-line task.
+Settings for the command-line task itself are attributes of ``config``.
+In that example, ``config.skyMap`` is a subtask and ``projection`` is a configuration of that ``skyMap`` subtask.
+
+.. _command-line-task-config-howto-obs:
+
+About configuration defaults and camera configuration override files
+====================================================================
+
+Command-line task configurations are a combination of configurations you provide and defaults from the observatory package and the task itself.
+
+When a command-line task is run, it loads two camera-specific configuration files, if found: one for the observatory package, and one for a specific camera defined in that observatory package.
+For an example observatory package named ``obs_package``, these configuration override files are, in order:
+
+- ``obs_package/config/taskName.py`` (overrides for an observatory package in general).
+- ``obs_package/config/cameraName/taskName.py`` (overrides for a specific camera, named “\ ``cameraName``\ ”).
+
+The ``taskName`` is the command-line task, such as ``processCcd`` for :command:``processCcd.py`` (this is the `lsst.pipe.base.CmdLineTask._DefaultName` class variable).
+
+Here are two examples:
+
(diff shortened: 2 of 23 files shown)


doc/lsst.pipe.base/task-framework-overview.rst
```diff
@@ -0,0 +1,47 @@
...
+
+Tasks may process multiple items of data in parallel, using Python's `multiprocessing` library.
+Support for this is built into the `~lsst.pipe.base.ArgumentParser` and `~lsst.pipe.base.TaskRunner`.
+
+Most tasks have a ``run`` method that performs the primary data processing.
+Each task's `run` method should return a `~lsst.pipe.base.Struct`.
+This allows named access to returned data, which provides safer evolution than relying on the order of returned values.
+All task methods that return more than one or two items of data should return the data in a `~lsst.pipe.base.Struct`.
+
+Many tasks are found in the ``pipe_tasks`` package, especially tasks that use many different packages and don't seem to belong in any one of them.
```
mrawls: Might be clearer to refer to `pipe_tasks` as `lsst.pipe.tasks` instead/in addition to match the `lsst.pipe.base` format. Eventually it would link to equivalent docs! (Same goes for referring to `ip_isr` right after.) Or perhaps the general term should be `module_name` and only python specific things should be `lsst.module.name`? I think you understand the goal of consistency I'm getting at here.

doc/lsst.pipe.base/command-line-task-data-repo-howto.rst
```diff
@@ -0,0 +1,136 @@
+.. _command-line-task-data-repo-howto:
+
+#################################################################
+Using Butler data repositories and reruns with command-line tasks
+#################################################################
+
+:ref:`Command-line tasks <using-command-line-tasks>` use Butler data repositories for reading and writing data.
+Butler data repositories are specially-configured directories on a filesystem.
```
mrawls: This simple statement could use more emphasis: **a repository is a specially-configured directory**. This may or may not be the place to go into what "special configurations" are necessary for a directory to be anointed as a proper repository; if it's not, a (future) link would be very helpful.

Along similar lines, I wonder if this section could be renamed "Using command-line tasks to process data" and then, at the top, explain more about how/why LSST uses Butler repositories for this (and sometimes reruns).

doc/lsst.pipe.base/command-line-task-data-repo-howto.rst
```diff
@@ -0,0 +1,136 @@
...
+
+The output repository for one task can become the input repository for the next command-line task.
+For example:
+
+.. code-block:: text
+
+   task2.py outputrepo --output outputrepo2 ...
+
+Because Butler data repositories are *chained*, the output repository (here, :file:`outputrepo2`) will appear to contain all the datasets from the input repositories (here: :file:`inputrepo`, :file:`outputrepo`, and :file:`outputrepo2` itself).
+
```
mrawls: "will appear to contain all the datasets from the input repositories" - this statement may be misleading because data files are not actually copied/linked from `inputrepo` to `outputrepo` and on down the chain. Instead there is some behind-the-scenes configuration magic that tells each repo who its parent is.

doc/lsst.pipe.base/command-line-task-data-repo-howto.rst
```diff
@@ -0,0 +1,136 @@
...
+An output repository can be the same as the input repository:
+
+.. code-block:: text
+
+   task3.py outputrepo2 --output outputrepo2 ...
+
+This pattern is useful for reducing the number of repository directories on a filesystem.
+Packing outputs from multiple tasks into one output repository does reduce your flexibility to run a task several times with different configurations and compare outputs, though.
+
+You can also run the same task multiple times with the same output repository.
```
mrawls: Can you really though? I suppose this may have been resolved with the implementation of DM-10340 and I just adjusted my workflow in the meantime to never do anything to a non-empty output repo.

doc/lsst.pipe.base/command-line-task-data-repo-howto.rst
```diff
@@ -0,0 +1,136 @@
...
+Be aware that the Science Pipelines will help you maintain the integrity of the processed data's provenance.
+If you change a task's configuration and re-run the task into the same output repository, an error "Config does not match existing task config" will be shown.
+
+How to use repository path environment variables
```
mrawls: Emphasize that these are optional; one doesn't need to have environment variables set to specify regular input and output paths.

doc/lsst.pipe.base/command-line-task-data-repo-howto.rst
```diff
@@ -0,0 +1,136 @@
...
+
+To use input data from a :file:`DATA` Butler repository and write outputs to a rerun called ``A``, set a command-line task's :option:`REPOPATH` and :option:`--rerun` like this:
+
+.. code-block:: bash
+
+   task1.py DATA --rerun A ...
+
+The rerun will be located inside the :file:`DATA` repository at :file:`DATA/rerun/A`.
+
+How to use one rerun as input to another (chaining)
```
mrawls: Please explicitly call this "rerun chaining" and not just "chaining," because the latter is reminiscent of how all repos have parents and are also "chained" in a sense.

mrawls: ... or is it the same chaining mechanism behind the scenes? I've never used reruns so I don't actually know.

doc/lsst.pipe.base/command-line-task-data-repo-howto.rst
ktlim: You definitely can, with the same configuration (or with `--clobber-config`) and with different dataIds (or if you don't care if you overwrite data).  This is especially useful with distributed jobs writing to a shared filesystem.

doc/lsst.pipe.base/command-line-task-data-repo-howto.rst
ktlim: Except that the statement is wrong.  A repository _is not_, in general, a directory.  A repository may be an object store, a database, or some other place where datasets can be stored.  Ideally, no one should ever have to know how to "anoint" a directory as a repository; that is a Butler implementation detail.

doc/lsst.pipe.base/command-line-task-data-repo-howto.rst
ktlim: I read "appear" as "appears to the Butler user, but not in the filesystem", but if you are assuming that repository == directory, then I can see how that is insufficient.

doc/lsst.pipe.base/command-line-task-data-repo-howto.rst
ktlim: It's the same mechanism.

doc/lsst.pipe.base/command-line-task-data-repo-howto.rst
RobertLuptonTheGood: To second KT, you should think of users as giving the butler a URI, that happens currently to be a directory.   The point of the butler is that it handles getting and saving data objects for you and the details are someone else's problem.

Incidentally, we need to explain somewhere the difference between the butler and the actual I/O;  I hear people complaining that "it's slow to read afwTables using the butler" -- they mean that the afwTable I/O code is slow, and that's nothing to do with the butler.

jonathansick: Excellent, this is something I wanted to bring up this was still unclear to me (I need to take another read through [LDM-463](https://ldm-463.lsst.io) and better yet, work with you on `lsst.daf.*` documentation!).

- Will the environment variable facility still work with non-POSIX repositories? Instead of a filesystem path the environment variables can be a general URI?
- Do we have a facility for deleting re-runs in non-POSIX repositories? The advice I've gotten is that folks `ls` inside the rerun directory and delete un-needed experimental reruns.

>  Ideally, no one should ever have to know how to "anoint" a directory as a repository; that is a Butler implementation detail.

Do we have a facility for creating a POSIX repository? Right now the standard pattern seems to be creating a directory and planting a `_mapper` in it. See this tutorial draft from another ticket: https://pipelines.lsst.io/v/DM-11391/getting-started/data-setup.html#creating-a-butler-repository-for-hsc-data

> Along similar lines, I wonder if this section could be renamed "Using command-line tasks to process data" and then, at the top, explain more about how/why LSST uses Butler repositories for this (and sometimes reruns).

I wanted to have the keywords/phrases "Butler data repository" and "rerun" in the title because those are keywords that (will) appear elsewhere in the documentation. To make that first connection between command-line tasks and data processing and Butler data repositories I think we might need to lean on tutorials and overview topics (that aren't in this ticket). I don't want to put too much discussion in how-to topics like this. I'm intending to write overview topics that I can link back to from the how-to.

doc/lsst.pipe.base/command-line-task-data-repo-howto.rst
jonathansick: Some `lsst.daf.persistance` user guide topics are in order!

The thing that's a little off-putting to me about rerun UX is that they're a facility for command-line tasks (you specify both the root repository and the rerun with a `--rerun` argument). But with the Butler Python API there's no concept of reruns and you need to construct the path to the rerun yourself. It's an uneasy asymmetry between command-line tasks and the Python API.

I think it'd be easier to explain if reruns were a first-class concept in the Butler API (or we drop reruns altogether, but I don't think that's likely!).

doc/lsst.pipe.base/command-line-task-data-repo-howto.rst
jonathansick: Chaining needs a topic in the `lsst.daf.persistence` user guide that I can link to here rather than explain it fully in this how-to.

doc/lsst.pipe.base/task-framework-overview.rst
jonathansick: I didn't write this page or the "creating" pages, but you're on to something

doc/lsst.pipe.base/command-line-task-data-repo-howto.rst
ktlim: >  Will the environment variable facility still work with non-POSIX repositories? Instead of a filesystem path the environment variables can be a general URI?

Certain specific URI forms are/will be supported, not any URI in general.

>  Do we have a facility for deleting re-runs in non-POSIX repositories? The advice I've gotten is that folks `ls` inside the rerun directory and delete un-needed experimental reruns.

We currently have no utilities for manipulating repositories. Subsetting, relocating, removing etc. all currently need to be done external to the Butler.  Some of these are conceptually simple but just haven't been done; others have been waiting on a comprehensive view of repository management (intended to be supported in part by a global metadata service).

Creating a repository should generally be done by executing a `CmdLineTask`.  The problem is two-fold: `CmdLineTask`s currently infer their output repository configuration from their input, so there's a question of how to create the first input; and `CmdLineTask` has no argument to allow specification of Butler configuration information.  I think this lack of progress on integrating `CmdLineTask` and Butler has come about because of reluctance to invest in `CmdLineTask` when it is soon to be replaced by Activators.

doc/lsst.pipe.base/command-line-task-data-repo-howto.rst
ktlim: From the Butler's perspective, `--rerun` is just a way of forming a repository root URL by combining two pieces.  It's still not clear to me that the API should privilege two-part naming with a particular fixed string interposed over any other kind of repository root naming scheme.

doc/lsst.pipe.base/command-line-task-data-repo-howto.rst
jonathansick: Thanks for the clarifications. I re-work the discussions around Butler repositories to ensure it's not biased towards one type of repository backend.

kfindeisen: @ktlim are there specific tickets for creating repository manipulation utilities? I recently asked on `#dm` how to copy a repository, and got a very directory-specific answer; I'd like to upgrade to a more generic approach once it exists.

doc/lsst.pipe.base/command-line-task-dataid-howto.rst
```diff
@@ -0,0 +1,203 @@
...
+This documentation also applies to the :option:`--selectId` argument for many-to-one processing tasks.
+
+.. _command-line-task-dataid-howto-about-dataid-keys:
+
+About data ID keys
+==================
+
+Individual data IDs are composed of a set of keys and their values.
+The set of keys that are included in a data ID depend on the dataset type and even the observatory package.
+The important categories of data IDs are for exposures (generally, individual observations from an observatory) and coadds.
```
mrawls: Exposures and Coadds mean very specific things in LSST-speak, and it would be good to add links to more comprehensive definitions here.

doc/lsst.pipe.base/command-line-task-config-howto.rst
```diff
@@ -0,0 +1,141 @@
...
+For these more complex configuration types you must use configuration files.
+
+.. _command-line-task-config-howto-configfile:
+
+How to use configuration files
+==============================
+
+You also provide configurations to a command-line task through a *configuration file*
+In fact, configuration files are Python modules; anything you can do in Python you can do in a configuration file.
+
```
mrawls: Missing a period for the first sentence.

This section would really benefit from an example configuration file text block, even if it's just a few lines long.

doc/lsst.pipe.base/command-line-task-retargeting-howto.rst
```diff
@@ -0,0 +1,131 @@
+.. _command-line-task-retargeting-howto:
+
+##########################################
+Retargeting subtasks of command-line tasks
+##########################################
+
+Subtasks of :ref:`command-line tasks <using-command-line-tasks>` are dynamically retargetable, meaning that you can configure what task class is run by a parent class.
+Subtask retargeting is a special case of :ref:`command-line task configuration <command-line-task-config-howto>`.
+
+A common use case of retargeting subtasks to use a camera-specific variant of a subtask.
```
mrawls: Weird sentence is weird.

doc/lsst.pipe.base/command-line-task-config-howto.rst
```diff
@@ -0,0 +1,141 @@
...
+Instead of an :option:`--output` argument you can specify a :option:`--rerun`, see :doc:`command-line-task-data-repo-howto`.
+
+.. _command-line-task-config-howto-show-subset:
+
+Viewing a subset of configurations
+----------------------------------
+
+To filter the :option:`--show config <--show>` output, include a search term with wildcard (``*``) characters.
+For example, this will show any configuration that includes the string ``calibration``:
+
```
mrawls: When we did `--show DATA` instead of `--show config`, you pointed out in a friendly box that this would put the task into dry-run mode. Is this the case for all tasks run with `--show`, or just ones where you feed `--show` a data repository? (The same question applies to the later instance of `--show tasks`.)

doc/lsst.pipe.base/command-line-task-retargeting-howto.rst
```diff
@@ -0,0 +1,131 @@
...
+   calibrate.astromRefObjLoader: lsst.meas.algorithms.loadIndexedReferenceObjects.LoadIndexedReferenceObjectsTask
+   calibrate.astrometry: lsst.meas.astrom.astrometry.AstrometryTask
+   calibrate.astrometry.matcher: lsst.meas.astrom.matchOptimisticB.matchOptimisticBContinued.MatchOptimisticBTask
+   ...
+
+This subtask hierarchy is interpreted as follows:
+
+- The ``calibrate`` subtask of the command line task (``processCcd.py``) configured to use `lsst.pipe.tasks.calibrate.CalibrateTask`.
+- ``calibrate`` (again, implemented by `~lsst.pipe.tasks.calibrate.CalibrateTask`) has a subtask named ``astrometry``, which is currently configured to use the `lsst.meas.astrom.astrometry.AstrometryTask` task.
+- ``calibrate.astrometry` has a subtasked named ``matcher``, which is implemented by `~lsst.meas.astrom.matchOptimisticB.matchOptimisticBContinued.MatchOptimisticBTask`.
```
mrawls: Mismatched ticks around `calibrate.astrometry` here.

doc/lsst.pipe.base/command-line-task-retargeting-howto.rst
```diff
@@ -0,0 +1,131 @@
...
+1. Import the Task class you intend to use.
+2. Assign that class to to the subtask configuration using the `~lsst.pex.config.ConfigurableField.retarget` method.
+
+For example, to retarget the subtask named ``configurableSubtask`` with a class ``FooTask``, this configuration file should contain:
+
+.. code-block:: python
+
+   from ... import FooTask
+   config.configurableSubtask.retarget(FooTask)
+
```
mrawls: I'm a little sad that the examples start getting more abstract here and are using "Foo" instead of actual LSST things.

After reading this whole section, it's also not clear to me how to tell whether something I want to retarget is a `ConfigurableField` or a `RegistryField` or what the distinction really means.

doc/lsst.pipe.base/command-line-task-logging-howto.rst
```diff
@@ -0,0 +1,83 @@
...
+Set the :option:`--longlog` argument to get a verbose logging message.
+This format is especially useful for parallel data processing since the data IDs of the data being processed are included in each message to let you later separate the processing streams.
+An example message looks like:
+
+.. code-block:: text
+
+   INFO  2017-07-18T04:40:14.014 processCcd.calibrate ({'taiObs': '2013-11-02', 'pointing': 671, 'visit': 904014, 'dateObs': '2013-11-02', 'filter': 'HSC-I', 'field': 'STRIPE82L', 'ccd': 1, 'expTime': 30.0})(calibrate.py:545)- Photometric zero-point: 30.674190
+
+Fields are:
+
```
mrawls: Bonus points if you can color-code or otherwise visually align each portion of the example message with the field definitions below.

doc/lsst.pipe.base/command-line-task-parallel-howto.rst
```diff
@@ -0,0 +1,97 @@
...
+
+###########################################
+Parallel processing with command-line tasks
+###########################################
+
+:ref:`Command-line tasks <using-command-line-tasks>` can operate in parallel.
+Generally if a command-line task is run against several :ref:`data IDs <command-line-task-dataid-howto>`, each data ID can be processed in parallel.
+Command-line tasks use two different categories of parallel processing: single-node parallel processing (:option:`-j`) and distributed processing (with the ``pipe_drivers`` and ``ctrl_pool`` packages).
+This page describes how to use these parallel processing features.
+
```
mrawls: A bit more about what situations might warrant single-node parallel processing vs. distributed processing would be nice here.

doc/lsst.pipe.base/command-line-task-parallel-howto.rst
```diff
@@ -0,0 +1,97 @@
...
+      This task does not support multiprocessing; using one process
+
+.. _command-line-task-parallel-howto-distributed:
+
+How to enable distributed processing
+====================================
+
+Command-line tasks implemented in the ``pipe_drivers`` package can distribute processing across multiple nodes using MPI or batch submission to clusters (PBS or SLURM).
+See the ``pipe_drivers`` documentation for more information.
+
```
mrawls: Looking forward to when this documentation exists :) I also noticed you mentioned `ctrl_pool` at the top as another option for distributed processing, but it is not mentioned here.

doc/lsst.pipe.base/command-line-task-argument-reference.rst
```diff
@@ -0,0 +1,353 @@
...
+.. _command-line-task-envvar:
+
+Environment variables
+=====================
+
+The :envvar:`PIPE_INPUT_ROOT`, :envvar:`PIPE_CALIB_ROOT`, and :envvar:`PIPE_OUTPUT_ROOT` environment variables let you more easily specify Butler data repositories.
+
+Each environment variable is used as a root directory for relative paths provided on the command line.
+If you set an absolute path on the command line, the environment variable is ignored.
+:ref:`see examples <command-line-task-envvar-examples>`.
```
mrawls: This reference isn't very useful since the examples are just a few lines down the page.

doc/lsst.pipe.base/creating-a-task.rst
```diff
@@ -0,0 +1,348 @@
+.. _creating-a-task:
```
mrawls: I get the feeling this and the following "Creating a command-line task" pages were copy-pasted from elsewhere and not written by you from scratch. The tone doesn't match the rest of the docs, they don't capitalize Butler everywhere, there is a "Documentation" section that simply says "to be written," and they feel a bit less accessible overall (for instance: "for now read existing tasks to see how it is done" re. how to use the Butler would be nicer as a not-yet-existing link to Butler documentation).

It's certainly better to have this content here than not, but these two pages would hugely benefit from feeling more like friendly tutorials and less like tired instruction manuals.

doc/lsst.pipe.base/command-line-task-data-repo-howto.rst
ktlim: I'm not aware of specific tickets at this time; @n8pease would know.  A trivial subsetting utility that works (terribly inefficiently) like a `CmdLineTask` that just copies input datasets to its output can be written in a few lines.  Relocation (without copying) and removal would require a bit more work.

doc/lsst.pipe.base/command-line-task-parallel-howto.rst
```diff
@@ -0,0 +1,97 @@
...
+   - OpenBLAS: ``OPENBLAS_NUM_THREADS``, ``GOTO_NUM_THREADS``, and ``OMP_NUM_THREADS``.
+   - MLK: ``MKL_NUM_THREADS``, ``MKL_DOMAIN_NUM_THREADS``, ``OMP_NUM_THREADS``.
+
+If necessary, you may bypass a command-line task's control over threading in low-level math libraries by setting the ``LSST_ALLOW_IMPLICIT_THREADS`` environment variable:
+
+.. code-block:: bash
+
+   export LSST_ALLOW_IMPLICIT_THREADS=1
+
```
RobertLuptonTheGood: While the preferred unix shell for scripting has been Bourne shell since unix began, people do use csh.  You should probably have a note on csh environment variable syntax somewhere.

doc/lsst.pipe.base/command-line-task-parallel-howto.rst
RobertLuptonTheGood: It might be worth mentioning here that you can disable this multithreading with `--batchType None`;  then people can at least get started before the documentation exists and Meredith is happy.

doc/lsst.pipe.base/command-line-task-parallel-howto.rst
```diff
@@ -0,0 +1,97 @@
...
+When parallel processing at a higher level (using :option:`-j`, for example), threading in these libraries can result in a **net decrease in performance** because of thread contention.
+
+When running with :option:`-j` multiprocessing or via ``ctrl_pool``, command-line tasks check if low-level libraries are using their default threading behavior (same number of threads as CPUs).
+If so, the task will automatically disable multi-threading in these libraries.
+The command-line task also sends a warning when it does so:
+
+.. code-block:: text
+
+   WARNING: You are using OpenBLAS with multiple threads (16), but have not specified the number of threads using one of the OpenBLAS environment variables: OPENBLAS_NUM_THREADS, GOTO_NUM_THREADS, OMP_NUM_THREADS. This may indicate that you are unintentionally using multiple threads, which may cause problems. WE HAVE THEREFORE DISABLED OpenBLAS THREADING. If you know what you are doing and want threads enabled implicitly, set the environment variable LSST_ALLOW_IMPLICIT_THREADS.
+
```
RobertLuptonTheGood: It might be worth mentioning that these considerations are independent of `pipe_drivers` use of MPI; we're talking multi-threading here (openMP)

doc/lsst.pipe.base/command-line-task-prov-howto.rst
```diff
@@ -0,0 +1,76 @@
...
+
+:ref:`Command-line tasks <using-command-line-tasks>` include provenance checks to ensure datasets in a Butler repository are processed with consistent software versions and task configurations.
+This page provides background on what these provenance checks are, and how they can be overridden when running command-line tasks.
+
+.. _command-line-task-prov-howto-config:
+
+How to override configuration checks with the :option:`--clobber-config` argument
+=================================================================================
+
+When a command-line task outputs datasets it also persists the task's configuration to the same Butler output repository.
```
RobertLuptonTheGood: Remove "same"

doc/lsst.pipe.base/command-line-task-prov-howto.rst
```diff
@@ -0,0 +1,76 @@
...
+If the same command-line task is later run on the same Butler output repository, that task will ensure that the configuration is consistent with the earlier run.
```
RobertLuptonTheGood: I think that "consistent" means "identical" here.  Please elucidate

doc/lsst.pipe.base/command-line-task-prov-howto.rst
```diff
@@ -0,0 +1,76 @@
...
+
+.. code-block:: bash
+
+   task.py REPOPATH --output output --id --clobber-config ...
+
+.. warning::
+
+   :option:`--clobber-config` is subject to race conditions because it is modifying a file in the Butler repository.
+   It should not be used in conjunction with parallel processing (such as :option:`-j`/:option:`--processes` or with "driver" command-line tasks).
+
```
RobertLuptonTheGood: As noted elsewhere these docs are inconsistent on whether `--clobber-config` is safe with `-j`

doc/lsst.pipe.base/command-line-task-prov-howto.rst
```diff
@@ -0,0 +1,76 @@
...
+   task.py REPOPATH --output output --id --clobber-versions ...
+
+Like :option:`--clobber-config`, the :option:`--clobber-versions` argument replaces software version provenance information in the Butler data repository with current version information.
+This argument is useful if you wish to continue using an existing Butler data repository with a different version of the Science Pipelines, and are confident that there are not dataset incompatibilities.
+
+.. warning::
+
+   :option:`--clobber-versions` is subject to race conditions because it is modifying a file in the Butler repository.
+   It should not be used in conjunction with parallel processing (such as :option:`-j`/:option:`--processes` or with "driver" command-line tasks).
+
```
RobertLuptonTheGood: I think you can safely run a non-existent data ID with `-j 1 --clobber-config` and then proceed with the new config and a real data ID.

It'd be good to provide a less tacky way to do this.  Or maybe simply omitting the `--id` works (or is that taken to mean all IDs as `--id` is documented to mean?  If so, it's another reason to remove that implicit globbing behaviour.

doc/lsst.pipe.base/command-line-task-retargeting-howto.rst
```diff
@@ -0,0 +1,132 @@
+.. _command-line-task-retargeting-howto:
+
+##########################################
+Retargeting subtasks of command-line tasks
+##########################################
+
+Subtasks of :ref:`command-line tasks <using-command-line-tasks>` are dynamically retargetable, meaning that you can configure what task class is run by a parent class.
```
RobertLuptonTheGood: "which task class"

doc/lsst.pipe.base/command-line-task-retargeting-howto.rst
```diff
@@ -0,0 +1,132 @@
...
+
+A common use of retargeting is to change a default subtask for a camera-specific one.
+For example:
+
+- `lsst.obs.subaru.isr.SuprimeCamIsrTask`: a version of instrument signature removal (ISR or detrending) for Suprime-Cam and Hyper Suprime-Cam.
+- `lsst.obs.sdss.selectSdssImages.SelectSdssImagesTask`: an version of the task that selects images for co-addition of SDSS stripe 82 images.
+
+How you retarget a subtask and override its config parameters depends on whether the task is specified as an `lsst.pex.config.ConfigurableField` (the most common case) or as an `lsst.pex.config.RegistryField`.
+Both scenarios are described on this page.
+
```
RobertLuptonTheGood: Please describe why you'd use one and when the other, what the tradeoffs are, etc.

doc/lsst.pipe.base/command-line-task-retargeting-howto.rst
```diff
@@ -0,0 +1,132 @@
...
+Once a subtask is retargeted, you can configure it as normal by setting attributes to configuration parameters.
+For example:
+
+.. code-block:: python
+
+   config.configurableSubtask.subtaskParam1 = newValue
+
+.. warning::
+
+   When you retarget a task specified by an `lsst.pex.config.ConfigurableField` you lose all :ref:`configuration overrides <command-line-task-config-howto-obs>` for both the old and new task.
```
RobertLuptonTheGood: I don't know what this means.  Please add a bit more context/detail.

doc/lsst.pipe.base/command-line-task-retargeting-howto.rst
```diff
@@ -0,0 +1,132 @@
...
+   This limitation is not shared by `lsst.pex.config.RegistryField`.
+
+.. _command-line-task-retargeting-howto-registry-configfile:
+
+How to retarget a subtask configured as a RegistryField with a configuration file
+=================================================================================
+
```
RobertLuptonTheGood: Please explain what a registry is, and how they are unrelated to data registries!

doc/lsst.pipe.base/command-line-task-retargeting-howto.rst
```diff
@@ -0,0 +1,132 @@
...
+To retarget a subtask specified as an `lsst.pex.config.RegistryField`, set the field's `~lsst.pex.config.RegistryField.name` attribute in a configuration file (using :option:`--configfile`).
+Here is an example that assumes a task ``FooTask`` is defined in module :file:`.../foo.py` and registered using name ``foo``:
+
```
RobertLuptonTheGood: Explain what the name means

doc/lsst.pipe.base/command-line-task-retargeting-howto.rst
```diff
@@ -0,0 +1,132 @@
...
+   --config registrySubtask.active.subtaskParam1=newValue
+
+.. _command-line-task-retargeting-howto-registry-config:
+
+Configure any subtask in a registry
+-----------------------------------
+
+Alternatively, you can then configure parameters for any subtask in the registry using key-value access.
+For example:
+
```
RobertLuptonTheGood: Which do we recommend?  Why? Should the other be deprecated?

doc/lsst.pipe.base/creating-a-command-line-task.rst
```diff
@@ -0,0 +1,253 @@
...
+A command-line task is an enhanced version of a regular task (see :ref:`creating-a-task`).
+Regular tasks are only intended to be used as relatively self-contained stages in data processing pipelines, whereas command-line tasks can also be used as complete pipelines.
+As such, command-line tasks include :ref:`run scripts <creating-a-command-line-task-run-script>` that run them as pipelines.
+
+Command-line tasks have the following key attributes, in addition to the attributes for regular tasks:
+
+- They are subclasses of `lsst.pipe.base.CmdLineTask`, whereas regular tasks are subclasses of `lsst.pipe.base.Task`.
+- They have an associated :ref:`run script <creating-a-command-line-task-run-script>` to run them from the command-line as pipelines (this is common, but not required, for regular tasks).
+- They have a ``run`` method which performs the full pipeline data processing.
+- By default the ``run`` method takes exactly one argument: a data reference for the item of data to be processed.
```
RobertLuptonTheGood: This is wrong (although commonly adhered to, and the rules were not clearly laid down).

A `run` method *may not* take a dataRef; a `runDataref` must be provided to unpack the dataRef and pass the arguments to the `run` method (there are some exceptions for things like reference catalogues where a proxy may be passed).

Because this rule is often/usually ignored I think we have to be a bit careful here.  The docs should describe the correct situation, and tickets filed to fix all offending tasks.

doc/lsst.pipe.base/creating-a-command-line-task.rst
```diff
@@ -0,0 +1,253 @@
...
+
+   #!/usr/bin/env python2
+   from lsst.pipe.tasks.exampleCmdLineTask import ExampleCmdLineTask
+   ExampleCmdLineTask.parseAndRun()
+
+For most command-line tasks you should put the run script into your package's :file:`bin/` directory, so that it is on your ``$PATH`` when you setup your package with eups.
+We did not want the run script for ``ExampleCmdLineTask`` to be quite so accessible, so we placed it in the :file:`examples/` directory instead of :file:`bin/`.
+
+Remember to make your run script executable using :command:`chmod +x`.
+
```
RobertLuptonTheGood: Here's the place to remind readers that `parseAndRun()` will *not* ordinarily return control to the bin script (but see `--noExit` [or whatever I called that option])

doc/lsst.pipe.base/creating-a-command-line-task.rst
```diff
@@ -0,0 +1,253 @@
...
+.. _creating-a-command-line-task-data-io:
+
+Reading and Writing Data
+========================
+
+The :ref:`run method <creating-a-command-line-task-intro>` typically receives a single data reference, as mentioned above.
+It read and writes data using this data reference (or the underlying butler, if necessary).
```
RobertLuptonTheGood: As above, no.  This is not the correct behaviour (although very common).

doc/lsst.pipe.base/creating-a-command-line-task.rst
```diff
@@ -0,0 +1,253 @@
...
+Every time you write a task that writes a new kind of data (a new "dataset type") you must tell the butler about it.
+Similarly, if you write a new task for which you want to save configuration and metadata (which is the case for most tasks that process data), you have to tell the butler about it.
+
+To add a dataset, edit the mapper configuration file for each observatory package on whose data the task can be run.
+If the task is of general interest (wanted for most or all observatory packages) then this process of updating all the mapper configuration files can be time consuming.
+
+There are plans to change how mappers are configured.
+But as of this writing, mapper configuration files are contained in the policy directory of the observatory package.
+For instance the configuration for the lsstSim mapper is defined in ``obs_lsstSim/policy/LsstSimMapper.paf``.
+
```
RobertLuptonTheGood: Document `obs_base`

doc/lsst.pipe.base/creating-a-command-line-task.rst
```diff
@@ -0,0 +1,253 @@
...
+For instance in the case of the `pipe.tasks.makeSkyMap.MakeSkyMapTask` and various co-addition tasks, the co-add type must be part of the config and metadata dataset type name.
+To customize the dataset type of a task's config or metadata, define task methods ``_getConfigName`` and ``_getMetadataName`` to return the desired names.
+
+.. _creating-a-command-line-task-prevent-saving-config:
+
+Prevent Saving Config and Metadata
+----------------------------------
+
+For some tasks you may wish to not save config and metadata at all.
+This is appropriate for tasks that simply report information without saving data.
```
RobertLuptonTheGood: What sort of tasks?  And how do they return information without saving metadata -- it sounds as if these are exactly the tasks that *need* to write metadata.

This may require an extension to the metadata system.

doc/lsst.pipe.base/creating-a-command-line-task.rst
```diff
@@ -0,0 +1,253 @@
...
+To disable saving configuration and metadata, define task methods ``_getConfigName`` and ``_getMetadataName`` methods to return `None`.
+
+.. _creating-a-command-line-task-custom-argparse:
+
+Custom Argument Parser
+======================
+
```
RobertLuptonTheGood: Document that this is all built on python's `argparse`

doc/lsst.pipe.base/creating-a-command-line-task.rst
```diff
@@ -0,0 +1,253 @@
...
+  - The second argument is a dataset type, which specifies the keys that are used with this ID argument.
+    The keys associated with a particular dataset type are specified in the mapper configuration file for the observatory package and camera in question, and thus may vary from camera to camera.
+    In practice, the keys for ``raw`` and ``calexp`` dataset types usually do vary from camera to camera, but the keys for coadds do not.
+    However, this is not a fixed rule.
+    For most observatory packages ``deepCoadd`` is one of two coadd dataset types, and the other, ``goodSeeingCoadd``, would work just as well for this argument.
+  - A custom ``ContainerClass`` (for example, `lsst.coadd.utils.coaddDataIdContainer.CoaddDataIdContainer`) is provided to support iterating over missing keys (e.g. if you provide a tract but not a patch then the task will iterate over all available patches for that tract).
+    This happens automatically for ``raw`` and ``calexp`` dataset types, but not most other dataset types.
+    Examine the code in `~lsst.coadd.utils.coaddDataIdContainer.CoaddDataIdContainer` to see how it works.
```
RobertLuptonTheGood: Should they be reading this code or should we fix/generalise it?  I think the latter.

doc/lsst.pipe.base/creating-a-command-line-task.rst
```diff
@@ -0,0 +1,253 @@
...
+         """
+         parser = pipeBase.ArgumentParser(name=cls._DefaultName)
+         parser.add_id_argument("--id", "deepCoadd", help="data ID, e.g. --id tract=12345 patch=1,2",
+                                ContainerClass=CoaddDataIdContainer)
+         parser.add_id_argument("--selectId", "calexp", help="data ID, e.g. --selectId visit=6789 ccd=0..9",
+                                ContainerClass=SelectDataIdContainer)
+         return parser
+
+  In this case the custom container class `~lsst.pipe.tasks.coaddBase.SelectDataIdContainer` adds additional information for the task, to save processing time.
+
```
RobertLuptonTheGood: Please explain how it saves processing time

doc/lsst.pipe.base/creating-a-command-line-task.rst
```diff
@@ -0,0 +1,253 @@
...
+             if self.doRaise:
+                 results = task.run(butler)
+             else:
+                 try:
+                     results = task.run(butler)
+                 except Exception as e:
+                     task.log.fatal("Failed: %s" % e)
+                     if not isinstance(e, pipeBase.TaskError):
+                         traceback.print_exc(file=sys.stderr)
+             task.writeMetadata(butler)
```
RobertLuptonTheGood: Is this up-to-date with the boiler-plate changes needed to return correct exit codes to the shell?

doc/lsst.pipe.base/creating-a-task.rst
```diff
@@ -0,0 +1,350 @@
...
+Tasks typically have a ``run`` method that executes the task's main function.
+See :ref:`creating-a-task-class-methods` for more information.
+
+.. _creating-a-task-configuration:
+
+Configuration
+=============
+
+Every task requires a configuration: a task-specific set of configuration parameters.
+The configuration is read-only; once you construct a task, the same configuration will be used to process all data with that task.
```
RobertLuptonTheGood: Is the read-only enforced?

doc/lsst.pipe.base/creating-a-task.rst
```diff
@@ -0,0 +1,350 @@
...
+This makes the data processing more predictable: it does not depend on the order in which items of data are processed.
```
RobertLuptonTheGood: The order of configuration of multiple tasks is also critical, as e.g. tasks can add to an `afwTask.schema` as they are configured;  all of this must be done before the first task is processed.  This provides another reason why the Tasks are first created and configured, and only then run.

doc/lsst.pipe.base/creating-a-task.rst
```diff
@@ -0,0 +1,350 @@
...
+
+Some important details of configurations:
+
+- Supply useful defaults for all config parameters if at all possible.
+  Your task will be much easier to use if the default configuration usually suffices.
+
+- Document each field of the configuration carefully.
+  Pretend you don't know anything about your task and ask yourself what you would need to be told to change the parameter.
+  What does the parameter do and why might you change it?
+  What units does it have?
```
RobertLuptonTheGood: These are the documentation strings that `--show config` prints

doc/lsst.pipe.base/creating-a-task.rst
```diff
@@ -0,0 +1,350 @@
...
+
+- Subtasks are specified in your configuration as fields of type `lsst.pex.config.ConfigurableField` or (less commonly) `lsst.pex.config.RegistryField`.
+  This allows subtasks to be :doc:`retargeted <command-line-task-retargeting-howto>` (replaced with a variant subtask).
```
RobertLuptonTheGood: If one is deprecated, mention that here.

doc/lsst.pipe.base/creating-a-task.rst
```diff
@@ -0,0 +1,350 @@
...
+             dtype=str,
+             doc="Mask planes that, if set, indicate the associated pixel should "
+             "not be included when the calculating statistics.",
+             default=("EDGE",),
+         )
+         numSigmaClip = pexConfig.Field(
+             doc="number of sigmas at which to clip data",
+             dtype=float,
+             default=3.0,
+         )
```
RobertLuptonTheGood: I'm surprised that this doesn't enforce `numSigmaClip > 0` as part of its definition

doc/lsst.pipe.base/creating-a-task.rst
```diff
@@ -0,0 +1,350 @@
...
+
+Class variables
+===============
+
+Tasks require several class variables to function:
+
+- ``ConfigClass``: the :ref:`configuration class <creating-a-task-configuration>` used by the task.
+
+- ``_DefaultName``: a string used as the default name for the task.
+  This is required for a command-line task (`~pipe.base.cmdLineTask.CmdLineTask`), and strongly recommended for other tasks because it makes them easier to construct for unit tests.
```
RobertLuptonTheGood: Why is it not required?  Here's an example of a place that while writing the docs we should also file the tickets to fix obvious problems.  If there is a reason, on the other hand, it needs to be documented here.

doc/lsst.pipe.base/creating-a-task.rst
```diff
@@ -0,0 +1,350 @@
...
+
+.. _creating-a-task-class-methods:
+
+Methods
+=======
+
+Tasks have the following important methods:
+
+- :ref:`__init__ <creating-a-task-class-init-method>`: construct and initialize a task.
+- :ref:`run <creating-a-task-class-run-method>`: process one item of data.
```
RobertLuptonTheGood: Again, this is a very common error.

doc/lsst.pipe.base/creating-a-task.rst
```diff
@@ -0,0 +1,350 @@
...
+.. _creating-a-task-class-init-method:
+
+The ``__init__`` method
+-----------------------
+
+Use the ``__init__`` method (task constructor) to do the following:
+
+- Call the parent task's ``__init__`` method
+- Make subtasks by calling ``self.makeSubtask(name)``, where ``name`` is the name of a field of type `lsst.pex.config.ConfigurableField` in your :ref:`task's configuration <creating-a-task-configuration>`.
+- Make a schema if your task uses an `lsst.afw.table`.
```
RobertLuptonTheGood: Normally one would be passed in, and your job is to augment it

doc/lsst.pipe.base/creating-a-task.rst
```diff
@@ -0,0 +1,350 @@
...
+.. code-block:: python
+
+   def __init__(self, *args, **kwargs):
+       """!Construct an ExampleSigmaClippedStatsTask
+       The init method may compute anything that that does not require data.
+       In this case we create a statistics control object using the config
+       (which cannot change once the task is created).
+       """
+       pipeBase.Task.__init__(self, *args, **kwargs)
+       self._badPixelMask = MaskU.getPlaneBitMask(self.config.badMaskPlanes)
```
RobertLuptonTheGood: `MaskU` can and should now be written `Mask` (and if you don't like that, `MaskX` will at least run).

I don't quite know how JSick would know that -- how do we keep this sort of thing in sync?  The example as written won't run.

doc/lsst.pipe.base/creating-a-task.rst
```diff
@@ -0,0 +1,350 @@
...
+--------------
+
+Most tasks have a ``run`` method which perform the task's data processing operation.
+This is required for command-line tasks and strongly recommended for most other tasks.
+One exception is if your task needs different methods to handle different data types (C++ handles this using overloaded functions, but the standard technique is Python is to provide different methods for different call signatures).
+
+If your task's processing can be divided into logical units, then we recommend that you provide methods for each unit. ``run`` can then call each method to do its work.
+This allows your task to be more easily adapted: a subclass can override just a few methods.
+
+We strongly recommend that you make your task stateless, by not using instance variables as part of your data processing. Pass data between methods by calling and returning it.
```
RobertLuptonTheGood: Is this not an absolute rule?

doc/lsst.pipe.base/creating-a-task.rst
```diff
@@ -0,0 +1,350 @@
...
+This makes the task much easier to reason about, since processing one item of data cannot affect future items of data.
+
+The ``run`` method should always return its results in an `lsst.pipe.base.struct.Struct` object, with a named field for each item of data.
```
RobertLuptonTheGood: "must return its results", I think

doc/lsst.pipe.base/creating-a-task.rst
```diff
@@ -0,0 +1,350 @@
...
+Both of these are constants, and thus are the same for each invocation of the ``run`` method; this is strongly recommended, as explained in the next section.
+
+.. _creating-a-task-class-run-method:
+
+The run method
+--------------
+
+Most tasks have a ``run`` method which perform the task's data processing operation.
+This is required for command-line tasks and strongly recommended for most other tasks.
+One exception is if your task needs different methods to handle different data types (C++ handles this using overloaded functions, but the standard technique is Python is to provide different methods for different call signatures).
```
RobertLuptonTheGood: I think that we decided that this wasn't acceptable except under very special circumstances.  All tasks must have a run method, and it must take a list of variables that it uses (by name, not as a dict or other blob such as a dataRef).

While this rule has not been applied much for cmdLineTasks it has been almost universally obeyed for Tasks

doc/lsst.pipe.base/creating-a-task.rst
```diff
@@ -0,0 +1,350 @@
...
+       # Unpersist the raw exposure pointed to by the data reference
+       rawExp = dataRef.get("raw")
+       maskedImage = rawExp.getMaskedImage()
+       # Support extra debug output.
+       # -
+       import lsstDebug
+       display = lsstDebug.Info(__name__).display
+       if display:
+           frame = 1
+           mtv(rawExp, frame=frame, title="exposure")
```
RobertLuptonTheGood: This needs some afwDisplay boilerplate; one option is to import `afwDisplay` and `afwDisplay.Display.mtv(...)`

doc/lsst.pipe.base/creating-a-task.rst
```diff
@@ -0,0 +1,350 @@
...
+       # return the pipe_base Struct that is returned by self.stats.run
+       return self.stats.run(maskedImage)
+
+The statistics are actually computed by the ``stats`` subtask.
+Here is the ``run`` method for the default version of that task: ``exampleTask.ExampleSigmaClippedStatsTask.run``:
+
+.. code-block:: python
+
+   @pipeBase.timeMethod
+   def run(self, maskedImage):
```
RobertLuptonTheGood: Note that this run method obeys the rules!

doc/lsst.pipe.base/creating-a-task.rst
```diff
@@ -0,0 +1,350 @@
...
+Documentation
+=============
+
+To be written.
+
+.. _creating-a-task-subtasks:
+
+Subtasks
+========
+
```
RobertLuptonTheGood: This discussion wasn't clear (to me!) when the two methods were introduced.   I still think we need to decide how we are doing this and deprecate one version (probably `lsst.pex.config.ConfigurableField`, but I have no strong opinion)

doc/lsst.pipe.base/creating-a-task.rst
```diff
@@ -0,0 +1,350 @@
...
+    .. code-block:: python
+
+       config.registrySubtask[nameOfSelectedSubtask].subtaskParam1 = ...
+
+    Or use the `active` attribute to modify the configuration of the currently selected (active) subtask:
+
+    .. code-block:: python
+
+       config.registrySubtask.active.subtaskParam1 = ...
```
RobertLuptonTheGood: I don't understand this.  Is this basically an alias?  If so, how do we know that all subtasks have the same parameters?  It seems like a dangerous/confusing piece of syntactic sugar.

doc/lsst.pipe.base/creating-a-task.rst
```diff
@@ -0,0 +1,350 @@
...
+
+- `lsst.pex.config.RegistryField` advantages:
+
+  - You can specify overrides for any registered subtask and they are remembered if you retarget subtasks.
+    In comparison if the subtask is specified as an `lsst.pex.config.ConfigurableField` then you can only override parameters for the currently retargeted subtask, and all overrides are lost each time you retarget.
+    Thus using an `lsst.pex.config.RegistryField` offers the opportunity to specify suitable overrides for more than one variant subtask, making it safer for the user to use those variants.
+    Of course this can get out of hand if there are many variants, so users should not assume that all variants have suitable overrides.
+
+  - Retargeting a subtask can be done using :option:`--config` on the command line, as long as the module containing the desired subtask has been imported:
+
```
RobertLuptonTheGood: How common is this?

doc/lsst.pipe.base/creating-a-task.rst
```diff
@@ -0,0 +1,350 @@
...
+       from ... import FooTask
+       config.configurableSubtask.retarget(FooTask)
+
+  - Variants subtasks are kept together in one registry, making it easier to find them.
+
+Our recommendation: if you anticipate that users will often wish to override the subtask with a variant, then use an `lsst.pex.config.RegistryField`.
+Otherwise use an `lsst.pex.config.ConfigurableField` to keep config overrides simpler and easier to read.
+
+For example PSF determiners and star selectors are perhaps best specified using `lsst.pex.config.RegistryField` because there are several variants users may wish to select from.
```
RobertLuptonTheGood: I can think of ways to add syntactic sugar to `ConfigurableField` to allow command-line selection (basically another level of indirection).

doc/lsst.pipe.base/task-framework-overview.rst
```diff
@@ -0,0 +1,49 @@
...
+
+.. _task-framework-overview:
+
+##############################
+Overview of the task framework
+##############################
+
+:doc:`lsst.pipe.base <index>` provides a data processing pipeline infrastructure.
+Data processing is performed by **tasks**, which are instances of `lsst.pipe.base.Task` or `lsst.pipe.base.CmdLineTask`.
+Tasks perform a wide range of data processing operations, from basic operations such as assembling raw images into CCD images (trimming overscan), fitting a WCS or detecting sources on an image, to complex combinations of such operations.
```
RobertLuptonTheGood: You have to do more than just trim overscan -- there's no way to avoid estimating its level too.  This isn't the point you're trying to make, but the parenthetical comment is confusing.  I think it'd be better to use the full ISR as the example.

doc/lsst.pipe.base/task-framework-overview.rst
```diff
@@ -0,0 +1,49 @@
...
+
+**Tasks** are hierarchical.
+Each task may may call other tasks to perform some of its data processing; we say that a **parent task** calls a **subtask**.
+The highest-level task is called the **top-level task**.
+To call a subtask, the parent task constructs the subtask and then calls methods on it.
+Thus data transfer between tasks is simply a matter of passing the data as arguments to function calls.
```
RobertLuptonTheGood: Tasks are the fundamental level of code reuse, and we should be designing things that way.  This is _why_ the run method is required to take objects, not dataRefs as programmers usually has the former lying around but not the latter.

Think about the reuse of tasks from Notebooks -- all tasks should be usable that way.   Some tasks written as `CmdLineTask` (e.g. ISRTask) are useful from Notebooks too, which is why the Task and CmdLineTask rules are the same even though the latter, when run from the command line, happen to have a `dataRef` available.

doc/lsst.pipe.base/task-framework-overview.rst
```diff
@@ -0,0 +1,49 @@
...
+
+**Command-line tasks** are tasks that can be run from the command line.
+You might think of them as the LSST equivalent of a data processing pipeline.
+Despite their extra capabilities, command-line tasks can also be used as ordinary tasks and called as subtasks by other tasks.
```
RobertLuptonTheGood: Not quite.  They are tasks that can _also_ be run from the command line.  You sort-of say this, but this statement is stronger -- they are tasks first and foremost.

doc/lsst.pipe.base/command-line-task-argument-reference.rst
```diff
@@ -0,0 +1,362 @@
...
+   **Input Butler data repository URI or path.**
+
+   The input Butler data repository is always the first argument to a command-line task.
+   This argument is required for all command-line task runs, except when printing help (:option:`--help`).
+
+   In general, this can can be a URI that dependends on the Butler backend.
+   For example, ``swift://host/path`` for a Swift backend or ``file://path`` for a POSIX backend.
+
+   For POSIX backends, this may also be an absolute file path or a path relative to the current working directory.
+
```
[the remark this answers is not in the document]
jonathansick: I agree. I've created DM-11671 for a Butler user guide and I'm adding TODO notes to refactor these docs and link to Butler docs.

doc/lsst.pipe.base/command-line-task-data-repo-howto.rst
```diff
@@ -0,0 +1,162 @@
...
+
+Tasks can write to an existing rerun.
+For example, if rerun ``B`` was already created you can write additional outputs to it:
+
+.. code-block:: bash
+
+   task3.py DATA --rerun B ...
+
+Because reruns are chained, the Butler will start looking for datasets in this rerun ``B``, then in the chained ``A`` rerun, all the way to the root data repository (:file:`DATA`).
+
```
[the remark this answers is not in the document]
kfindeisen: Is it possible for two datasets with the same identifier to exist in `B` and `A`? In other words, is @jonathansick's more detailed explanation needed to understand how datasets are disambiguated?

doc/lsst.pipe.base/command-line-task-argument-reference.rst
```diff
@@ -0,0 +1,362 @@
...
+.. option:: --longlog
+
+   **Enable the verbose logging format.**
+
+   See :ref:`command-line-task-logging-howto-longlog` for more information.
+
+.. option:: --debug
+
+   **Enable debugging mode.**
+
```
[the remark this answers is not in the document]
jonathansick: I've added a note to cross link to these docs when they're ported in DM-11675.

doc/lsst.pipe.base/command-line-task-argument-reference.rst
```diff
@@ -0,0 +1,362 @@
...
+   **Backup and overwrite existing config files.**
+
+   Normally a command-line task checks existing config files in a Butler repository to ensure that the current configurations are consistent with previous pipeline executions.
+   This argument disables this check, which may be useful for development.
+
+   This argument is safe with :option:`-j` multiprocessing, but not necessarily with other forms of parallel execution.
+
+   See :ref:`command-line-task-prov-howto-config` for more information.
+
```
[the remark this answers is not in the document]
jonathansick: Fixed this in the linked how-to topic. ``clobber-config/clobber-versions`` are safe with `-j`.

doc/lsst.pipe.base/command-line-task-parallel-howto.rst
```diff
@@ -0,0 +1,97 @@
...
+You can change this timeout via the :option:`--timeout` command-line argument.
+
+Log messages from each process are combined into the same console output stream.
+To discern logging statements from each process use the :option:`--longlog` argument to enable a verbose logging format that includes the data ID being processed.
+See :ref:`command-line-task-logging-howto-longlog` for more information.
+
+.. note::
+
+   Internally, :option:`-j`-type multiprocessing is implemented with the `multiprocessing` module in the Python standard library.
+
```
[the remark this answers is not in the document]
jonathansick: It does. I've documented that with the `-j` argument reference.

doc/lsst.pipe.base/command-line-task-parallel-howto.rst
jonathansick: @RobertLuptonTheGood I'm not following. Are you saying the intoducing paragraph is wrong:

> When running with :option:`-j` multiprocessing or via ``ctrl_pool``, command-line tasks check if low-level libraries are using their default threading behavior (same number of threads as CPUs).
If so, the task will automatically disable multi-threading in these libraries.
The command-line task also sends a warning when it does so:

doc/lsst.pipe.base/command-line-task-dataid-howto.rst
```diff
@@ -0,0 +1,203 @@
...
+
+How to find data IDs available to a command-line task
+=====================================================
+
+Given an input Butler data repository at :option:`REPOPATH`, and optionally :ref:`a rerun <command-line-task-data-repo-howto>` given a :option:`--rerun` argument, you can discover available data and their data IDs with the :option:`--show DATA <--show>` command-line task argument:
+
+.. code-block:: bash
+
+   task.py REPOPATH --output output --id --show DATA ...
+
```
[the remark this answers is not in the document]
jonathansick: In my experience an `--id` argument (wildcard or not) is needed to make `--show data` work. Same for needing an output repo/rerun.

jonathansick:
**Thank you, everyone, for your helpful comments and edits.**

I've incorporated comments and created follow up tickets for additional content development, especially for topics that are essentially ports of existing Doxygen content that clearly do need more work. The [JIRA ticket|https://jira.lsstcorp.org/browse/DM-11253] links to these follow-ons.

PR #38, opened 2017-07-20
Exit to the shell with the number of failed dataRefs

RobertLuptonTheGood:
Can be overridden using --noExit

### DIFF
--- a/python/lsst/pipe/base/argumentParser.py
+++ b/python/lsst/pipe/base/argumentParser.py
@@ -347,6 +347,8 @@ class ArgumentParser(argparse.ArgumentParser):
         self.add_argument("--debug", action="store_true", help="enable debugging output?")
         self.add_argument("--doraise", action="store_true",
                           help="raise an exception on error (else log a message and continue)?")
+        self.add_argument("--noExit", action="store_true",
+                          help="Do not exit even upon failure (i.e. return a struct to the calling script)")
         self.add_argument("--profile", help="Dump cProfile statistics to filename")
         self.add_argument("--show", nargs="+", default=(),
                           help="display the specified information to stdout and quit "
--- a/python/lsst/pipe/base/cmdLineTask.py
+++ b/python/lsst/pipe/base/cmdLineTask.py
@@ -160,6 +160,10 @@ class TaskRunner(object):
             fail when using multiprocessing if the returned data cannot be
             pickled.
 
+            Note that even if doReturnResults is False a struct with a single
+            member "exitStatus" is returned, with value 0 or 1 to be returned
+            to the unix shell.
+
         @throws ImportError if multiprocessing requested (and the task
             supports it) but the multiprocessing library cannot be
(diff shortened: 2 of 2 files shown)


python/lsst/pipe/base/cmdLineTask.py
```diff
@@ -369,12 +373,16 @@ def __call__(self, args):
             self.log.MDC("LABEL", str([ref.dataId for ref in dataRef if hasattr(ref, "dataId")]))
         task = self.makeTask(args=args)
         result = None  # in case the task fails
+        exitStatus = 0 # exit status for the shell
```
PaulPrice: Picky: code standards say two spaces between the functional end of the line and the comment.

python/lsst/pipe/base/cmdLineTask.py
```diff
@@ -369,12 +373,16 @@ def __call__(self, args):
...
         if self.doRaise:
             result = task.run(dataRef, **kwargs)
         else:
             try:
                 result = task.run(dataRef, **kwargs)
             except Exception as e:
+                exitStatus = 1          # n.b. The shell exit value is the number of dataRefs returning
+                                        # non-zero, so the actual value used here is lost
```
PaulPrice: Not sure what code standards say about trailing multiline comments, but I'm gonna guess they want two spaces between the end of the functional part and the start of the comment.

python/lsst/pipe/base/cmdLineTask.py
```diff
@@ -512,6 +529,19 @@ def parseAndRun(cls, args=None, config=None, log=None, doReturnResults=False):
 
         taskRunner = cls.RunnerClass(TaskClass=cls, parsedCmd=parsedCmd, doReturnResults=doReturnResults)
         resultList = taskRunner.run(parsedCmd)
+
+        try:
+            nFailed = sum(((res.exitStatus != 0) for res in resultList))
+        except Exception as e:
+            parsedCmd.log.warn("Unable to retrieve exit status (%s); assuming success" % e)
```
PaulPrice: Probably not important in this context, but in general you shouldn't use `%` to expand the string but separate with commas:

    parsedCmd.log.warn("Unable to retrieve exit status (%s); assuming success", e)

(Another below).

python/lsst/pipe/base/argumentParser.py
```diff
@@ -347,6 +347,8 @@ def __init__(self, name, usage="%(prog)s input [options]", **kwargs):
         self.add_argument("--debug", action="store_true", help="enable debugging output?")
         self.add_argument("--doraise", action="store_true",
                           help="raise an exception on error (else log a message and continue)?")
+        self.add_argument("--noExit", action="store_true",
+                          help="Do not exit even upon failure (i.e. return a struct to the bin script)")
```
PaulPrice: Maybe don't assume that the caller's a "bin script"?

PR #43, opened 2017-10-25
DM-11931 Cleanup exitStatus handling

parejkoj:


python/lsst/pipe/base/cmdLineTask.py
```diff
@@ -374,10 +374,11 @@ def __call__(self, args):
             - ``dataRef``: the provided data reference.
             - ``metadata``: task metadata after execution of run.
             - ``result``: result returned by task run, or `None` if the task fails.
+            - ``exitStatus`: 0 if the task completed successfully, 1 otherwise.
 
             If ``doReturnResults`` is `False` the struct contains:
 
-            - ``exitStatus``.
+            - ``exitStatus`: 0 if the task completed successfully, 1 otherwise.
```
hsinfang: Reading [DM-4141 here](https://jira.lsstcorp.org/browse/DM-4141?focusedCommentId=81470&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-81470 )  it's not "1 otherwise" but  "a non-zero value the number of failed dataRefs is passed".   Do I read that correctly?

parejkoj: For `__call__`, I believe my statement is correct (at least, it matches what the code in that method does). `parseAndRun` is what does what you say.

@RobertLuptonTheGood can confirm that this was his intention.

RobertLuptonTheGood: I think that I intended to return 1 here.  As @parejkoj says, `parseAndRun` adds up the error codes, so it returns the number of failed `dataRefs`

PR #49, opened 2018-05-11
DM-14366: Make pipe_base and pipe_tasks pep8 compliant

r-owen:


python/lsst/pipe/base/cmdLineTask.py
```diff
@@ -50,7 +50,7 @@ def _poolFunctionWrapper(function, arg):
         return function(arg)
     except Exception:
         raise  # No worries
-    except:
+    except:  # noqa E722
```
timj: See the discussion in lsst/pipe_supertask#17 where I discussed this exact problem with @andy-slac -- I think this should be `BaseException` or should maybe disappear completely in Python 3 land.

python/lsst/pipe/base/cmdLineTask.py
```diff
@@ -173,7 +173,7 @@ class TaskRunner(object):
     .. __:  http://stackoverflow.com/questions/1408356/keyboard-interrupts-with-pythons-multiprocessing-pool
     """
 
-    TIMEOUT = 3600*24*30
+    TIMEOUT = 3600 * 24 * 30
```
timj: I thought we required multiplications to not have spaces.

kfindeisen:
Why isn't E226 on the ignore list? I thought that one was [mandatory](https://developer.lsst.io/python/style.html#exceptions-to-pep-8).

r-owen:
It would probably be better to include E226, but since this is already merged, I suggest a new ticket.

PR #50, opened 2018-05-11
DM-14359: Fix data ID handling in ap_*

kfindeisen:
This PR factors out data ID expansion and verification from `DataRefContainer`, moving it into `daf_persistence` so that it can be used by other code.

This PR must be merged after lsst/daf_persistence#93.


parejkoj:
@r-owen how long until you can merge your flake8 travis fixup that's causing these complaints? I think we need to be very careful to only turn on travis if the travis yaml and flake8 cleanups are going to be merged in less than a day, otherwise other people trip over the problem and end up redoing the work themselves.

timj:
I'm reviewing #49 today.

kfindeisen:
Ok, somehow I missed that one. I'll remove my second commit, then.

PR #51, opened 2018-05-21
DM-14003: Create minimal ShimButler for v2 on top of v3

pschella:


timj:
Shutting this PR down since it will never be merged and the associated Jira ticket was marked Won't Fix.

PR #52, opened 2018-06-12
DM-14770: tests mis-call PropertySet.get

r-owen:


RobertLuptonTheGood:
I know that KT has pushed to *not* make PropertySet look like a dict, but I am ever-less convinced that he's right on this (I'd be interested in @jbosch's opinion).  The gratuitous differences are irritating and a source of difficulty.  I am aware that there's an `toDict` method, but the mere fact that this exists to provide a different interface with no extra functionality suggests that we should just make a PropertySet look 95% like a dict.

TallJimbo:
My (not deeply considered) $0.02 on this is that the associative container we use to directly represent (and especially round-trip) FITS headers shouldn't try to be too `dict`-like, but we should probably have another class that is both usable in C++ and a true Mapping (i.e. be completely `dict`-like) in Python.

But I'd like to review the places we want to use these before really committing to that opinion.

r-owen:
I'm uneasy about making `PropertySet` and `PropertyList` too dict-like due to serious internal differences that make it a poor mental model. That said `__contains__` seems reasonable. I think DM-13222 will help a bit by getting rid of the `asArray` argument to `get` (which can too easily be mistaken for a `default` argument, as per this ticket).

I wish we could use a different container for FITS header cards. `PropertyList` has some unfortunate limitations there (e.g. no support for missing values, no ability to COMMENT cards and such in the original order).

PR #55, opened 2018-07-18
Add simple unit test of struct construction exception

isullivan:


tests/test_struct.py
```diff
@@ -49,6 +49,12 @@ def testInit(self):
         for name, val in self.valDict.items():
             self.assertEqual(getattr(s, name), val)
 
+    def testInitException(self):
+        """Test Struct.__init__
```
r-owen: I suggest a more informative description, something along the lines of "test that names cannot start with __" or "test Struct constructor errors"

isullivan: Oops.. yes I forgot to update that description!

PR #57, opened 2018-07-20
DM-15162: Improve documentation for DataIdContainer

r-owen:
and also clean up some other docs.

### DIFF
--- a/python/lsst/pipe/base/argumentParser.py
+++ b/python/lsst/pipe/base/argumentParser.py
@@ -58,7 +58,8 @@ def _fixPath(defName, path):
     Parameters
     ----------
     defName : `str`
-        Name of environment variable containing default root path; if the environment variable does not exist
+        Name of environment variable containing default root path;
+        if the environment variable does not exist
         then the path is relative to the current working directory
     path : `str`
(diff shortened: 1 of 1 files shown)


python/lsst/pipe/base/argumentParser.py
```diff
@@ -81,19 +83,40 @@ class DataIdContainer(object):
...
+        This class does not support `None`, but if it did, `None`
+        would mean the level should not be restricted.
 
     Notes
     -----
-    Override for data IDs that require special handling to be converted to ``data references``, and specify
-    the override class as ``ContainerClass`` for ``add_id_argument``. (If you don't want the argument parser
-    to compute data references, you may use this class and specify ``doMakeDataRefList=False`` in
-    ``add_id_argument``.)
+    **Fields:**
```
parejkoj: Should these be documented in a `Parameters` block, or should they be declared as class members and documented there?

https://developer.lsst.io/python/numpydoc.html#examples-of-constant-and-class-attribute-docstrings

r-owen:
I moved the field documentation and rebased the commit as per your suggestions

PR #58, opened 2018-07-26
DM-2639: {{Standardize primary method names, run/runDataRef, across PipeTasks}}

czwa:
Rename methods according to RFC-352.
Fifteen total repositories with changes.

### DIFF
--- a/python/lsst/pipe/base/cmdLineTask.py
+++ b/python/lsst/pipe/base/cmdLineTask.py
@@ -37,7 +37,7 @@ from .argumentParser import ArgumentParser
 from lsst.base import Packages
 from lsst.log import Log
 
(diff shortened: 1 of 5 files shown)


python/lsst/pipe/base/cmdLineTask.py
```diff
@@ -429,6 +429,17 @@ def __call__(self, args):
                 exitStatus=exitStatus,
             )
 
+    def _runDataRef(self, task, dataRef, kwargs):
```
PaulPrice: This method needs a docstring.
I suggest naming it `runTask`.
I'm not a fan of the leading underscore on the name, which tells the user that this very useful method should never be overridden, which I think is a bit much to ask.

czwa: Accepted and changed throughout.  This makes the LegacyTaskRunner clearer, as it never called runDataRef().

PR #59, opened 2018-07-30
DM-15220: Move SuperTask class here from pipe_supertask

andy-slac:
SuperTask has new name now, from now on it is to be called
PipelineTask, grepping for "super" should only find calls to super()
method. Configuration classes for PipelineTaks were moved here too (in
config module).

Note that this cannot be merged until we have `daf_butler` included in `lsst_distrib` (which should happen ~soon).


timj:
@andy-slac I think you can merge this any time you get it reviewed. If pipe_base now depends on daf_butler then daf_butler will be pulled in automatically. We don't want to put daf_butler into lsst_distrib on its own if pipe_base already pulls it in.

timj:
Although I don't see any diff for the table file to pull in daf_butler.

andy-slac:
My fault, thanks for starting the review :smiley: I'll fix table file.

timj:
I would have preferred the history was retained during the transfer. I think `git format-patch` and `git am` could have worked.

andy-slac:
Let me see if I can bring that back. pipe_supertask history was itself very messy, not sure if anyone would want to see that.

andy-slac:
@timj, do you like https://github.com/lsst/pipe_base/compare/tickets/DM-15220-1 better? If yes then I'll switch that branch for this one.

timj:
The new branch looks okay to me from a quick glance.

andy-slac:
OK, now all history is here too.

PR #60, opened 2018-08-02
DM-14823: Simplify PipelineTask API for common use case

andy-slac:
The run() metod interface was changed to only accept input data as
keyword arguments and does not have arguments for output data. This
should simplify task implementation for most common use case with a
single output DataId (per dataset type). More complex cases will be
handled by re-implementing new method runWithOutputs() which has
arguments for both input data and output DataIds.

Piece of `runQuantum()` code was refactored into a separate method
`_saveStruct()` which can also be re-implemented in a subclass.

### DIFF
--- a/python/lsst/pipe/base/pipelineTask.py
+++ b/python/lsst/pipe/base/pipelineTask.py
@@ -72,9 +72,9 @@ class PipelineTask(Task):
     initInputs : `dict`, optional
         A dictionary of objects needed to construct this PipelineTask, with
         keys matching the keys of the dictionary returned by
-        :py:meth:`getInitInputDatasetTypes` and values equivalent to what
-        would be obtained by calling `Butler.get` with those DatasetTypes and
-        no data IDs.  While it is optional for the base class, subclasses are
+        `getInitInputDatasetTypes` and values equivalent to what would be
+        obtained by calling `Butler.get` with those DatasetTypes and no data
+        IDs.  While it is optional for the base class, subclasses are
         permitted to require this argument.
     """
 
@@ -98,10 +98,10 @@ class PipelineTask(Task):
         -------
         datasets : `dict`
             Dictionary with keys that match those of the dict returned by
-            :py:meth:`getInitOutputDatasetTypes` values that can be written
-            by calling `Butler.put` with those DatasetTypes and no data IDs.
-            An empty `dict` should be returned by tasks that produce no
-            initialization outputs.
+            `getInitOutputDatasetTypes` values that can be written by calling
+            `Butler.put` with those DatasetTypes and no data IDs. An empty
+            `dict` should be returned by tasks that produce no initialization
+            outputs.
         """
         return {}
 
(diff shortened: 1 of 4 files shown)


python/lsst/pipe/base/pipelineTask.py
```diff
@@ -244,22 +244,77 @@ def getDatasetTypes(cls, config, configClass):
...
-        butler or databases. All interaction with external data happens in
-        `runQuantum` method.
-
-        With default implementation of `runQuantum()` this method will
-        receive keyword arguments whose names will be the same as names
-        of configuration fields describing input and output dataset types.
-        For input dataset types argument values will be lists of the data
-        object retrieved from data butler. For output dataset types argument
-        values will be the lists of units from DataRefs in a Quantum.
+        This method is called by :py:meth:`runQuantum` to operate on input
```
TallJimbo: You should just be able to say `runQuantum` without the `:py:meth:` business; I'm not sure where the latter is required (@jonathansick?) but I'm pretty sure links in docstrings will work fine without it as long as there are no namespace-qualification issues (and there shouldn't be for a method of the same class).

(Several other instances of this throughout the changeset that should also be fixed.)

python/lsst/pipe/base/pipelineTask.py
```diff
@@ -309,24 +364,47 @@ def runQuantum(self, quantum, butler):
...
+                outputDataIds[key] = [dataRef.dataId for dataRef in dataRefs]
 
         # call run method with keyword arguments
-        struct = self.run(**inputs, **outUnits)
+        struct = self.runWithOutputs(inputs, outputDataIds)
+
+        # store produced ouput data
+        self._saveStruct(struct, outputDataRefs, butler)
+
+    def _saveStruct(self, struct, outputDataRefs, butler):
```
TallJimbo: I don't see a good reason to treat `_saveStruct` and `runWithOutputs` differently w.r.t. leading underscores - in C++, I'd have called them both (non-pure) virtual protected, and while I don't have a strong opinion about whether that merits an underscore in Python, I think they should be treated consistently.

python/lsst/pipe/base/pipelineTask.py
natelust: I think that is correct, only the backticks should be needed

python/lsst/pipe/base/pipelineTask.py
```diff
@@ -309,24 +364,47 @@ def runQuantum(self, quantum, butler):
                 inputs[key] = [butler.get(dataRef.datasetType.name, dataRef.dataId)
                                for dataRef in dataRefs]
```
natelust: There is no good way to put a comment in a non diff line, so I am putting this here. The changes made require that the documentation in the docstring for run quantum need to be updated. The described behavior is not what happens any more.

Also in the doc string "must be lists of data bjects corresponding"
bjects -> objects

python/lsst/pipe/base/pipelineTask.py
```diff
@@ -244,22 +244,77 @@ def getDatasetTypes(cls, config, configClass):
                 dsTypes[key] = cls.makeDatasetType(value)
         return dsTypes
 
-    def run(self, *args, **kwargs):
+    def runWithOutputs(self, inputData, outputDataIds):
```
natelust: I have very strong reservations about a method called runWithOutputs that in the default implementation does nothing with the outputDataIds variable, or anything output related

TallJimbo:
@natelust and I discussed this a bit in person, and we came up with the following recommendation:
 - Let's rename `runWithOutputs` to `adaptArgsAndRun` (open to counter proposals, but we think that better captures what it's doing).
 - Let's add an `inputDataIds` argument to `adaptArgsAndRun` containing a dictionary of *input* data IDs, with the same structure as `outputDataIds`.  In all of the real cases I could think of where the concrete `PipelineTask` wants output IDs, they'd want the input IDs, too.  We'd guarantee that the lists of IDs in `inputDataIds` and the lists of unpersisted objects in `inputData` would be `zip`-iterable in case an override of `adaptArgsAndRun` needed.  We think it'd be better to make those separate-but-related data structures so the default implementation of `adaptArgsAndRun` simple (as it is now).

I still think it'd be good to do that "make guaranteed single-element lists into single objects" transformation I brought up elsewhere on this ticket into `runQuantum` rather than `adaptArgsAndRun` (so the dicts passed to `adaptArgsAndRun` would sometimes have single objects as dict values rather than lists).

python/lsst/pipe/base/pipelineTask.py
```diff
@@ -244,22 +244,77 @@ def getDatasetTypes(cls, config, configClass):
...
+        returned struct are lists, if tasks produces more than one object
+        for some dataset type then data objects returned in ``struct`` must
+        match in count and order corresponding DataIds in ``outputDataIds``.
+
+        Parameters
+        ----------
+        inputData : `dict`
+            Dictionary whose keys are the names of the configuration fields
+            describing input dataset types and values are lists of
+            Python-domain data objects retrieved from data butler.
```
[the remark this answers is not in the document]
andy-slac: So, I guess the proposal is to add an extra attribute `scalar` to In/OutputDatasetConfig and make it `False` by default?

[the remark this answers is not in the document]
TallJimbo: Yup, sounds good.  @natelust, will that require any extra work to be included in the syntactic sugar for those config objects?

python/lsst/pipe/base/pipelineTask.py
andy-slac: In my (twisted) mental model `_saveStruct()` is indeed protected (or even private) and not to be called by clients directly, while `runWithOutputs()` is public because it is the same as `run()`, only better :) Anyways, I'll make thing uniform-no-underscores.

python/lsst/pipe/base/pipelineTask.py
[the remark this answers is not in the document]
natelust: I think it would just be one small change to add that to the wrapper function signature, everything else would propagate though

andy-slac:
@TallJimbo, @natelust, I think I have addressed all issues, including adding `scalar` field. Could you check quickly that it looks OK now?

python/lsst/pipe/base/pipelineTask.py
```diff
@@ -299,34 +382,80 @@ def runQuantum(self, quantum, butler):
...
             if isinstance(value, InputDatasetConfig):
                 dataRefs = quantum.predictedInputs[value.name]
-                inputs[key] = [butler.get(dataRef.datasetType.name, dataRef.dataId)
-                               for dataRef in dataRefs]
-
-        # lists of units for output datasets
-        outUnits = {}
+                dataIds = [dataRef.dataId for dataRef in dataRefs]
+                data = [butler.get(dataRef.datasetType.name, dataRef.dataId)
+                        for dataRef in dataRefs]
```
TallJimbo: After DM-14822 this can just be
```
data = [butler.get(dataRef) for dataRef in dataRefs]
```

python/lsst/pipe/base/pipelineTask.py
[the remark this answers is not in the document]
andy-slac: Yes, I did that and also added something to docstring for the wrapper method..

andy-slac:
Rebased to master after DM-15220 was merged

PR #70, opened 2018-12-17
DM-16819: Add shims for Gen2 APIs with Gen3 implementations.

TallJimbo:


python/lsst/pipe/base/shims.py
```diff
@@ -0,0 +1,320 @@
...
+
+class ShimDataRef:
+    """A shim for a Gen2 `~lsst.daf.persistence.ButlerDataRef` with a Gen3
+    `~lsst.daf.butler.Butler` backend.
+
+    Parameters
+    ----------
+    butlerSubset : `ShimButlerSubset`
+        ButlerSubset shim instance.
+        Name of the dataset type.
```
natelust: Is is not clear what the "Name of the dataset type" applies to in the doc string

python/lsst/pipe/base/shims.py
```diff
@@ -0,0 +1,320 @@
...
+    dataId : `dict` or `~lsst.daf.butler.DataId`
+        Generation 3 data ID associated with this reference.
+    """
+
+    GENERATION = 2.5
+    """This is a Generation 2 shim for a Generation3 Butler.
+    """
+
+    def __init__(self, butlerSubset, dataId):
```
natelust: Why are we taking a subset here, it seems like it is not really used in lue of the datasetType. Is it just because of the API of the old DataRef?

python/lsst/pipe/base/shims.py
```diff
@@ -0,0 +1,320 @@
...
+    """
+
+    GENERATION = 2.5
+    """This is a Generation 2 shim for a Generation3 Butler.
+    """
+
+    def __init__(self, butler, datasetType, dataIds):
+        self.butler = butler
+        self.datasetType = datasetType
+        self._dataIds = list(dataIds)
```
natelust: If you are going to have iter defined in the same class as it is iterating over, then I think that _dataIds should be constructed as a tuple instead of a list. This will prevent someone from messing with the _dataIds while an existing iterator is outstanding. Unless of course this is needed for backwards compatibility were people abuse modifying the private variable. The other option is to have __iter__ construct and return an object that has its own list pointing to each of the elements.

python/lsst/pipe/base/shims.py
```diff
@@ -0,0 +1,320 @@
...
+
+        Parameters
+        ----------
+        datasetType : `str`
+            Name of the Gen2 dataset type.
+        dataId : `dict` or `~lsst.daf.butler.DataId`, optional
+            A Generation 3 data ID that identifies the dataset.
+        immediate : `bool`
+            This option is provided for compatibility with
+            `lsst.daf.persistence.Butler`, but must be `True`.
```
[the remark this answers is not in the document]
TallJimbo: I suppose in this case that might be better, just because it's possible there is code with `immediate=False` that doesn't really care whether it gets a proxy back.  But I do think we want to check the values of the other not-actually-supported options in this commit, just to make errors earlier and friendlier.

python/lsst/pipe/base/shims.py
TallJimbo: Oops, dropped a line or something.  Will fix.

python/lsst/pipe/base/shims.py
TallJimbo: It's just the old API - we almost always use `DataRef` as a combination of a dataset type and a data ID, but it originated as the thing iterated over by a `ButlerSubset`.  One relic of that is that it's moderately common to see `dataRef.butlerSubset.butler`, and I wanted to support that.

PR #72, opened 2018-12-20
Tickets/DM-16797

natelust:


python/lsst/pipe/base/config.py
```diff
@@ -209,6 +214,12 @@ class _BaseDatasetTypeConfig(pexConfig.Config):
                            doc="name of the DatasetType")
     storageClass = pexConfig.Field(dtype=str,
                                    doc="name of the StorageClass")
+    nameTemplate = pexConfig.Field(dtype=str,
+                                   default='',
+                                   optional=True,
+                                   doc=("Templated name of string, used to set name "
+                                        "field according to a shared substring at "
+                                        "configuration time"))
```
TallJimbo: Maybe change "configuration time" to "when `formatTemplateNames` is called", to be a bit clearer.

python/lsst/pipe/base/config.py
```diff
@@ -284,6 +295,15 @@ class ResourceConfig(pexConfig.Config):
                                   doc="Minimal number of cores needed by task.")
 
 
+def _templateNameStorageDecorator(func):
+    storedParamsDict = {}
```
TallJimbo: I think this variable ends up effectively being a class attribute, by virtue of being a local variable in a closure defined at class scope.  I think what you want is a new instance attribute, which I *think* you can do by just dropping the decorator and doing something like the following in `formatTemplateNames` directly:
```
if not hasattr(self, "_storedParamsDict"):
    self._storedParamsDict = {}
self._storedParamsDict.update(templateParamsDict)
templateParamsDict.update(self._storedParamsDict)
```

A completely different (and possibly cleaner approach) would be to use [Formatter.parse](https://docs.python.org/3.6/library/string.html#string.Formatter.parse) to skip substitution for fields that don't actually have the parameters given in a particular call.

python/lsst/pipe/base/config.py
```diff
@@ -294,6 +314,12 @@ class PipelineTaskConfig(pexConfig.Config):
     quantum = pexConfig.ConfigField(dtype=QuantumConfig,
                                     doc="configuration for PipelineTask quantum")
 
+    @_templateNameStorageDecorator
+    def formatTemplateNames(self, templateParamDict):
```
TallJimbo: I'd recommend renaming this to `substituteDatasetNames` for consistency with the command-line argument.

PR #73, opened 2018-12-31
DM-16867: Move pieces of pipe_supertask to pipe_base

andy-slac:
Modules that have been moved here:
- `configOverrides`
- `graph`
- `graphBuilder`
- `pipeline`
- `pipeTools`
- `pipelineBuilder`
- new ABC `TaskFactory` was added
- all existing unit tests for the above

More or less complete history was copied too, branch `tickets/DM-16867-supertask-history` contains that history and its HEAD corresponds to the master of `pipe_supertask`. All migration changes to the code were done after merging that branch into `tickets/DM-16867`.


python/lsst/pipe/base/pipelineBuilder.py
```diff
@@ -55,123 +50,83 @@
...
     """
+    def __init__(self, taskFactory, pipeline=None):
+        if pipeline is None:
+            pipeline = Pipeline()
+        self._taskFactory = taskFactory
+        self._pipeline = pipeline
 
-    def __init__(self, taskFactory):
-        self.taskFactory = taskFactory
+    def pipeline(self, order_pipeline=False):
```
TallJimbo: Keyword arguments in this corner of the codebase should be camelCase.  Given that "pipeline" is already the method name, though, I'd just called this argument `ordered`.

python/lsst/pipe/base/pipelineBuilder.py
```diff
@@ -55,123 +50,83 @@
...
-                self._configOverrideFile(pipeline, action.label, action.value)
-
         # conditionally re-order pipeline if requested, but unconditionally
         # check for possible errors
-        orderedPipeline = pipeTools.orderPipeline(pipeline, self.taskFactory)
-        if args.order_pipeline:
-            pipeline = orderedPipeline
-
-        return pipeline
+        orderedPipeline = pipeTools.orderPipeline(self._pipeline, self._taskFactory)
```
TallJimbo: Missing import for `pipeTools`?

python/lsst/pipe/base/pipelineBuilder.py
```diff
@@ -55,123 +50,83 @@
...
+        # need instrument/package name to find overrides
+        obsPkg = None
+        instrument = None
+        if instrument_overrides:
+            # TODO: will need something different for gen3
+            # mapperClass = dafPersist.Butler.getMapperClass(args.input)
+            # instrument = mapperClass.getCameraName()
+            # obsPkg = mapperClass.getPackageName()
+            pass
+
```
TallJimbo: I would recommend just removing all of the placeholder logic for applying obs-package overrides.  It doesn't do anything yet, and I'm not at all confident that we'll want to do it at this stage anyway.

python/lsst/pipe/base/pipelineBuilder.py
```diff
@@ -55,123 +50,83 @@
...
         # load task class, will throw on errors
-        taskClass, taskName = self.taskFactory.loadTaskClass(taskName)
+        taskClass, taskName = self._taskFactory.loadTaskClass(taskName)
```
TallJimbo: Will this ever return a different `taskName` than the one that was passed in?

python/lsst/pipe/base/pipelineBuilder.py
```diff
@@ -195,97 +150,86 @@ def _newTask(self, pipeline, taskName, label=None, obsPkg=None, instrument=None)
...
         """
-        idx = pipeline.labelIndex(label)
+        idx = self._pipeline.labelIndex(label)
         if idx < 0:
             raise LookupError("Task label is not found: " + label)
-        del pipeline[idx]
+        del self._pipeline[idx]
 
-    def _moveTask(self, pipeline, label, new_idx):
+    def moveTask(self, label, new_idx):
```
TallJimbo: Is the index of a `PipelineTask` in a `Pipeline` or `PipelineBuilder` really user-visible?  I'd think it'd be better for these objects to make the ordering deterministic (i.e. topologically sorted given DatasetType input/output relationships, maybe lexicographical to break ties).  I'm not opposed to letting them having an unsorted state with an explicit validation or sort operation, but I don't think it's a good idea for them to expose an index-based interface to the Tasks.

python/lsst/pipe/base/taskFactory.py
```diff
@@ -0,0 +1,97 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Module definig TaskFactory interface.
```
TallJimbo: Typo.  Also not sure this docstring adds any value.

python/lsst/pipe/base/taskFactory.py
```diff
@@ -0,0 +1,97 @@
...
+        """Locate and import PipelineTask class.
+
+        Returns tuple of task class and its full name, `None` is returned
+        for both if loading fails.
+
+        Parameters
+        ----------
+        taskName : `str`
+            Name of the PipelineTask class, interpretation depends entirely on
+            activator, e.g. it may or may not include dots.
```
TallJimbo: Would it be appropriate to say that the interpretation now is now entirely up to the `TaskFactory` and the contexts in which it is used?  Or are there other places where assumptions are made about the form of these names?

python/lsst/pipe/base/graphBuilder.py
```diff
@@ -85,7 +85,7 @@ def __init__(self, taskName, refs):
 # ------------------------
 
 
-class GraphBuilder:
+class GraphBuilder(object):
```
TallJimbo: No need to inherit from `object` in Python 3.

tests/test_exprParserLex.py
```diff
@@ -0,0 +1,267 @@
...
+        if lineno is not None:
+            self.assertEqual(token.lineno, lineno)
+        if lexpos is not None:
+            self.assertEqual(token.lexpos, lexpos)
+
+    def setUp(self):
+        pass
+
+    def tearDown(self):
+        pass
```
TallJimbo: Better to just leave this methods out if they do nothing, I think.

TallJimbo:
Also: I looked at each of the commits since the transfer and commented on them individually (and didn't look carefully at the mass-rename ones).  So there may be comments on things you'd already fixed in later commits, and if you want me to look at any of the code you transferred as whole rather than the changes, please let me know.

python/lsst/pipe/base/pipelineBuilder.py
```diff
@@ -19,21 +19,16 @@
 # the GNU General Public License along with this program.  If not,
 # see <http://www.lsstcorp.org/LegalNotices/>.
 #
+"""Module defining PipelineBuilder class and related methods.
```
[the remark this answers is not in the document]
andy-slac: I do agree that docstring like that add very little (or very little negative) value to the module. OTOH modules are supposed to have a summary docstring in them. I could either drop them entirely or try to expand them (but that would probably duplicate class docstring content).

python/lsst/pipe/base/pipelineBuilder.py
```diff
@@ -55,123 +50,83 @@
 
 
 class PipelineBuilder(object):
```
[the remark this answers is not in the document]
andy-slac: This is leftover of 2-to-3 migration which was never completed for `pipe_supertask`. It is removed in one of the latter commits.

python/lsst/pipe/base/pipelineBuilder.py
andy-slac: I think it should be there, unit test would fail without that import

python/lsst/pipe/base/pipelineBuilder.py
andy-slac: The index is not visible (though we can make it visible when doing `pipetask --show=pipeline`), but tasks are ordered in a pipeline as Pipeline is just a `list`. For execution we'll sort things topologically of course, but my idea was that people may prefer some tasks to be prioritized compared to others (within topological constraints) and one way to do it is re-ordering pipeline. If re-ordering is not useful I can certainly drop it, and there is always an option for an activator to implement its own prioritization later.

python/lsst/pipe/base/taskFactory.py
andy-slac: At this point there are no other assumption, task names are just passed from Pipeline to task factory. TaskFactory implemented in `ctrl_mpexec` expects either something that can be resolved w.r.t. `PYTHONPATH` (e.g. `module.TaskClass`) or just a task class name (super-inefficient to search). Other factories may use different conventions, e.g. registry can be built which uses abstract task names instead of class names, though that would probably make pipeline definitions non-portable.

python/lsst/pipe/base/pipelineBuilder.py
andy-slac: It will return full class name including all package/module names, on input it can be ~anything.

python/lsst/pipe/base/taskFactory.py
andy-slac: And I may be confusing myself here, I think that logic may be a bit different - TaskFactory is used to resolve any user-provided name to a full class name and that is passed then to a Pipeline. Though even that is an implementation detail now and is not documented anywhere.

andy-slac:
Regarding `expr_parser` stuff - it is used though I'm not very happy with its current "interface" (or rather lack of it). `pipetask` accepts user expression string which it parses using `expr_parser` and then makes a string again out of that parsed expression and passes it to a pre-flight. I think better approach would be to have some abstract interface for user expression defined in `daf_butler`, trouble is that I have no idea how that interface can look like. One possible option is to define an interface as a parsed expression tree (AST) as it is defined now in `expr_paser/exprTree.py` but that implies that we "freeze" the grammar for the expression. I need to think a little bit about that, maybe having `expr_paser/exprTree.py` in `daf_butler` is a good first step and the rest can be kept in `pipe_base` as a reusable parser for one sort of syntax.

TallJimbo:
I think it'd make slightly more sense to just put it all in daf_butler for now - while I agree that we don't want to freeze the grammar now, I think any kind of engineering to make extending the parser easier (like trying to divide interface and implementation across packages) is premature.  I am also skeptical that there is any parsing functionality we'd want in pipe_base but not in daf_butler.

andy-slac:
OK, I'm going to move `expr_parser` to `daf_butler` then and undo some commits here, hope I will not make too much mess with that.

python/lsst/pipe/base/pipelineBuilder.py
TallJimbo: Yes, I believe so.

python/lsst/pipe/base/pipelineBuilder.py
TallJimbo: I suspect it was fixed on a later commit and I didn't notice.

python/lsst/pipe/base/taskFactory.py
TallJimbo: I don't know that we need to enforce this, but I can't actually think of a use case for having different `TaskFactory` implementations that might expect different kinds of names, and I think it would be desirable for all activators to handle task names consistently.

I gather the reason you didn't want to move the TaskFactory implementation here was because you didn't quite consider it mature (and DM-17044 backs that up).  Perhaps after it does mature we should just move the implementation here and remove the ABC?

andy-slac: Well, I think I actually prefer TaskFactory to be an ABC here, at least it tells clients what minimal interface is needed for the rest of the thing to work. Implementation in ctrl_mpexec does a little bit more that is needed from this interface - it also supports functionality that is needed by `pipetask` script like making a list of available tasks.

PR #75, opened 2019-01-07
tickets/DM-17060

natelust:
Add Switch for setting Datatype name templates

### DIFF
--- a/python/lsst/pipe/base/configOverrides.py
+++ b/python/lsst/pipe/base/configOverrides.py
@@ -24,6 +24,7 @@
 
 __all__ = ["ConfigOverrides"]
 
+from lsst.pipe.base import PipelineTaskConfig
 import lsst.pex.config as pexConfig
 import lsst.pex.exceptions as pexExceptions
 
@@ -82,6 +83,21 @@ class ConfigOverrides:
         """
         self._overrides += [('value', (field, value))]
 
+    def addDatasetNameSubstitution(self, nameDict):
+        """Add keys and values to be used in formatting config nameTemplates
+
+        This method takes in a dictionary in the format of key:value which
+        will be used to format fields in all of the nameTemplates found in
+        a config object. I.E. a nameDictString = {'input': 'deep'} would be
+        used to format a nameTemplate of "{input}CoaddDatasetProduct".
+
+        Parameters
+        ----------
+        nameDict : `dict`
+            a python dict used in formatting nameTemplates
+        """
+        self._overrides += [('namesDict', nameDict)]
+
     def applyTo(self, config):
         """Apply all overrides to a task configuration object.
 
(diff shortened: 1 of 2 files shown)


python/lsst/pipe/base/configOverrides.py
```diff
@@ -123,3 +140,16 @@ def applyTo(self, config):
                     # this can throw
                     value = eval(value, {})
                     setattr(obj, field[-1], value)
+            elif otype == 'namesDict':
+                try:
+                    parsedNamesDict = ast.literal_eval(override)
+                    if not isinstance(parsedNamesDict, dict):
+                        raise ValueError()
```
andy-slac: Some message to user would be useful here

python/lsst/pipe/base/configOverrides.py
```diff
@@ -82,6 +84,21 @@ def addValueOverride(self, field, value):
...
+        a string in the format of key:value which will be used to format
+        fields in all of the nameTemplates found in a config object. I.E.
+        a nameDictString = {'input': deep} would be used to format a
+        nameTemplate of "{input}CoaddDatasetProduct".
+        Parameters
+        ----------
+        nameDictString : str
+            String formatted as a python dict used in formatting nameTemplates
+        """
+        self._overrides += [('namesDict', (nameDictString))]
```
andy-slac: Parens around `nameDictString` don't do anything

python/lsst/pipe/base/configOverrides.py
```diff
@@ -82,6 +84,21 @@ def addValueOverride(self, field, value):
...
+
+        This method takes in a dictionary passed in on the command line as
+        a string in the format of key:value which will be used to format
+        fields in all of the nameTemplates found in a config object. I.E.
+        a nameDictString = {'input': deep} would be used to format a
+        nameTemplate of "{input}CoaddDatasetProduct".
+        Parameters
+        ----------
+        nameDictString : str
+            String formatted as a python dict used in formatting nameTemplates
```
andy-slac: I'd strongly prefer more type-safe interface here. Why no pass a `dict` here and construct that dict in `ctrl_mpexec` instead? This code is supposed to be more generic and reusable now and we should avoid depending on command-line-specific things here.

python/lsst/pipe/base/configOverrides.py
timj: Python code isn't meant to raise `pexExceptions.RuntimeError`.

andy-slac: Regarding ValueError - I think it is more complicated. `literal_eval()` can certainly throw exceptions other than ValueError (SyntaxError is probably even more common). Also I'm not a big fan of chained exceptions as they can be confusing for end users (especially as in this case ValueError does not have any useful message) but I think we do not have any explicit policy against chaining. Still I think adding a message to ValueError here would be better.

PR #78, opened 2019-01-15
Add new method for iterating over QuantumGraph.

andy-slac:
New method returns topologically ordered Quanta together with their
pre-requisites. This is used by new refactored code in ctrl_mpexec.

### DIFF
--- a/python/lsst/pipe/base/graph.py
+++ b/python/lsst/pipe/base/graph.py
@@ -29,15 +29,19 @@ representations.
 """
 
 # "exported" names
-__all__ = ["QuantumGraphNodes", "QuantumGraph"]
+__all__ = ["QuantumGraph", "QuantumGraphTaskNodes", "QuantumIterData"]
 
 # -------------------------------
 #  Imports of standard modules --
 # -------------------------------
+from itertools import chain
 
 # -----------------------------
 #  Imports for other modules --
 # -----------------------------
+from .pipeline import Pipeline
+from .pipeTools import orderPipeline
+from lsst.daf.butler import DataId
 
 # ----------------------------------
 #  Local non-exported definitions --
@@ -48,15 +52,53 @@ __all__ = ["QuantumGraphNodes", "QuantumGraph"]
 # ------------------------
 
 
-class QuantumGraphNodes:
-    """QuantumGraphNodes represents a bunch of nodes in an quantum graph.
+class QuantumIterData:
+    """Helper class for iterating over quanta in a graph.
+
+    `QuantumGraph.traverse` method needs to return topologically ordered
+    Quanta together with their dependencies. This class is used as a value
+    for iterator, it contains enumerated Quantum and its prerequisites.
+
+    Parameters
+    ----------
+    quantumId : `int`
+        Index of this Quantum, unique but arbitrary integer.
+    quantum : `~lsst.daf.butler.Quantum`
+        Quantum corresponding to a graph node.
+    taskDef : `TaskDef`
+        Task to be run on this quantum.
+    prerequisites : iterable of `int`
+        Possibly empty sequence of indices of prerequisites for this Quantum.
+        Prerequisites include other nodes in a graph, they do not reflect
+        data already in butler (there are no graph nodes for those).
+    """
+
+    __slots__ = ["quantumId", "quantum", "taskDef", "prerequisites"]
+
+    def __init__(self, quantumId, quantum, taskDef, prerequisites):
+        self.quantumId = quantumId
+        self.quantum = quantum
+        self.taskDef = taskDef
+        self.prerequisites = frozenset(prerequisites)
+
+    def __str__(self):
+        return "QuantumIterData({}, {}, {})".format(self.quantumId,
+                                                    self.taskDef,
+                                                    self.prerequisites)
+
+
+class QuantumGraphTaskNodes:
+    """QuantumGraphTaskNodes represents a bunch of nodes in an quantum graph
+    corresponding to a single task.
 
     The node in quantum graph is represented by the `PipelineTask` and a
-    single `Quantum` instance. One possible representation of the graph is
-    just a list of nodes without edges (edges can be deduced from nodes'
-    quantum inputs and outputs if needed). That representation can be reduced
-    to the list of PipelineTasks and the corresponding list of Quanta.
-    This class defines this reduced representation.
+    single `~lsst.daf.butler.Quantum` instance. One possible representation
+    of the graph is just a list of nodes without edges (edges can be deduced
+    from nodes' quantum inputs and outputs if needed). That representation can
+    be reduced to the list of PipelineTasks (or their corresponding TaskDefs)
+    and the corresponding list of Quanta. This class is used in this reduced
+    representation for a single task, and full `QuantumGraph` is a sequence of
+    tinstances of this class for one or more tasks.
 
     Different frameworks may use different graph representation, this
     representation was based mostly on requirements of command-line
@@ -64,9 +106,9 @@ class QuantumGraphNodes:
 
     Attributes
     ----------
-    taskDef : :py:class:`TaskDef`
+    taskDef : `TaskDef`
         Task defintion for this set of nodes.
-    quanta : `list` of :py:class:`lsst.daf.butler.Quantum`
+    quanta : `list` of `~lsst.daf.butler.Quantum`
         List of quanta corresponding to the task.
     """
     def __init__(self, taskDef, quanta):
@@ -75,7 +117,7 @@ class QuantumGraphNodes:
 
 
 class QuantumGraph(list):
-    """QuantumGraph is a sequence of QuantumGraphNodes objects.
+    """QuantumGraph is a sequence of `QuantumGraphTaskNodes` objects.
 
     Typically the order of the tasks in the list will be the same as the
     order of tasks in a pipeline (obviously depends on the code which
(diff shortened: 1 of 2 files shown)


python/lsst/pipe/base/graph.py
```diff
@@ -48,25 +52,60 @@
...
+    Quanta together with their dependencies. This class is used as a value
+    for iterator, it contains enumerated Quantum and its pre-requisites.
+
+    Parameters
+    ----------
+    quantumId : `int`
+        Index of this Quantum, unique but arbitrary integer.
+    quantum : `~lsst.daf.butler.Quantum`
+        Quantum corresponding to a graph node.
+    taksDef : `TaskDef`
```
TallJimbo: typo: `taksDef`

I'm also a bit more generally concerned that `TaskDef` is being used too broadly; it's conceptually just a `Task` class and ` Config` instance, and I think it's usually a good simplification of the design to flatten those out into separate attributes for other structs like this one and `QuantumGraphNodes`.

But maybe we should go further than that - `Quantum` actually already has its own `task` attribute.  That's not actually used anywhere (except to get/put the `Task` name from/to the database).  It doesn't have a `config` attribute, but perhaps it should (while allowing it to be `None`).  That would remove the need for any kind of task or config information here.

In fact, I think we should consider removing that information from `QuantumGraphNodes` as well, and then maybe flatten `QuantumGraph` into *just* a sequence of `Quantum`s.  That essentially denormalizes the task and config information into the individual `Quantums`, but I think that's a good thing: while a single preflight invocation will generate a `QuantumGraph` for which the current structure makes sense, if you instead obtain a `QuantumGraph` by querying provenance information from the database, the current structure does not make sense in general (if the provenance extends beyond a single run).  We haven't implemented that yet, but I certainly plan to.  I could also imagine us in the future wanting to run preflight multiple times with different configurations (or pipelines) for different data IDs, and merge the `QuantumGraph`s prior to execution.

python/lsst/pipe/base/graph.py
```diff
@@ -75,15 +114,15 @@ def __init__(self, taskDef, quanta):
 
 
 class QuantumGraph(list):
-    """QuantumGraph is a sequence of QuantumGraphNodes objects.
+    """QuantumGraph is a sequence of `QuantumGraphTaskNodes` objects.
```
TallJimbo: I've commented much more extensively above on other possibilities for how to implement `QuantumGraph`, but regardless of what we decide there, I'd prefer to have it not inherit from `list`; inheritance brings a lot of operations that don't obviously make sense for this class, including mutators that make it very easy to violate any invariants the class has (I don't know if we treat the ordering of tasks as an invariant right now, but I think we should at least leave the option of doing that in the future).

python/lsst/pipe/base/graph.py
```diff
@@ -96,18 +135,67 @@ def __init__(self, iterable=None):
...
+        ------
+        nodes : `QuantumGraphTaskNodes`
+        """
+        # Tasks in a graph are probably topologically sorted already but there
+        # is no guarantee for that. Just re-construct Pipeline and order tasks
+        # in a pipeline using existing method.
+        nodesMap = {id(item.taskDef): item for item in self}
+        pipeline = orderPipeline(Pipeline(item.taskDef for item in self))
+        for taskDef in pipeline:
+            yield nodesMap[id(taskDef)]
```
TallJimbo: If we don't have a concrete use case for this method, I would remove it; as implied by my earlier comments on the internal data structure for `QuantumGraph`, I think it implies and exposes an implementation that may not be something we want to keep.

python/lsst/pipe/base/graph.py
```diff
@@ -96,18 +135,67 @@ def __init__(self, iterable=None):
...
+
+    def orderedQuanta(self):
+        """Return topologically ordered Quanta and their pre-requisites.
+
+        This method enumerates all Quanta in a graph and iterates over
+        them in topological order so that pre-requisites for given Quantum
+        are always returned before the Quantum.
```
TallJimbo: I would be happy to just call this `quanta()` or `__iter__` and drop the existing `quanta()` method.

python/lsst/pipe/base/graph.py
```diff
@@ -96,18 +135,67 @@ def __init__(self, iterable=None):
...
+
+        Yields
+        ------
+        quantumData : `QuantumIterData`
+        """
+        index = 0
+        outputs = {}  # maps DataId to its producing quantum index
```
TallJimbo: I don't think this is safe: an output needs to be keyed by DatasetType as well as DataId.  If `DatasetRef` were equality comparable and hashable, you could just use that, but I think it's tricky to implement because it's unclear whether it should compare `dataset_id` or just DatasetType and DataId.  After DM-16227 you could use the `DatasetRef.hash()` I've added there as the key, but maybe it'd be better to just use a tuple of `(ref.dataset_type_name, ref.dataId)` here anyway.

python/lsst/pipe/base/graph.py
```diff
@@ -48,25 +52,60 @@
...
+    preRequisites : `list` of `int`
+        Possibly empty list of indices of pre-requisites for this Quantum.
+        Pre-requisites include other nodes in a graph, they do not reflect
+        data already in butler (there are no graph nodes for those).
+    """
+    def __init__(self, quantumId, quantum, taskDef, preRequisites):
+        self.quantumId = quantumId
+        self.quantum = quantum
+        self.taskDef = taskDef
+        self.preRequisites = preRequisites
```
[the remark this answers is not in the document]
andy-slac: I did want to make it `namedtuple` but it's impossible to document that except in comment which is invisible to sphinx. I think we want it documented as it's useful for people implementing activators, so I'll just add `__slots__`.

python/lsst/pipe/base/graph.py
```diff
@@ -48,25 +52,60 @@
...
+    quantumId : `int`
+        Index of this Quantum, unique but arbitrary integer.
+    quantum : `~lsst.daf.butler.Quantum`
+        Quantum corresponding to a graph node.
+    taksDef : `TaskDef`
+        Task to be run on this quantum.
+    preRequisites : `list` of `int`
+        Possibly empty list of indices of pre-requisites for this Quantum.
+        Pre-requisites include other nodes in a graph, they do not reflect
+        data already in butler (there are no graph nodes for those).
```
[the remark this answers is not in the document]
andy-slac: True, indices are transient and are only valid within iteration, to help activator to figure out dependencies between Quanta (which are not hashable).
I'll extend docstring with more explicit info.

python/lsst/pipe/base/graph.py
andy-slac: I agree that TsakDef looks somewhat artificial, and I think this deserves some discussion outside this ticket. I can add here that TaskDef also has `label` attribute and that serves as a sort of connection to original `Pipeline` from which graph was built.

python/lsst/pipe/base/graph.py
andy-slac: Original implementation indeed relied on ordering of this list and it made sense with the old CmdLineFwk. With the addition of the new methods it makes less sense now. 
One more point for having per-task list of Quanta is that it allows doing topological sort at task level rather that at individual Quantum level.

python/lsst/pipe/base/graph.py
andy-slac: I't only used by `orderedQuanta()` now, I'll move it into that method.

python/lsst/pipe/base/graph.py
TallJimbo: If we treat topological ordering as a class invariant established (or checked) at construction, we could keep that advantage for the single-preflight-invocation construction pattern.

python/lsst/pipe/base/graph.py
TallJimbo: Good point about `label`.  Makes me wonder if we should add that to `Quantum`, too.  But as you said, we can discuss this on another ticket.

python/lsst/pipe/base/graph.py
andy-slac: Existing `quanta()` is also used by ctrl_mpexec, that use case does not need topologically ordered Quanta or their dependencies. I think we should support two methods, or alternatively we can require that graph always stores topologically sorted list of quanta then two methods are not needed. I'm not very happy with `__iter__`, as it implies that object is a container of the things returned by iterator, which is not entirely true in this case. We probably need more thought about what perfect interface for a graph is, and I think it should be driven by how it is used. I think that two current methods - `quanta()` and `orderedQuanta()` - are what actually makes class API, everything else at this point can probably be considered implementation detail.

python/lsst/pipe/base/graph.py
andy-slac: 😶 Somehow I keep forgetting about DatasetType, thanks for spotting this. Will fix with a tuple as you suggest.

python/lsst/pipe/base/graph.py
TallJimbo: As you might have guessed my preference would probably be make the graph always topologically sorted, but I don't want to insist on that unless we can convince ourselves that it doesn't make some other operation (e.g. incremental building or merging) much harder.

If we keep both for now, perhaps we could rename `orderedQuanta()` to something that doesn't also imply (as I agree `__iter__` does) that the iterated objects are just `Quantums`?  Maybe `traverse()`?

andy-slac: OK with `traverse()`.

TallJimbo:
One more thought: I know you're not planning to do major refactoring of `QuantumGraph` on this ticket, but we should have a discussion with @MichelleGower about her possible use cases for the class before doing them on any other ticket.

PR #79, opened 2019-01-16
tickets/DM-14819

natelust:
Allow DatasetType to be marked for later load

Although DatasetTypes need to be known prior to task execution to
build a QuantumGraph, not all data products need to be loaded before
task execution. This change introduces a way to indicate a
DatasetType will be manually read in and handled by a task.

### DIFF
--- a/python/lsst/pipe/base/config.py
+++ b/python/lsst/pipe/base/config.py
@@ -129,7 +129,8 @@ def _makeDatasetField(name, dtype):
         extraFields = ""
     elif issubclass(dtype, _DatasetTypeConfig):
         # Handle dataset types like InputDatasetConfig, note these take a dimensions argument
-        def wrappedFunc(*, doc, dimensions, storageClass, name="", scalar=False, check=None, nameTemplate=''):
+        def wrappedFunc(*, doc, dimensions, storageClass, name="", scalar=False, check=None, nameTemplate='',
+                        manualLoad=False):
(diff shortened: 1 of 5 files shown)


python/lsst/pipe/base/config.py
```diff
@@ -245,6 +250,13 @@ class should not be used directly, instead one of `InputDatasetConfig` or
                                   "objects/DataIds will be unpacked before calling task "
                                   "methods, returned data is expected to contain single "
                                   "objects as well."))
+    manualLoad = pexConfig.Field(dtype=bool,
+                                 default=False,
+                                 optional=True,
+                                 doc=("If this is set to True, the class intendes to load "
```
TallJimbo: Typo: `intendes`.

python/lsst/pipe/base/graphBuilder.py
```diff
@@ -337,6 +337,7 @@ def _dataRefKey(dataRef):
                     for key in dataRefs.keys():
                         if dataRefs[key].id is not None:
                             dataRefs[key] = self.registry.getDataset(dataRefs[key].id)
+                            self.registry.expandDataId(dataRefs[key].dataId, region=True)
```
TallJimbo: Should make this one-liner a separate commit, with a message like:
```
Ensure metadata and regions are expanded in QuantumGraph data IDs.

This is a workaround for DM-17231.
```

PR #80, opened 2019-01-22
DM-17045: Convert [SafeClip/CompareWarp]AssembleCoaddTask to PipeTask

yalsayyad:
Fix _sub handling in Gen2->Gen3 shim Butler code


python/lsst/pipe/base/shims.py
```diff
@@ -133,7 +133,7 @@ def get(self, datasetType, dataId=None, immediate=True, **rest):
         fullDataId = self._makeDataId(dataId, **rest)
         if datasetType.endswith("_sub"):
             import lsst.afw.image
```
isullivan: It's not part of these changes, but shouldn't this import be at the top, or at least have a `# noqa` statement?

PR #88, opened 2019-02-06
DM-17741: Improve logging forwarding setup

andy-slac:
Previous assumption that root logger has no handlers does not hold in
case of pytest, resulting in a lot of messages generated by unit test.
Fix is to set default logging level for root logger to INFO (like for
lsst.log) and then adjust levels for individual loggers in sync for both
lsst.log and logging.

### DIFF
--- a/python/lsst/pipe/base/argumentParser.py
+++ b/python/lsst/pipe/base/argumentParser.py
@@ -504,7 +504,7 @@ log4j.appender.A1.layout.ConversionPattern=%c %p: %m%n
 
         # Forward all Python logging to lsst.log
         lgr = logging.getLogger()
-        lgr.setLevel(logging.DEBUG)
+        lgr.setLevel(logging.INFO)  # same as in log4cxx config above
         lgr.addHandler(lsstLog.LogHandler())
 
     def add_id_argument(self, name, datasetType, help, level=None, doMakeDataRefList=True,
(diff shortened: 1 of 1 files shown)


python/lsst/pipe/base/argumentParser.py
```diff
@@ -1279,6 +1279,10 @@ def __call__(self, parser, namespace, values, option_string):
                 namespace.log.setLevel(logLevel)
             else:
                 lsstLog.Log.getLogger(component).setLevel(logLevel)
+            # Python logging levels are same as lsst.log divided by 1000,
+            # logging does not have TRACE level by default but it is OK to use
+            # that numeric level and we may even add TRACE later.
+            logging.getLogger(component).setLevel(logLevel//1000)
```
ktlim: We have `LogHandler.translateLevel()` to multiply by 1000; should we add a routine to divide by 1000 as well, since it appears here as well as in https://github.com/lsst/log/blob/master/python/lsst/log/log/logContinued.py#L123-L124 ?

andy-slac: Yes, I though about that while reviewing Tim's ticket 🙂 I think I'll do it now, need to think where these two methods should be (either a module level method or in `Log` class or `LogHandler`)

PR #95, opened 2019-06-27
DM-19988: rewrite of QuantumGraph generation

TallJimbo:


python/lsst/pipe/base/graphBuilder.py
```diff
@@ -28,36 +29,674 @@
...
+            Registry for the data repository; used for all data ID queries.
+        originInfo : `lsst.daf.butler.DatasetOriginInfo`
+            Object holding the input and output collections for each
+            `DatasetType`.
+        userQuery : `str`, optional
+            User-provided expression to limit the data IDs processed.
+        """
+        # Initialization datasets always have empty data IDs.
+        emptyDataId = DataId(dimensions=registry.dimensions.empty)
+        for datasetType, scaffolding in itertools.chain(self.initInputs.items(),
```
andy-slac: `datasetType` is not used by the loop, you can probably iterate over `values()`.

python/lsst/pipe/base/graphBuilder.py
```diff
@@ -28,36 +29,674 @@
...
+        `prerequisites`).
+
+        Parameters
+        ----------
+        registry : `lsst.daf.butler.Registry`
+            Registry for the data repository; used for all data ID queries.
+        originInfo : `lsst.daf.butler.DatasetOriginInfo`
+            Object holding the input and output collections for each
+            `DatasetType`.
+        skipExisting : `bool`
```
andy-slac: `optional`?

python/lsst/pipe/base/graphBuilder.py
```diff
@@ -93,16 +727,24 @@ class GraphBuilder(object):
...
+    def __init__(self, taskFactory, registry, skipExisting=True, clobberExisting=False):
         self.taskFactory = taskFactory
         self.registry = registry
         self.dimensions = registry.dimensions
         self.skipExisting = skipExisting
+        self.clobberExisting = clobberExisting
+
+    ALLOWED_QUANTUM_DIMENSIONS = frozenset(["instrument", "visit", "exposure", "detector",
+                                            "physical_filter", "abstract_filter",
+                                            "skymap", "tract", "patch"])
```
andy-slac: Should something like this come from configuration instead? Is this used by anything at all? I looked at this package and `ctrl_mpexec` and could not find any references.

TallJimbo: Good catch; looks like a relic of an earlier draft.  I'll just remove it.

PR #97, opened 2019-08-07
DM-20376: Correct exit status

timj:


kfindeisen:
Please make sure this PR also updates the [Sphinx documentation on status codes](https://pipelines.lsst.io/modules/lsst.pipe.base/command-line-task-argument-reference.html#status-code) (in particular, the cap of 127).

ktlim:
Do you think I should increase that to 255?  Traditionally 128-255 has been reserved for signal-caused exits, but I can imagine we would like to have, for example, 189 (CCDs in focal plane) available.
255 is a hard limit, of course.  I guess there should also be at least a log message if we exceed whatever cap there is.

andy-slac:
Exit codes above 127 are still interpreted as "exited due to signal". It may confuse users if parent process says "child was killed by signal 61 (core dumped)".

timj:
@ktlim are we abandoning this branch?  It's gen2-specific. @andy-slac is there an equivalent case in pipetask?

andy-slac:
@timj, I _believe_ pipetask should exit with non-zero code on any error. Proving that may be as hard as solving halting problem 🙂

timj:
Are we going to try to merge this or are we going to close it and the associated Jira ticket?

timj:
The CmdLineTask code no longer exists.

PR #98, opened 2019-08-15
DM-20205: Update to Final PipelineTasks API

natelust:


python/lsst/pipe/base/butlerQuantumContext.py
```diff
@@ -0,0 +1,204 @@
...
+
+        self._get = types.MethodType(_get, self)
+        self._put = types.MethodType(_put, self)
+
+    def get(self, dataset: typing.Union[InputQuantizedConnection, typing.List[DatasetRef], DatasetRef]):
+        """Fetches data from the butler
+
+        Parameters
+        ----------
+        dataset : `InputQuantizedConnection` or `list` of `lsst.daf.butler.DatasetRef`
```
andy-slac: I think our latest convention is to write it as 

    `list` [`lsst.daf.butler.DatasetRef`]

python/lsst/pipe/base/butlerQuantumContext.py
```diff
@@ -0,0 +1,204 @@
...
+                or `lsst.daf.butler.DatasetRef`
+            This argument may either be an `InputQuantizedConnection` which describes
+            all the inputs of a quantum, a list of `lsst.daf.butler.DatasetRef`, or
+            a single `lsst.daf.butler.DatasetRef`. The function will get and return
+            the corresponding datasets from the butler.
+
+        Raises
+        ------
+        ValueError
+            If a `DatasetRef` is passed to get that is not defined in the quantum object
```
andy-slac: Needs module name for `DatasetRef`. I'd add a `~` for each of those too to reduce verbosity in generated docs.

python/lsst/pipe/base/butlerQuantumContext.py
```diff
@@ -0,0 +1,204 @@
...
+        """
+        if isinstance(dataset, InputQuantizedConnection):
+            retVal = {}
+            for name, ref in dataset:
+                if isinstance(ref, list):
+                    val = []
+                    for r in ref:
+                        val.append(self._get(r))
```
andy-slac: `val = [self._get(r) for r in ref]` is just as readable

python/lsst/pipe/base/butlerQuantumContext.py
```diff
@@ -0,0 +1,204 @@
...
+
+        Parameters
+        ----------
+        dataset : `InputQuantizedConnection` or `list` of `lsst.daf.butler.DatasetRef`
+                or `lsst.daf.butler.DatasetRef`
+            This argument may either be an `InputQuantizedConnection` which describes
+            all the inputs of a quantum, a list of `lsst.daf.butler.DatasetRef`, or
+            a single `lsst.daf.butler.DatasetRef`. The function will get and return
+            the corresponding datasets from the butler.
+
```
andy-slac: `Returns` section would be useful too.

python/lsst/pipe/base/butlerQuantumContext.py
```diff
@@ -0,0 +1,204 @@
...
+                else:
+                    val = self._get(ref)
+                retVal[name] = val
+            return retVal
+        elif isinstance(dataset, list):
+            return [self._get(x) for x in dataset]
+        elif isinstance(dataset, DatasetRef) or isinstance(dataset, DeferredDatasetRef):
+            return self._get(dataset)
+        else:
+            raise TypeError("Dataset argument is not a type that can be used to get")
```
andy-slac: I'm not a big fan of the `isinstance`-based switch, it feels a bit fragile for my taste. But I'm not sure anything can be done to avoid it?

python/lsst/pipe/base/butlerQuantumContext.py
```diff
@@ -0,0 +1,204 @@
...
+        Raises
+        ------
+        ValueError
+            If a `DatasetRef` is passed to put that is not defined in the quantum object, or
+            the type of values does not match what is expected from the type of dataset.
+        """
+        if isinstance(dataset, OutputQuantizedConnection):
+            if not isinstance(values, Struct):
+                raise ValueError("dataset is a OutputQuantizedConnection, a Struct with corresponding"
+                                 "attributes must be passed as the values to put")
```
andy-slac: Needs a space before "attributes".

python/lsst/pipe/base/butlerQuantumContext.py
```diff
@@ -0,0 +1,204 @@
...
+            for name, refs in dataset:
+                valuesAttribute = getattr(values, name)
+                if isinstance(refs, list):
+                    if len(refs) != len(valuesAttribute):
+                        raise ValueError(f"There must be a object to put for every Dataset ref in {name}")
+                    for i, ref in enumerate(refs):
+                        self._put(valuesAttribute[i], ref)
+                else:
+                    self._put(valuesAttribute, refs)
```
andy-slac: I think this could be slightly simplified if `_put()` supported lists.

python/lsst/pipe/base/butlerQuantumContext.py
```diff
@@ -0,0 +1,204 @@
...
+        if not isinstance(ref, list):
+            ref = [ref]
+        if inout is RefDirection.IN:
+            for r in ref:
+                if (r.datasetType, r.dataId) not in self.allInputs:
+                    raise ValueError("DatasetRef is not part of the Quantum being processed")
+        if inout is RefDirection.OUT:
+            for r in ref:
+                if (r.datasetType, r.dataId) not in self.allOutputs:
+                    raise ValueError("DatasetRef is not part of the Quantum being processed")
```
andy-slac: There is repetition in this method, I think you can simplify it if you just pass `self.allInputs` or `self.allOutputs` as a parameter instead of enum (and enum will not even have to exist in that case).

python/lsst/pipe/base/pipelineTask.py
```diff
@@ -182,309 +80,32 @@ class PipelineTask(Task):
...
-        names of configuration fields describing input dataset types. Argument
-        values will be data objects retrieved from data butler. If a dataset
-        type is configured with ``scalar`` field set to ``True`` then argument
-        value will be a single object, otherwise it will be a list of objects.
+        This method should be implemented in a subclass. This method will
+        receive keyword arguments whose names will be the same as names of
+        connection fields describing input dataset types. Argument values will
+        be data objects retrieved from data butler. If a dataset type is
+        configured with ``multiple`` field set to ``True`` then the argument
+        value will be a list of objects, otherwise it will be a single objects.
```
andy-slac: single objects - > single object

tests/test_connections.py
```diff
@@ -0,0 +1,163 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Simple unit test for ResourceConfig.
```
andy-slac: ResourceConfig -> connections?

tests/test_connections.py
```diff
@@ -0,0 +1,163 @@
...
+"""
+
+import unittest
+import pytest
+
+import lsst.utils.tests
+import lsst.pipe.base as pipeBase
+
+
+class TestConncetionsClass(unittest.TestCase):
```
andy-slac: Conncetions -> Connections

python/lsst/pipe/base/connections.py
```diff
@@ -0,0 +1,389 @@
...
+
+class PipelineTaskConnections(metaclass=PipelineTaskConnectionsMetaclass):
+    """PipelineTaskConnections is a class used to declare desired IO when a PipelineTask is run by an
+    activator
+
+    PipelineTaskConnection classes are created by declaring class attributes of types defined in
+    lsst.pipe.base.connectionTypes and are listed as follows:
+
+    * InitInput - Defines connections in a quantum graph which are used as inputs to the __init__ function
+                  of the PipelineTask corresponding to this class
```
andy-slac: The lines should not be indented or they will become indented in sphinx output too. I think you want this:

    * InitInput - Defines connections in a quantum graph which are used as inputs to the __init__ function
      of the PipelineTask corresponding to this class

(but also properly wrapped at 79 characters).

You can also try try definition list syntax: http://docutils.sourceforge.net/docs/ref/rst/restructuredtext.html#definition-lists

python/lsst/pipe/base/connections.py
```diff
@@ -0,0 +1,389 @@
...
+
+    The process of declaring a PipelineTaskConnection class involves parameters passed in the declaration
+    statement.
+
+    The first parameter is dimensions which is an iterable of strings which defines the unit
+    of processing the run method of a corresponding `PipelineTask` will operate on. These dimensions must
+    match dimensions that exist in the butler registry which will be used in executing the corresponding
+    `PipelineTask`.
+
+    The second parameter is labeled defaultTemplates and is conditionally optional. The name attributes of
```
andy-slac: I'd put defaultTemplates in double backticks.

python/lsst/pipe/base/connections.py
```diff
@@ -0,0 +1,389 @@
...
+    connections can be specified as python format strings, with named format arguments. If any of the
+    name parameters on connections defined in a `PipelineTaskConnections` class contain a template, then
+    a default template value must be specified in the defaultTemplate argument. This is done by passing
```
andy-slac: defaultTemplate -> defaultTemplates?

python/lsst/pipe/base/connections.py
```diff
@@ -0,0 +1,389 @@
...
+        The `lsst.daf.butler.DatasetRef` that will be eventually used to resolve a dataset
+    """
+    __slots__ = ()
+
+
+class PipelineTaskConnections(metaclass=PipelineTaskConnectionsMetaclass):
+    """PipelineTaskConnections is a class used to declare desired IO when a PipelineTask is run by an
+    activator
+
+    PipelineTaskConnection classes are created by declaring class attributes of types defined in
```
andy-slac: This long description should probably be moved into Notes section after Parameters.

python/lsst/pipe/base/connections.py
```diff
@@ -0,0 +1,389 @@
...
+    name parameters on connections defined in a `PipelineTaskConnections` class contain a template, then
+    a default template value must be specified in the defaultTemplate argument. This is done by passing
+    a dictionary with keys corresponding to a template identifier, and values corresponding to the value
+    to use as a default when formatting the string. For example if ConnectionClass.calexp.name =
+    '{input}Coadd_calexp' then defaultTemplates = {'input': 'deep'}.
+
+    Once a `PipelineTaskConnections` class is created, it is used in the creation of a
+    `PipelineTaskConfig`. This is further documented in the documentation of `PipelineTaskConfig`. For the
+    purposes of this documentation, the relevant information is that the config class allows configuration
+    of connection names by users when running a pipeline.
```
andy-slac: One small example of defining connections class and task config class would be super-helpful here.

python/lsst/pipe/base/butlerQuantumContext.py
natelust: In this case I dont think so if we are going to try and preserve this sort of overload. That is until we get to python 3.8. If (when?) we can take that as the base line, there is a nice built in single dispatch for methods. That will make the code cleaner and more extensible, but fundamentally we want this method overload based on the type of the input.

python/lsst/pipe/base/butlerQuantumContext.py
natelust: I debated back and forth on this, if it was better to have a simple function that did one thing, or do duplicate the handling the lists and single values. I also debated having an intermediate function to handle lists. Ultimately I decided on a simpler _put, but I don't really have strong feelings.

python/lsst/pipe/base/butlerQuantumContext.py
natelust: Really good point, I changed this so many times I could not see the forest for the trees

python/lsst/pipe/base/config.py
```diff
@@ -21,22 +21,18 @@
...
 """Module defining config classes for PipelineTask.
 """
-
-__all__ = ["InputDatasetConfig", "InputDatasetField",
-           "OutputDatasetConfig", "OutputDatasetField",
-           "InitInputDatasetConfig", "InitInputDatasetField",
-           "InitOutputDatasetConfig", "InitOutputDatasetField",
-           "ResourceConfig", "QuantumConfig", "PipelineTaskConfig"]
+__all__ = ["ResourceConfig", "PipelineTaskConfig",
+           "PipelineTaskConnections"]
```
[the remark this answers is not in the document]
natelust: no, was just something left over from before I moved it to its own file, good catch

python/lsst/pipe/base/config.py
```diff
@@ -46,249 +42,81 @@
...
+class PipelineTaskConfig(pexConfig.Config, metaclass=PipelineTaskConfigMeta):
+    """Configuration class for `PipelineTask`s
+
+    This Configuration class functions in largely the same manor as any other derived from
+    `lsst.pex.config.Config`. The only difference is in how it is declared. `PipelineTaskConfig` children need
+    to be declared with a pipelineConnections argument. This argument should specify a child class of
+    `PipelineTaskConnections`. During the declaration of a `PipelineTaskConfig` a config class is created with
+    information from the supplied connections class to allow configuration of the connections class. This
+    dynamically created config class is then attached to the `PipelineTaskConfig` via a
+    `lsst.pex.config.ConfigField` with the attribute name connections.
```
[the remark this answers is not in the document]
natelust: So in this case, config should not be a problem with pickle because it implements its own reduce functionality (and it is not a problem as it is currently used successfully inside a python multiprocessing environment).

python/lsst/pipe/base/connections.py
```diff
@@ -0,0 +1,389 @@
...
+import itertools
+import string
+
+from . import config as configMod
+from .connectionTypes import (InitInput, InitOutput, Input, PrerequisiteInput,
+                              Output, BaseConnection)
+from lsst.daf.butler import DatasetRef, Quantum
+
+if typing.TYPE_CHECKING:
+    from .config import PipelineTaskConfig
```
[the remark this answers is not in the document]
natelust: Actually no, it introduces a circular import problem. config.py imports connections.py at the top of it. In the process of importing connections (in config.py) to create PipelineTaskConfig, connections.py tries to reference the not yet created PipelineTaskConfig. And typing will only resolve a string into a type if it is within the module, not chained off of something it imports. There is a page on type annotations describing this problem, and this was the recommended solution.

python/lsst/pipe/base/connections.py
```diff
@@ -0,0 +1,389 @@
...
+
+    Parameters
+    ----------
+    key : `str`
+        Name of the configuration field for dataset type.
+    numDataIds : `int`
+        Actual number of DataIds in a Quantum for this dataset type.
+    """
+    def __init__(self, key, numDataIds):
+        super().__init__(("Expected scalar for output dataset field {}, "
```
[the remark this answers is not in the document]
natelust: :) This is code I moved from somewhere else into here, and I was pretty sure you were the original author....

natelust:
> It looks OK (if my understanding of how it work is correct), though a bit too much meta-magic for my taste. I think it probably needs good user manual or maybe more code examples in docstrings.

I am working on as a separate body of work (as to not hold this up) documentation that will go in the package documentation talking about how this works, and including a long tutorial on how this all gets used to make PipelineTasks.

PR #103, opened 2019-10-09
DM-21390: Add a show option to examine a qgraph

hsinfang:


python/lsst/pipe/base/graph.py
```diff
@@ -173,6 +173,21 @@ def quanta(self):
             for quantum in taskNodes.quanta:
                 yield taskDef, quantum
 
+    def quanta_as_qgraph(self):
```
andy-slac: For consistency with other methods I think it better to use camelCase for method name (`quantaAsQgraph`).

python/lsst/pipe/base/graph.py
```diff
@@ -173,6 +173,21 @@ def quanta(self):
...
+        """Iterator over quanta in a graph.
+
+        QuantumGraph containing individual quanta are returned.
+
+        Yields
+        ------
+        graph : `QuantumGraph`
+        """
+        for (taskDef, quantum) in self.quanta():
```
andy-slac: Parentheses are not needed here.

PR #111, opened 2019-12-06
DM-21911: Retrofit Gen 3 functionality onto MetricTask

kfindeisen:
This PR fixes a number of Sphinx errors in `pipe_base` that were leaking into `verify` builds.


python/lsst/pipe/base/butlerQuantumContext.py
```diff
@@ -91,8 +91,7 @@ def get(self, dataset: typing.Union[InputQuantizedConnection,
 
         Parameters
         ----------
-        dataset : `InputQuantizedConnection` or `list` [`~lsst.daf.butler.DatasetRef`]
-                or `~lsst.daf.butler.DatasetRef`
+        dataset
```
natelust: I'm not sure dropping the types is the best thing here and below.

python/lsst/pipe/base/connections.py
```diff
@@ -266,46 +266,55 @@ class PipelineTaskConnections(metaclass=PipelineTaskConnectionsMetaclass):
...
-    correspond to an attribute of a `Struct` that is returned by a
-    `PipelineTask` run method.  E.g. if an output connection is defined with
-    the name measCat, then the corresponding PipelineTask.run method must
-    return Struct(measCat=X,..) where X matches the storageClass type defined
-    on the output connection.
-
-    The process of declaring a PipelineTaskConnection class involves parameters
-    passed in the declaration statement.
-
-    The first parameter is dimensions which is an iterable of strings which
```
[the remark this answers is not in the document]
kfindeisen: I think definition lists are more consistent with the overall DM style (e.g., [parameter lists](https://developer.lsst.io/python/numpydoc.html#parameters) or [member lists](https://developer.lsst.io/python/numpydoc.html#py-docstring-parameter-types-struct)), as well as more readable. Maybe it's worth discussing on `#dm-docs` whether there should be any guidelines for a case like this?

python/lsst/pipe/base/butlerQuantumContext.py
kfindeisen: This was recommended by @jonathansick on `#dm-docs`, on the grounds that the original form [was giving a formatting error](https://lsstc.slack.com/archives/C2B6DQBAL/p1575660845102500) and that the types are explicitly listed in the description anyway.

PR #116, opened 2020-02-07
DM-22599: Develop PipelineTask unit test framework

kfindeisen:
This PR is a reopening of #114. It adds a module, `lsst.pipe.base.tests`, with test utilities specific to PipelineTask subclasses. The intent of these tests is to enable unit testing of:
* whether a task's `Connections` are correctly written and whether they match the inputs and outputs of the `run` method
* any logic in a custom `runQuantum` method
* configuration logic, such as optional or alternative inputs or outputs

The test utilities should be usable with minimal knowledge of the Butler, overlapping as much as possible with the knowledge needed to write a `PipelineTask` (i.e., dimensions and storage classes).

This PR makes heavy use of Butler-specific test utilities previously merged on lsst/daf_butler#229. All except for the `MetricsExample` mock data type would be needed for testing real tasks as well.


python/lsst/pipe/base/tests.py
```diff
@@ -0,0 +1,192 @@
...
+    datasetType = connection.makeDatasetType(universe)
+    try:
+        butler.registry.getDatasetType(datasetType.name)
+    except KeyError:
+        raise ValueError(f"Invalid dataset type {connection.name}.")
+    try:
+        return DatasetRef(
+            datasetType=datasetType,
+            dataId=dataId,
+        )
```
mrawls: I think it would be a bit clearer to explicitly define `ref = DatasetRef(stuff)` and then return `ref`.

python/lsst/pipe/base/tests.py
```diff
@@ -0,0 +1,192 @@
...
+    except KeyError as e:
+        raise ValueError(f"Dataset type ({connection.name}) and ID {dataId.byName()} not compatible.") \
+            from e
+
+
+def runQuantum(task, butler, quantum, mockRun=True):
+    """Run a PipelineTask on a Quantum.
```
mrawls: Since the default behavior is to return a Mock, I think this would be clearer if it were called something more specific. Maybe `validateRunQuantum`? The name `runQuantum` is too similar to `runDataRef` or just `run` and incorrectly implies that somehow, here in these validation tests, is where the running of a PipelineTask is to occur.

tests/test_pipelineTaskTests.py
```diff
@@ -0,0 +1,416 @@
...
+import lsst.utils.tests
+import lsst.pex.config
+import lsst.daf.butler
+import lsst.daf.butler.tests as butlerTests
+
+from lsst.pipe.base import Struct, PipelineTask, PipelineTaskConfig, PipelineTaskConnections, connectionTypes
+from lsst.pipe.base.tests import runQuantum, makeQuantum, validateOutputConnections
+
+
+class VisitConnections(PipelineTaskConnections, dimensions={"instrument", "visit"}):
```
mrawls: I believe we're attempting to phase out `visit` in favor of `expId`. Unless you know otherwise, please do so throughout.

tests/test_pipelineTaskTests.py
```diff
@@ -0,0 +1,416 @@
...
+    @classmethod
+    def tearDownClass(cls):
+        shutil.rmtree(cls.root, ignore_errors=True)
+        super().tearDownClass()
+
+    def setUp(self):
+        super().setUp()
+        self.butler = butlerTests.makeTestCollection(self.repo)
+
+    def testMakeQuantumNoSuchDatatype(self):
```
mrawls: My instinct in these tests is to have significantly fewer functions with multiple situations tested in each so you don't have to duplicate the setup code lines as much. For instance, `testMakeQuantum` which would have a comment indicating that where you test the "no such dataset type" case, then later a comment that you're testing the "invalid dimension" case, then the "missing multiple" case, etc. Is there a stylistic or other reason you prefer to do it this way?

python/lsst/pipe/base/tests.py
```diff
@@ -0,0 +1,192 @@
...
+
+
+def _checkDataIdMultiplicity(dataIds, multiple):
+    if multiple:
+        if isinstance(dataIds, collections.abc.Sequence):
+            return dataIds
+        else:
+            raise ValueError(f"Expected multiple data IDs, got {dataIds}.")
+    else:
+        if isinstance(dataIds, collections.abc.Mapping) or isinstance(dataIds, DataCoordinate):
```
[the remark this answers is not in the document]
timj: `DataCoordinate` is indeed a `collections.abc.Mapping` so you only need the first test.

tests/test_pipelineTaskTests.py
timj: I think that is waiting on daf_butler to catch up with the dimensions definitions. visit is allowed when it's really a visit as opposed to a pipeline that's working on individual exposures.

kfindeisen:
> Is there a strong need to call the file tests.py here? "Test" is an overloaded term and I think what's happening is more validation-like.

Well, these are tools for writing unit tests, and we tend to say "validation" or "verification" more in the context of integration testing. `tests` is also a name used elsewhere for test support code, as in `lsst.utils.tests`.

Could we call it `testUtils.py`? That's a convention used in, e.g., `afw`.

tests/test_pipelineTaskTests.py
kfindeisen: I think most of the tasks we're interested in in AP, including anything that acts on a calexp, use visits. Though I haven't seen an explicit specification to that effect...

tests/test_pipelineTaskTests.py
kfindeisen: It's considered good practice to have a separate function for each test (testing purists would say any function with multiple asserts is too big, though I wouldn't go that far). It means you get more information up-front about what is working and what is failing, and you don't have to worry (in theory) about different cases interfering with each other.

I can try to factor the setup code into separate functions. The main problem is that in a few cases (like the "WithRun" tests) the test really does depend on the details of the mock data.

mrawls: That's fair, thanks for the context about test good practice. I don't want you to spend too much time on this, but if you can factor out a bit more of the setup code (e.g., so you don't define `inA = butlerTests.MetricsExample(data=[1, 2, 3])` multiple times), that would make it easier to read.

mrawls:
I'd be happier with `testUtils.py`, thanks. Is there any place we could explicitly document "These are tools for writing unit tests"? Maybe it should be obvious, but that simple sentence made the whole endeavor about 200% clearer to me.

python/lsst/pipe/base/tests.py
kfindeisen: Hmm... `verifyRunQuantum` or `testRunQuantum` are also a bit misleading, because they imply the function actually does any testing (when it's really just "I want to run `runQuantum` without all the quantumgraph/workflow baggage").

`driveRunQuantum` is accurate, but I'm not sure it's clear at first glance.

kfindeisen:
Hi @mrawls, can you take another look? I've made the following changes:
* new documentation for the public code; built version on Jira
* I've replaced the old true/false `verifyOutputConnections` with an exception-based `assertValidOutput`, so that the user has some feedback on what went wrong. The change is a separate commit for now, in case it needs to be reverted; I plan to squash it before merging.

doc/lsst.pipe.base/testing-a-pipeline-task.rst
```diff
@@ -0,0 +1,191 @@
...
+   butlerTests.addDatasetType(
+       repo, "OutputType", {"instrument", "visit", "detector"},
+       "ExposureF")
+
+   ...
+
+   # Set up what we need
+   dataId = {"instrument": "notACam", "visit": 101, "detector": 42}
+   butler = butlerTests.makeTestCollection(repo)
+   task = SuperTask()
```
mrawls: Did you mean `PipelineTask`?

python/lsst/pipe/base/testUtils.py
```diff
@@ -213,14 +213,13 @@ def verifyOutputConnections(task, result):
...
         # multiple
         if connection.multiple:
             if not isinstance(output, collections.abc.Sequence):
-                return False
+                raise AssertionError(f"Expected {name} to be a sequence, got {output} instead.")
         else:
             if isinstance(output, collections.abc.Sequence) \
                     and not issubclass(connections.storageClass.pytype, collections.abc.Sequence):
-                return False
+                return AssertionError(f"Expected {name} to be a single value, got {output} instead.")
```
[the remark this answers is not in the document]
kfindeisen: How on earth did that get through testing???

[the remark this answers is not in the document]
kfindeisen: Turns out the test was raising `AssertionError` over something completely different. In reality, this branch of the code never worked, even in the non-exception version.

Thanks for catching this!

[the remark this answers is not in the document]
mrawls: Partial credit to @parejkoj who was looking over my shoulder as I did the review and pointed it out :)

PR #118, opened 2020-02-12
DM-23477: Allow ScalarError to be constructed from a single parameter

timj:
When this class is used in multiprocessing environment ScalarError
must be pickled. When being pickled only the single string passed
to TypeError is serialized. This leads to an initialization error
when the object is recreated and breaks error reporting.

Change things so that a single parameter acts like any other
standard exception.


natelust:
I prefer to not implement reduce if we can avoid it. I guess we never had a failure at this specific point with multiprocessing turned on before, I cant see any minor change to python version could cause this to start failing.

PR #121, opened 2020-03-05
DM-22655: Genericize gen2to3.py to be useable with any gen2 repo

parejkoj:


parejkoj:
@timj : sorry, I forgot to create a PR for this one when you reviewed the others.

python/lsst/pipe/base/graphBuilder.py
```diff
@@ -677,8 +684,13 @@ def fillQuanta(self, registry, inputCollections, *, skipExisting=True):
                                 expand=True,
                             )
                         )
+                    if refs == []:
```
natelust: Should not compare equality to an empty list, either check the length equal to zero, or just do `if refs` as lists are truthy.

python/lsst/pipe/base/graphBuilder.py
```diff
@@ -677,8 +684,13 @@ def fillQuanta(self, registry, inputCollections, *, skipExisting=True):
...
+                        _LOG.warn("queryDatasets returned an empty list for %s", datasetType)
```
natelust: Because this is a warning it might not be a big deal, but some tasks may be just fine with not getting the prerequisite it asks for and makes use of that behavior. I think there are some examples of this in the ISR stuff Chris has been working on. In that case this warning would just create noise. That is one of the reasons for the lookup function. If a task wants it can throw a message or exception if it does not get what it wants.

parejkoj: I guess we'll just have to be better about managing that on the Task end; I added this warning here for a prereq that was in fact required in the Task but where I didn't know it was being returned empty. I'll remove this for now, but we should think about how to better manage this kind of error handling.

PR #123, opened 2020-03-25
DM-23616: Run converted ap_verify testdata through gen3 pipeline

kfindeisen:
This PR fixes bugs and missing features in the PipelineTask test framework (`pipe.base.testUtils`) in response to trying to apply it to `CalibrateTask` on lsst/pipe_tasks#362.

### DIFF
--- a/python/lsst/pipe/base/testUtils.py
+++ b/python/lsst/pipe/base/testUtils.py
@@ -140,7 +140,14 @@ def _refFromConnection(butler, connection, dataId, **kwargs):
     """
     universe = butler.registry.dimensions
     dataId = DataCoordinate.standardize(dataId, **kwargs, universe=universe)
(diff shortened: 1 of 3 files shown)


python/lsst/pipe/base/testUtils.py
```diff
@@ -139,7 +143,13 @@ def _refFromConnection(butler, connection, dataId, **kwargs):
     """
     universe = butler.registry.dimensions
     dataId = DataCoordinate.standardize(dataId, **kwargs, universe=universe)
-    datasetType = connection.makeDatasetType(universe)
+
+    # copied from TaskDatasetTypes.fromTaskDef
+    if "skypix" in connection.dimensions:
```
parejkoj: A short explanation as to why this specialization is necessary would be useful.

PR #125, opened 2020-04-09
DM-23173: change to binary operators at beginning of new line

r-owen:
Also:
- Update to use f strings
- Eliminate "u" strings (the "u" is meaningless in Python 3)
- Standardize on sys.stderr.write instead of print(..., file=sys.stderr)
- Change remaining % formatted log messages to deferred formatting (most already were)

### DIFF
--- a/python/lsst/pipe/base/argumentParser.py
+++ b/python/lsst/pipe/base/argumentParser.py
@@ -149,7 +149,7 @@ class DataIdContainer:
         try:
             idKeyTypeDict = butler.getKeys(datasetType=self.datasetType, level=self.level)
         except KeyError as e:
-            msg = "Cannot get keys for datasetType %s at level %s" % (self.datasetType, self.level)
+            msg = f"Cannot get keys for datasetType {self.datasetType} at level {self.level}"
             raise KeyError(msg) from e
 
         for dataDict in self.idList:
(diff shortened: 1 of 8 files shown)


python/lsst/pipe/base/argumentParser.py
```diff
@@ -1039,7 +1038,7 @@ def write(self, showStr):
             matHistory = re.search(r"^(?:config.)?(.+)?", showArgs)
             globPattern = matHistory.group(1)
             if not globPattern:
-                print("Please provide a value with --show history (e.g. history=*.doXXX)", file=sys.stderr)
+                sys.stderr.write("Please provide a value with --show history (e.g. history=*.doXXX)\n")
```
timj, line 1041: I don't like this at all. What's the justification given that `file=sys.stderr` is a good option already.

r-owen: The justification is that the existing code is a mix of the two styles (even within the same module). I feel write is a bit cleaner, though it does risk forgetting the final \n. In any case I think we should pick one and stick to it. I'm willing to switch to print if you feel strongly about it. I find the mix confusing.

r-owen: Since you prefer print and I don't have strong feelings either way I updated the code to use print everywhere. It all appeared to be in the one module.

PR #134, opened 2020-07-01
DM-25447: Change how comparison with registry DatasetType happens

timj:
We now require a parent storage class for components.

### DIFF
--- a/python/lsst/pipe/base/pipeline.py
+++ b/python/lsst/pipe/base/pipeline.py
@@ -483,15 +483,30 @@ class TaskDatasetTypes:
                                          f"connections ({rest1}) are inconsistent with those in "
                                          f"registry's version of this dataset ({rest2}).")
                 else:
-                    datasetType = DatasetType(c.name, registry.dimensions.extract(dimensions),
-                                              c.storageClass)
+                    # Component dataset types are not explicitly in the
+                    # registry.  This complicates consistency checks with
+                    # registry and requires we work out the composite storage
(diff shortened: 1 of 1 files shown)


python/lsst/pipe/base/pipeline.py
```diff
@@ -483,15 +483,29 @@ def makeDatasetTypesSet(connectionType, freeze=True):
...
+                    # class.
+                    registryDatasetType = None
                     try:
                         registryDatasetType = registry.getDatasetType(c.name)
                     except KeyError:
+                        compositeName, componentName = DatasetType.splitDatasetTypeName(c.name)
+                        parentStorageClass = "UnknownParent" if componentName else None
+                        datasetType = DatasetType(c.name, registry.dimensions.extract(dimensions),
+                                                  c.storageClass,
+                                                  parentStorageClass=parentStorageClass)
```
TallJimbo, line 498: I'm surprised this is necessary - didn't you add logic to `DatasetRecordStorageManager` to handle components there?  `Registry.getDatasetType` should just be delegating to that.  Maybe that logic needs to be extended to add the parent storage class?

timj: This is the branch for wear the registry did not have the dataset type defined so we have to define it ourselves.

TallJimbo: Ah, that makes sense.  Sorry for the noise.

PR #135, opened 2020-07-02
DM-25767 Add checking for duplicate keys in pipeline definitions

natelust:


python/lsst/pipe/base/pipelineIR.py
```diff
@@ -27,6 +29,28 @@
 import warnings
 
 
+class PipelineYamlLoader(yaml.SafeLoader):
+    """This is a specialized version of yaml's SafeLoader. It checks and raises
+    an exception if it finds that there a multiple instances of the same key
```
TallJimbo, line 34: typo: "there a multiple"

PR #143, opened 2020-09-11
DM-26371 Introduce new QuantumGraph object

natelust:


python/lsst/pipe/base/pipeline.py
```diff
@@ -111,6 +112,18 @@ def __str__(self):
...
+    def __eq__(self, other: TaskDef) -> bool:
+        return self.taskName == other.taskName and\
+            self.config == other.config and\
+            self.taskClass == other.taskClass and\
+            self.label == other.label
+
+    def __hash__(self):
+        with io.StringIO() as sio:
+            self.config.saveToStream(sio)
+            configHash = hash(sio.getvalue())
```
andy-slac, line 124: I'm not very comfortable with this, I'm not convinced that formatting pex_config will always produce the same output for two identical configs. Can you just ignore config and not include it in a hash?

python/lsst/pipe/base/pipeline.py
```diff
@@ -111,6 +112,18 @@ def __str__(self):
...
+        return hash(self.taskName) ^ configHash ^ hash(self.taskClass) ^ hash(self.label)
```
andy-slac, line 125: Hmm, usually people do `hash(tuple(....))` on multiple values, I don't know if `^` is better or worse.

andy-slac (approved):
Looks OK, few minor comments, but I'd prefer hash methods be improved for TaskDef and QuantumNode classes.

python/lsst/pipe/base/connections.py
```diff
@@ -497,7 +498,8 @@ def adjustQuantum(self, datasetRefMap: NamedKeyDict[DatasetType, typing.Set[Data
         return datasetRefMap
 
 
-def iterConnections(connections: PipelineTaskConnections, connectionType: str) -> typing.Generator:
+def iterConnections(connections: PipelineTaskConnections,
+                    connectionType: Union[str, Iterable[str]]) -> typing.Generator:
```
TallJimbo, line 502: Making the return type `typing.Generator[BaseConnection, None, None]` is a lot more useful than saying `Generator` with no generics.  And I know we might differ on this, but if the user shouldn't care that this is implemented as a generator, consider `Iterator` instance

python/lsst/pipe/base/graph/_implDetails.py
```diff
@@ -0,0 +1,85 @@
...
+    This is an internal implementation detail
+    """
+    def add(self, element: _T):
+        super().add(element)
+
+    @property
+    def first(self) -> Optional[_T]:
+        if len(self) == 0:
+            return None
+        return next(iter(self))
```
TallJimbo, line 51: I think this would be a lot better as free function; you don't want to have to copy some external `set` (or even convert some other `set`-like object into one of these) just to be able to use this method.

python/lsst/pipe/base/graph/graph.py
```diff
@@ -0,0 +1,779 @@
...
+                else:
+                    cumHash ^= _hashDsRef(item)
+        if self.quantum.outputs:
+            for item in self.quantum.outputs.values():
+                if isinstance(item, abc.Iterable):
+                    for value in item:
+                        cumHash ^= _hashDsRef(value)
+                else:
+                    cumHash ^= _hashDsRef(item)
+        return cumHash
```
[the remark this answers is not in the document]
TallJimbo: I'm also curious what problem `__eq__` and `__hash__` solve here.  Is it just for test comparisons?  Under what contexts are these being compared (i.e. within one graph, the task label and data ID would be sufficient for uniqueness).

If we do need some complicated hierarchical hash, I'd be inclined to implement more of it on the objects `Quantum` holds, but it seems even better to try to avoid needing it.

python/lsst/pipe/base/graph/graph.py
```diff
@@ -0,0 +1,779 @@
...
+        self._count: Optional[int] = None
+
+        self._buildGraphs()
+
+    def _buildGraphs(self):
+        """Builds the graph that is used to store the relation between tasks,
+        and the graph that holds the relations between quanta
+        """
+        # Datastructures used to identify relations between components; TaskDef
+        # for taks, and QunatumNode for the quanta
```
TallJimbo, line 135: typo: taks

python/lsst/pipe/base/graph/graph.py
```diff
@@ -0,0 +1,779 @@
...
+        ------
+        `IndexError`
+            If the requested nodeNumber is greater than the number of nodes in
+            the graph.
+        """
+        return self._nodeList[nodeNumber]
+
+    @property
+    def inputQuanta(self) -> List[QuantumNode]:
+        """Makes a `list` of all `QuantumNode`s that are 'input' nodes to the
```
TallJimbo, line 268: Start-of-docstring verbs [should be in the imperative voice](https://developer.lsst.io/python/numpydoc.html#writing-summaries-for-functions-and-methods), i.e. "Make" rather than "Makes" here (also true for several methods below, with "Returns" -> "Return").

python/lsst/pipe/base/graph/graph.py
```diff
@@ -0,0 +1,779 @@
...
+    @property
+    def outputQuanta(self) -> List[QuantumNode]:
+        """Makes a `list` of all `QuantumNode`s that are 'output' nodes to the
+        graph, meaning those nodes have no nodes that depend them in the graph.
+
+        Returns
+        -------
+        outputNodes : list of `QuantumNode`
+        """
+        return [q for q, n in self._connectedQuanta.out_degree if n == 0]
```
TallJimbo, line 287: If this needs to return a container because the convenience is more important than guarding against it being a very large list, I'd recommend a tuple rather than a `list`, because it's a property and it's all too easy to think e.g. `graph.outputQuanta.append(quantum)` is doing something useful if it doesn't raise.  It's not always worthwhile to guard against that, but doing it here is trivial.

But it's also worth asking whether this should just be a generator or return a generator expression.

Same comments apply (sometimes about `set` vs. `frozenset` vs. a lazy iterator) in many other accessors below.

python/lsst/pipe/base/graph/graph.py
```diff
@@ -0,0 +1,779 @@
...
+        """Returns all the `Quantum` associated with a `TaskDef`
+
+        Parameters
+        ----------
+        taskDef : `TaskDef`
+            The `TaskDef` for which `Quantum` are to be queried
+
+        Returns
+        -------
+        set of `Quantum`
```
TallJimbo, line 311: I thought return values needed to be named, just like parameters.  I may be wrong, but please check that this renders as you expect.  Also, `set` should have backticks, as should `list` in many other docstrings here.

python/lsst/pipe/base/graph/graph.py
```diff
@@ -0,0 +1,779 @@
...
+        """
+        return set(self._quanta[taskDef])
+
+    def tasksWithInput(self, datasetTypeName: DSTypeName) -> List[TaskDef]:
+        """Finds all tasks that have the specified dataset type name as an
+        input.
+
+        Parameters
+        ----------
+        datasetTypeName : `DSTypeName`
```
TallJimbo, line 323: I'm a little uncomfortable including a `NewType` in docstrings.  It'll be unfamiliar to most developers, and while they can ignore type annotations type annotations if they find them confusing, making the fact that this is just a `str` at runtime so indirected from usage point seems bad for them.

python/lsst/pipe/base/graph/_implDetails.py
```diff
@@ -0,0 +1,85 @@
...
+from dataclasses import dataclass, field
+from typing import (Generic, Mapping, Optional, Set, TypeVar, Generator, Tuple, NewType, Union)
+
+# Generic type parameters
+_T = TypeVar("_T")
+_U = TypeVar("_U")
+_S = TypeVar("_S")
+
+# NewTypes
+DSTypeName = NewType("DSTypeName", str)
```
TallJimbo, line 34: I hate to be That Guy, but our conventions say this really should be spelled out as `DatasetTypeName`, or at least `DsTypeName` (and while I find the latter detestable, the former really is not too much longer and much more readable, at least to me).

python/lsst/pipe/base/graph/graph.py
```diff
@@ -0,0 +1,779 @@
...
+        output.
+
+        Parameters
+        ----------
+        datasetTypeName : `DSTypeName`
+            A DSTypeName representing the name of a dataset type to be queried
+
+        Returns
+        -------
+        `TaskDef` or None
```
TallJimbo, line 350: `None` should always be in backticks in documentation.

python/lsst/pipe/base/graph/graph.py
```diff
@@ -0,0 +1,779 @@
...
+        results = self.tasksWithInput(datasetTypeName)
+        output = self.taskWithOutput(datasetTypeName)
+        if output is not None:
+            results.append(output)
+        return results
+
+    def findTaskDefByName(self, taskName: str) -> List[TaskDef]:
+        """Determines which `TaskDef`s in this graph are associated with a `str`
+        representing a tasks name (looks at the taskName property of
+        `TaskDefs`).
```
TallJimbo, line 389: `TaskDefs` is plural and won't render properly.  FWIW, I'm not sure
```
`TaskDef`s
```
(which I've seen a few times as well) will either.

python/lsst/pipe/base/graph/graph.py
```diff
@@ -0,0 +1,779 @@
...
+        reconstructed with this information, and it preseves the ordering of
+        the graph ndoes.
+        """
+        q2Task = {}
+        # Here the config is turned into a string instead of associating the
+        # dictionary with the instance, because on unpickling the config
+        # class is known, and decreases un-persisting time for configs by
+        # about a factor of 2
+        for quantum, task in self._quantumToTask.items():
+            q2Task[quantum] = (task.label, task.config)
```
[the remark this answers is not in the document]
TallJimbo: _If_ saving the `TaskDef` instance here would still mean we only save the `TaskDef` once (due to pickles internal memoization), I'm all for it.  If not (I don't recall the details of how pickle works, and haven't quite followed all of the data structures here), we should be saving only the label, as that's sufficient to uniquely identify the `TaskDef` within the graph, and that's sufficient to find the config.  We definitely don't want to be saving the same config instance multiple times without memoization, as that'd be hugely wasteful.

python/lsst/pipe/base/graph/graph.py
timj: yes, they need to be named.

python/lsst/pipe/base/graph/graph.py
timj: I think it's
```
`TaskDef`\ s
```

python/lsst/pipe/base/graph/graph.py
natelust: Sorry I was used to numpy-docs where they are not required and either forgot or did not know that we had a different standard.

python/lsst/pipe/base/graph/_implDetails.py
```diff
@@ -0,0 +1,81 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+from __future__ import annotations
+
+__all__ = ("_DatasetTracker", "_OutputSet", "_datasetDictToEdgeIterator")
```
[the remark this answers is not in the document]
natelust: At some point there was issues in some code I wrote where the documentation was messed up because of imports being exposed at the module scope, and as explicit is better than implicit I like to include an `__all__`

python/lsst/pipe/base/graph/graph.py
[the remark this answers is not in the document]
natelust: So this was a bit rushed (you may remember my comments at the middleware meeting) and there is a few other things to be considered here. One such use is if you had a node that came from one of two graphs, and you wanted to know which graph that node came from, it is important to know something about the whole `QuantumNode`. I kept going back and forth on this, but thinking about this response has convinced me that TaskDef should go back into `QuantumNode`, as thinks like label are important to deciding if nodes are equals. Now for this constraint you strictly dont need all the logic in the `QuantumNode` object, some could be in the `__contains__` method on `QuantumGraph` or something, but I think that A) having it all centralized is good (lets you compare nodes w/o a graph) B) keep it fast by exploiting hash algorithm.

Other uses are things like comparing the equality of QuantumGraphs. This is useful since we support subsetting, and may want to see if one subset is equal to another. Related to this, a discussion with Michelle happened yesterday raised the possibility of keeping NodeNumber constant in a subset, which would allow a node to be better tracked back to the original graph. But there are other considerations with this, like should NodeNumber be part of a hash and equality, or should it just be `Quantum` and `TaskDef`? That would allow you to see if a node is in a graph w/o caring about where it is located within the graph (This is currently what I am leaning towards), in which case NodeNumber is just a convenient number to locate a Node w/n a `QuantumGraph` instance, but has no bearing on two graphs being considered equal (which I like).

As it is now Quantum contains a lot of mutable things that unimportant to the equality of a Node (at least I have conceived it currently) it seems better to have a hash on the node object on the attributes that are important. There may be cases where we do care about those attributes to hash/equals a Quantum. So unless we refactor quantum too, I would like to leave this hear.

Most of this is thinking 'out loud' so if it is rambling I apologize. A lot of this code has grown organically as new features the graph should have were discovered and so were not "designed" as a whole from the beginning, something I am trying to go back and reason through now.

natelust:
A lot of comments on the implementation details (which are great) but have people looked at the interface? How does it feel to people, are there methods you think that are missing that could be useful?

MichelleGower (commented):
As for whether the class has all needed functionality, that is a difficult question to answer (especially since the networkx library used internally has a large set of graph operations).  I pointed out one function I specifically noticed wasn't available.  The good news is that I was able to successfully prototype changes to BPS to use this.  And the BPS code <-> QuantumGraph interaction happens more at the API level than before (e.g., save and load, and some because BPS was just never modified to use new QuantumGraph API).

python/lsst/pipe/base/graph/_implDetails.py
```diff
@@ -0,0 +1,85 @@
...
+    considered connected by the graph.
+    """
+    for datasetType, entry in datasetDict.items():
+        if len(entry.output) > 1:
+            raise ValueError(f"Datasets can only be produced by one task: "
+                             f"{entry.output} each claim to produce {datasetType}")
+        output = entry.output.first
+        # If there is no inputs and only outputs (likely in test cases or
+        # building inits or something) use None as a Node, that will then be
+        # removed later
```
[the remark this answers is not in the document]
natelust: In practice I can't imagine a pipeline task that would get run but produce nothing, but in unit tests this certainly happens. This is handled by the `first` method of the `_OutputSet` that will return `None` in the case it is empty, which gets removed as well, and is commented elsewhere. This comment reflects why the `inputs` variable is created here.

python/lsst/pipe/base/graph/graph.py
```diff
@@ -0,0 +1,779 @@
...
+            taskGraph.remove_node(None)
+
+        self._nodeList = state['nodes']
+        self._quanta = dict(quanta.items())
+        self._quantumToTask = quantumToTask
+        self._connectedQuanta = connectedQuanta
+        self._datasetRefDict = datasetRefDict
+        self._datasetDict = datasetDict
+        self._taskGraph = taskGraph
+        self._count = None
```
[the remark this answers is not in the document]
natelust: Dropped in favor of doing things differently with changes to construction

python/lsst/pipe/base/pipeline.py
natelust: No, as it is the only thing that can disambiguate a TaskDef created in the same pipeline with the same input repos, only differing by a config option. This is useful for comparing one QuantumGraph to another.

andy-slac:
One comment regarding interfaces which also touches on sub-graphs and provenance.

One thing that I think sticks out in a new interface is the `QuantumNode.nodeNumber` thing. I think it is supposed to work OK for single graph but using it across multiple graphs is clearly non-trivial. If we need to identify nodes that may be shared across multiple graphs we need better abstraction to replace `nodeNumber`. Using numbers for identification is probably not going to work, we would need something like globally unique IDs (`uuid`) for the node itself, and maybe for the "origin" graph where the node was originally created. And then of course you need to track them down through all transformations/serializations/etc.

python/lsst/pipe/base/pipeline.py
andy-slac: The hash is not required to be unique, the requirement is that objects that compare as equal have the same hash, but does not say that objects that are different must have different hash (and hash in general is not unique). You can drop config from hash and still keep it in `__eq__` (comparing configs is well defined). And in the same pipeline tasks have to have different labels so hash that includes labels should already be OK.

python/lsst/pipe/base/graph/_implDetails.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+@dataclass
+class _DatasetTracker(Generic[_U]):
+    """Container class that holds sets of inputs, and _OutputSets of outputs.
+    This class exposes an `all` property which is useful for looking at the set
+    of all inputs and outputs
+    """
+    inputs: Set[_U] = field(default_factory=set)
+    output: _OutputSet[_U] = field(default_factory=_OutputSet)
```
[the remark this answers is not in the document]
natelust: because where this is used (tracking dsTypes / quanta) there can be multiple inputs, but only one thing should output it, so it is named such, it is a full container to catch cases where there is an error rather than clobbering. I could probably roll all these things together into a unified class though, which might make a few things cleaner looking.

python/lsst/pipe/base/graph/graph.py
```diff
@@ -0,0 +1,779 @@
...
+        # Attempting to create a graph that contains a quantum unconnected from
+        # any other node will create an edge from None to the task, but None is
+        # not a node we actually want so remove it here. This is a consequence
+        # of using add_edges from and is an implementation detail.
+        if None in self._connectedQuanta.nodes():
+            self._connectedQuanta.remove_node(None)
+
+    @property
+    def taskGraph(self) -> nx.DiGraph:
+        """Returns a graph representing the relations between the tasks inside
```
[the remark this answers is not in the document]
natelust: Because QuantumGraph is a extension/abstraction around the `graph` property (if you iterate over both containers you get the relations between `QantumNodes` so in some cases they could be interchangeable. The `taskGraph` property is just a convenience that exposes some information, and could never be used as or confused with `QuantumGraph`

python/lsst/pipe/base/graph/graph.py
```diff
@@ -0,0 +1,779 @@
...
+
+        # Graph of task relations, used in various methods
+        self._taskGraph = nx.DiGraph()
+        # _datasetDictToEdgeIterator takes the stored inputs and outputs
+        # associated with each key in the default dict to yield a tuple which
+        # represents an edge between each yielded node
+        self._taskGraph.add_edges_from(_datasetDictToEdgeIterator(self._datasetDict))
+        # Attempting to create a graph that contains a single task will create an
+        # edge from None to the task, but None is not a node we actually want
+        # so remove it here. This is a consequence of using add_edges from and
```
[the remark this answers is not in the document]
natelust: :+1: As said in another comment, I think factoring all this away would probably make things cleaner, and will do that.

natelust:
I intended for `nodeNumber` to be only meaningful w/n one creation of a graph (and live through pickling), Something that you say give something a single pickle of the `QuantumGraph` and a list of node numbers to execute. This would save the need for creating many pickles of subgraphs (but would be a `pipetask` API change so I was just putting it in place for now). Then Michelle suggested it would be useful to preserve in sub setting for debugging reasons, which is reasonable. However, the changing nature of this property does have me re-evaluating it. That does not mean I am rushing to keep or get rid of it, but as you say it is a much more complex thing to think about.

python/lsst/pipe/base/graph/graph.py
```diff
@@ -0,0 +1,779 @@
...
+
+        Returns
+        -------
+        graph : `networkx.Digraph`
+            Internal datastructure that holds relations of `QuantumNode`s
+        """
+        return self._connectedQuanta
+
+    def taskDefForNode(self, node: QuantumNode) -> TaskDef:
+        """Retrives the TaskDef associated with a `QuantumNode`.
```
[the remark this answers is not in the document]
natelust: Method removed

python/lsst/pipe/base/graph/graph.py
```diff
@@ -0,0 +1,779 @@
...
+    def countQuanta(self) -> int:
+        """Returns the number of nodes in the graph
+
+        Returns
+        -------
+        int
+            Number of nodes in the graph
+        """
+        if self._count is None:
+            self._count = int(sum(len(values) for values in self._quanta.values()))
```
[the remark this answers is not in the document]
natelust: moved to build graph, and removed count method entirely as it is redundant with `__len__`

python/lsst/pipe/base/graph/graph.py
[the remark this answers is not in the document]
natelust: I wrote a test and verified that Pickle was not duplicating storage space for TaskDefs, and so I simplified things

python/lsst/pipe/base/graph/_implDetails.py
natelust: Dropped in a refactor

python/lsst/pipe/base/graph/graph.py
```diff
@@ -0,0 +1,779 @@
...
+            for item in self.quantum.outputs.values():
+                if isinstance(item, abc.Iterable):
+                    for value in item:
+                        cumHash ^= _hashDsRef(value)
+                else:
+                    cumHash ^= _hashDsRef(item)
+        return cumHash
+
+
+class QuantumGraph:
```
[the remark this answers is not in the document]
natelust: I have added TaskDef back to `QuantumNode`, it is something I kept going back and forth with. I think that addresses most of your concerns. Related though is the question if the `taskGraph` property and `iterTaskGraph` method should exist, and if no one objects I would like to keep it as a convenience, so if someone wants to interact with the `QuantumGraph` for only specific tasks (such as subset), they do not need to iterate over the whole graph to build the associations themselves.

python/lsst/pipe/base/graph/_implDetails.py
```diff
@@ -0,0 +1,85 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+from __future__ import annotations
+
+__all__ = ("_DatasetTracker", "_OutputSet", "_datasetDictToEdgeIterator")
+
+from dataclasses import dataclass, field
+from typing import (Generic, Mapping, Optional, Set, TypeVar, Generator, Tuple, NewType, Union)
+
+# Generic type parameters
```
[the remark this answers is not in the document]
natelust: In the new implementation it is down to two `TypeVar`s. From what it seems, it does currently seem to be the convention that generic parameters be short one or two letter variables. The new implementation has restricted `TypeVar`s which does give more guidance. If this was not an implementation detail module that is not exported anywhere, I would generally agree that this information would probably belong in the class level docstring.

python/lsst/pipe/base/graph/_implDetails.py
```diff
@@ -0,0 +1,85 @@
...
+_T = TypeVar("_T")
+_U = TypeVar("_U")
+_S = TypeVar("_S")
+
+# NewTypes
+DSTypeName = NewType("DSTypeName", str)
+
+
+class _OutputSet(set, Generic[_T]):
+    """Simple set wrapper that provides a "first" method
```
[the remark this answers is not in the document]
natelust: This class is now gone

python/lsst/pipe/base/graph/_implDetails.py
```diff
@@ -0,0 +1,81 @@
...
+    to sets, that gets the first element of a set as if
+    it was being iterated over.
+
+    This is an internal implementation detail
+    """
+    def add(self, element: _T):
+        super().add(element)
```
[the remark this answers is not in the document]
natelust: This class is gone now

python/lsst/pipe/base/graph/graph.py
```diff
@@ -0,0 +1,779 @@
...
+        self._nodeList = state['nodes']
+        self._quanta = dict(quanta.items())
+        self._quantumToTask = quantumToTask
+        self._connectedQuanta = connectedQuanta
+        self._datasetRefDict = datasetRefDict
+        self._datasetDict = datasetDict
+        self._taskGraph = taskGraph
+        self._count = None
+
+    def __eq__(self, other: object) -> bool:
```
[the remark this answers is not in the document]
natelust: I think equality of graphs is needed, as it allows comparison of things like subsets, or graphs prepared at different times. I have removed any dependency on node identifications.

python/lsst/pipe/base/graph/graph.py
```diff
@@ -0,0 +1,779 @@
...
+        taskDefs: Dict[str, TaskDef] = {}
+        datasetRefDict: DefaultDict[DatasetRef, _DatasetTracker[QuantumNode]] =\
+            defaultdict(_DatasetTracker[QuantumNode])
+        for quantumNode in nodeList:
+            quantum = quantumNode.quantum
+            nodeNumber = quantumNode.nodeNumber
+            label, config = q2Task[nodeNumber]
+            # This dict is simply to ensure that there is only one TaskDef
+            # created for each label
+            taskDef = taskDefs.setdefault(label, TaskDef(quantum.taskName, config, quantum.taskClass, label))
```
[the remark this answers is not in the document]
natelust: There is, but there will only be small numbers of TaskDefs, even if there are hundreds its small for something like this. I feel the simplicity of the code outweighs the small amount of extra runtime.

python/lsst/pipe/base/pipeline.py
natelust: But the same pipeline might be run two ways, I don't see anything that says the TaskDefs should have the context of anything else other than themselves. I have verified creating a new config and turning it to a string over 2000 times leads to the same serialization. This may not be true from one version of the stack to another, but that is possible with any stack version changes.

python/lsst/pipe/base/pipeline.py
natelust: So it is close to the same thing, but without the overhead of creating temporary containers. This is what is recommended in official documentation for `__hash__` in python (though its not exclusively restricted to this method)

python/lsst/pipe/base/connections.py
natelust: I didn't think to expand this when making the change, but I updated it now

python/lsst/pipe/base/graph/graph.py
natelust: I looked and changed this were it seemed appropriate

python/lsst/pipe/base/graph/graph.py
natelust: I changed it to str, since that is basically what it is, but added notes in the documentation about ability to use DatasetTypeName for type safety in type checking

andy-slac (commented):
I have done a quick re-review, small bunch of comments. I understand it's still not final, so I don't check "Approve" yet.

python/lsst/pipe/base/pipeline.py
timj: Shouldn't we be letting `Config` define how it handles equality and `__hash__`?

andy-slac: I'm not sure `Config` wants to be hashable. It is a mutable structure unless you freeze it and in general making complicated dynamic data structures hashable is not worth trouble in my opinion. In this particular case we don't even need its hash, other attributes of `TaskDef` are sufficient for building reasonable hash.

natelust: Any you make a good point, I was so focused on the tree I was not seeing the forest (thinking too closely of equals and not hash interplay). Thanks for pointing out the data model again, equal things must hash the same, but the converse of hash must be everything in equals is not necessary (if you are fine with a bit higher collision rate). I'll update this now.

python/lsst/pipe/base/graph/graph.py
```diff
@@ -0,0 +1,641 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+from __future__ import annotations
+import io
```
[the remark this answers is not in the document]
natelust: Thanks, if I use something i have not imported my editor is "helpful" and adds the import. I'll try to keep a closer eye on this.

python/lsst/pipe/base/graph/graph.py
```diff
@@ -0,0 +1,641 @@
...
+        """
+        return [q for q, n in self._connectedQuanta.out_degree if n == 0]
+
+    @property
+    def allDatasetTypes(self) -> Tuple[DatasetTypeName, ...]:
+        """Return all the `DatasetTypeNames` that are contained inside the graph.
+
+        Returns
+        -------
+        list of `DatasetTypeName`
```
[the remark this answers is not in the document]
natelust: I really cant wait until we upgrade our sphinx, it supposedly has the ability to get the types from the annotations

python/lsst/pipe/base/pipeline.py
natelust: It seems like the suggestion has changed sometime between python 3.4 (the last time I looked at it) and python 3.8 docs. I will update to use the new suggestion of making a tuple

andy-slac (approved):
Looks good, few minor comments

tests/test_quantumGraph.py
```diff
@@ -175,6 +176,8 @@ def testGetQuantumNodeByNodeId(self):
         inputQuanta = tuple(self.qGraph.inputQuanta)
         node = self.qGraph.getQuantumNodeByNodeId(inputQuanta[0].nodeId)
         self.assertEqual(node, inputQuanta[0])
+        wrongNode = NodeId(15, BuildId("alternative build Id"))
+        self.assertRaises(IncompatibleGraphError, self.qGraph.getQuantumNodeByNodeId, wrongNode)
```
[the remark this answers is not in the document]
timj: I agree completely. You know exactly what is being tested in the context manager version and it's easy to make it trigger simply by commenting out the with and putting in an `if True:` (so you don't need to change the indent).

python/lsst/pipe/base/graph/quantumNode.py
```diff
@@ -84,42 +82,12 @@ class QuantumNode:
     def __eq__(self, other: object) -> bool:
         if not isinstance(other, QuantumNode):
             return False
```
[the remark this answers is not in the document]
timj: but surely you don't want python to do something so clever that somehow it ends up returning True?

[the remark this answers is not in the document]
andy-slac: I don't but maybe someone else will 🙂 `NotImplemented` utility is in raising an exception for non-comparable types, maybe we do not want that, but OTOH I have no idea why it's OK to compare Quantum and non-Quantum.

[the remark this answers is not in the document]
timj: Thanks. I just read up on this and I see that returning `NotImplemented` causes python to switch round the equality and try to see if `other.___eq__(self)` gives a different answer. Then it seems it tries an `is` check and eventually returns false.  So it seems that returning `NotImplemented` is the "right" thing to do but also doesn't it guarantee to slow everything down when we already know the answer is false?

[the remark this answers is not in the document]
TallJimbo: I think returning `False` is fine here; it's the equivalence-relationship analog of marking a class as "final" for inheritance relationships.  So it's not something you want to do accidentally - which is why `NotImplemented` is the general recommendation - but a useful way to express "no future type shall ever compare as equal to instances of this type".

I actually think that making disparate types actually compare as equal to each other is almost always more trouble than its worth (and some Python devs recognized this; note that a list never compares as equal to a tuple), so I'd personally be willing to go farther and say that we should always consider `False` to be a better default than `NotImplemented`, and the Python "general recommendation" of `NotImplemented` is actually slightly bad advice.

MichelleGower (approved):
Was able to successfully use this in ctrl_bps.  Would use a subset by NodeId list if available, but only a couple of lines without it.   Approve merging changes.

PR #144, opened 2020-09-11
DM-26591: Include instrument data ID value when provided on pipetask command-line or Pipeline yaml file

n8pease:


python/lsst/pipe/base/graphBuilder.py
```diff
@@ -817,7 +819,79 @@ def makeGraph(self, pipeline, collections, run, userQuery):
...
+        query : `str`
+            The query string.
+
+        Returns
+        -------
+        instrumentName : `str` or `None`
+            The name of the instrument in the query string, or `None` if an
+            instrument is not named.
+        """
+        instrumentRegex = r"instrument *="
```
timj, line 893: As written this does not need to be a raw string. Saying that, in regexes I tend to prefer `\s*` rather than ` *` for whitespace matching.  Also though, why can't this regex also include the instrument in it as a group so that you can ask for group 1 directly, rather than taking the unmatched part and splitting on space after?

Something like:

```python
instrumentRegex =r"instrument\s*=\s*[\"'](.*?)[\"']"
```

works for me and the instrument is then `match.group(1)`.

python/lsst/pipe/base/graphBuilder.py
```diff
@@ -817,7 +819,79 @@ def makeGraph(self, pipeline, collections, run, userQuery):
...
+        Raises
+        ------
+        RuntimeError
+            If the query contains an instrument restriction and the pipeline
+            has an instrument and the two do not match.
+        """
+        instrument = pipeline.getInstrument()
+        if instrument is not None:
+            if isinstance(instrument, str):
+                instrument = getInstrument(pipeline._pipelineIR.instrument, self.registry)
```
timj, line 858: isn't this `pipeline._pipelinesIR` exactly what `pipeline.getInstrument()` does above? I'm trying to work out why the argument here isn't `instrument`.

TallJimbo: I think it's fairly safe to add the instrument to the query without checking.  We also have another ticket open for checking that an instrument is present when it needs to be, which would be implemented down in `daf_butler` where we actually fully parse this expression and can do the check quite cleanly.  I hadn't though about checking for two different instruments there (since we won't _want_ to do that someday), but it might be worth doing now.

tests/test_graphBuilder.py
```diff
@@ -0,0 +1,44 @@
...
+import lsst.utils.tests
+
+
+class HelperTestCase(unittest.TestCase):
+
+    def test_getInstrument(self):
+        """Test getting the instrument from the query."""
+        queries = (("tract = 42 and instrument = 'HSC'", "HSC"),
+                   ("tract=42 and INSTRUMENT='HSC'", "HSC"),
+                   ('tract=42 and Instrument =  "HSC"', "HSC"),
```
timj, line 36: There need to be tests where instrument is first and then tract is second, and also one where instrument is in the middle of two other clauses.

python/lsst/pipe/base/graphBuilder.py
andy-slac: Even better:

    r"\Winstrument\s*=\s*[\"'](.*?)[\"']"

to avoid matching "NotAnInstrument = '1'".
I would actually suggest using expression parser that knows about query syntax to avoid surprises (ask me for more artificial examples if you want to know 🙂)

andy-slac: And in our query language string literals use single quotes, like in SQL.

python/lsst/pipe/base/graphBuilder.py
```diff
@@ -817,7 +819,79 @@ def makeGraph(self, pipeline, collections, run, userQuery):
...
+            queryInstrument = self._getInstrumentFromQuery(query)
+            if queryInstrument is None:
+                # There is not an instrument in the query, add it:
+                restriction = f"instrument = '{instrument.getName()}'"
+                _LOG.info(f"Adding restriction \"{restriction}\" to query.")
```
andy-slac, line 863: Maybe `debug`, that is not super-useful as INFO, unless you like to annoy people 🙂

andy-slac (changes requested):
Let's do the right thing and parse user expression correctly. I'll give you an example of how to do it on JIRA, hopefully today.

python/lsst/pipe/base/graphBuilder.py
timj: I'd probably use `\binstrument` for word boundary...

tests/test_graphBuilder.py
timj: and I guess also a `tract=42 and notinstrument = "HSC" and instrument = "LSSTCam"`

python/lsst/pipe/base/graphBuilder.py
```diff
@@ -745,6 +746,78 @@ def makeQuantumGraph(self):
         return graph
 
 
+from lsst.daf.butler.registry.queries.exprParser import TreeVisitor, ParserYacc, ParseError
```
andy-slac, line 749: Should go to the top

python/lsst/pipe/base/graphBuilder.py
```diff
@@ -817,7 +890,58 @@ def makeGraph(self, pipeline, collections, run, userQuery):
...
+        if not queryInstruments:
+            # There is not an instrument in the query, add it:
+            restriction = f"instrument = '{instrumentName}'"
+            _LOG.debug(f"Adding restriction \"{restriction}\" to query.")
+            query = f"{restriction} AND ({query})"
+        elif queryInstruments[0] != instrumentName:
+            # Since there is an instrument in the query, it should match
+            # the instrument in the pipeline.
+            raise RuntimeError(f"The instrument named in the query (\"{queryInstruments[0]}\") does not "
+                                f"match the instrument named by the pipeline (\"{instrumentName}\")")
```
andy-slac, line 946: Indentation.

tests/test_graphBuilder.py
```diff
@@ -0,0 +1,84 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+"""Tests of things related to the GraphBuilder class."""
+
+import unittest
+from unittest.mock import patch
```
andy-slac, line 25: flake8 says its not used

tests/test_graphBuilder.py
```diff
@@ -0,0 +1,84 @@
...
+
+from lsst.pipe.base import GraphBuilder
+from lsst.pipe.base.graphBuilder import findInstruments
```
andy-slac, line 28: This one too, and I think it does not exist.

andy-slac (changes requested):
Travis failed due to flake8 errors, and I'm not sure that test would run because it imports something that is not there anymore.

andy-slac (approved):
Looks OK.

PR #145, opened 2020-09-21
DM-26774: Instrument-finding code incorrectly requires a data query

n8pease:


timj (approved):
This looks okay. Can we add a simple test to tests/test_graphBuilder.py with an empty query? (also, HelperTestCase is a bit too generic a unittest TestCase class name for my liking).

python/lsst/pipe/base/tests/simpleQGraph.py
```diff
@@ -153,7 +180,21 @@ def registerDatasetTypes(registry, pipeline):
                 registry.registerDatasetType(datasetType)
 
 
-def makeSimpleQGraph(nQuanta=5, pipeline=None, butler=None, root=None, skipExisting=False, inMemory=True):
+def makeSimplePipeline(nQuanta, instrument=None):
```
andy-slac, line 183: Docstring would be nice.

python/lsst/pipe/base/tests/simpleQGraph.py
```diff
@@ -153,7 +180,21 @@ def registerDatasetTypes(registry, pipeline):
...
+        pipeline.addTask(AddTask, f"task{lvl}")
+        pipeline.addConfigOverride(f"task{lvl}", "connections.in_tmpl", f"{lvl}")
+        pipeline.addConfigOverride(f"task{lvl}", "connections.out_tmpl", f"{lvl+1}")
+    if instrument:
+        pipeline.addInstrument(instrument)
+    return pipeline
+
+
+def makeSimpleQGraph(nQuanta=5, pipeline=None, butler=None, root=None, skipExisting=False, inMemory=True,
+                     userQuery=""):
```
andy-slac, line 197: Can you add docstring for `userQuery`?

tests/test_graphBuilder.py
```diff
@@ -86,6 +92,30 @@ def testNoQueryNoInstruments(self):
         self.assertEqual(GraphBuilder._verifyInstrumentRestriction("", ""), "")
 
 
+class GraphBuilderTestCase(unittest.TestCase):
+
+    def testDefault(self):
+        """Simple test to verify makeSimpleQGraph can be used to make a Quantum
+        Graph."""
+        with tempfile.TemporaryDirectory() as root:
```
andy-slac, line 100: Tim pointed out to me that `tempfile.TemporaryDirectory` may not behaving well if there are errors during deleting the directory, it does not specify `ignore_errors=True` when it calls `shutil.rmtree`. I had to implement trivial context manager to workaround that: https://github.com/lsst/ctrl_mpexec/blob/master/tests/test_cmdLineFwk.py#L78-L87. I think you need something like that too (better of course is to have it somewhere in the common code base but I don't know where that could be). Same for the method below.

andy-slac (approved):
Looks good, but may be worth replacing `tempfile.TemporaryDirectory`, check my comments.

PR #146, opened 2020-09-21
DM-26726: freeze the config

timj:


natelust (approved):
Can you update the commit message a bit, capitalize First character and put a bit more description something like "Freeze task configs when expanding a pipeline"

python/lsst/pipe/base/pipeline.py
```diff
@@ -83,6 +83,7 @@ class TaskDef:
     """
     def __init__(self, taskName, config, taskClass=None, label=""):
         self.taskName = taskName
+        config.freeze()
```
natelust, line 86: I'm still not comfortable that this will have a side effect on the reference held in `toExapandedPipeline` (or any other place this may end up used). That said it is not unreasonable to ensure the config is frozen, and you did put a note in the docs, so I have no strong opinion.

PR #148, opened 2020-09-24
DM- 26629: switch to calibration collections instead of the calibration_label dimension

TallJimbo:


python/lsst/pipe/base/graphBuilder.py
```diff
@@ -707,22 +707,37 @@ def resolveDatasetRefs(self, registry, collections, run, commonDataIds, *, skipE
...
-                # the user expression used for the initial query.
                 for datasetType in task.prerequisites:
                     lookupFunction = lookupFunctions.get(datasetType.name)
                     if lookupFunction is not None:
+                        # PipelineTask has provided its own function to do the
+                        # lookup.  This always takes precedence.
                         refs = list(
                             lookupFunction(datasetType, registry, quantum.dataId, collections)
                         )
+                    elif (datasetType.isCalibration()
```
natelust, line 718: My only comment about this block is that I had always intended to go back and make the else a free function so it was easier to use in a custom lookupFunction that wanted to extend that behavior. Perhaps you could do that and the same for the calibration bits if you think it is worth it. Its not too much code to copy for someone, and its use will be small so if you dont think its a good use of time I am fine with you not.

python/lsst/pipe/base/graphBuilder.py
```diff
@@ -707,22 +707,37 @@ def resolveDatasetRefs(self, registry, collections, run, commonDataIds, *, skipE
...
+                            and datasetType.dimensions <= quantum.dataId.graph
+                            and quantum.dataId.graph.temporal):
+                        # This is a master calibration lookup, which we have to
+                        # handle specially because the query system can't do a
+                        # temporal join on a non-dimension-based timespan yet.
+                        timespan = quantum.dataId.timespan
+                        try:
+                            refs = [registry.findDataset(datasetType, quantum.dataId,
```
natelust, line 726: Can you be sure that this will only ever return one, is an exception raised if more than one could be found for a timespan? Does the certification process prevent this?

python/lsst/pipe/base/graphBuilder.py
```diff
@@ -707,22 +707,37 @@ def resolveDatasetRefs(self, registry, collections, run, commonDataIds, *, skipE
...
+                            refs = []
                     else:
+                        # Most general case.
                         refs = list(registry.queryDatasets(datasetType,
                                                            collections=collections,
                                                            dataId=quantum.dataId,
                                                            deduplicate=True).expanded())
-                    quantum.prerequisites[datasetType].update({ref.dataId: ref for ref in refs})
+                    quantum.prerequisites[datasetType].update({ref.dataId: ref for ref in refs
+                                                               if ref is not None})
```
natelust, line 740: If my ticket lands first, this is broken, infact... looking at this makes me feel I need to go look back at my ticket...

python/lsst/pipe/base/testUtils.py
```diff
@@ -163,6 +163,41 @@ def _refFromConnection(butler, connection, dataId, **kwargs):
...
+                                                          collections=butler.collections)
+                if resolvedRef is None:
+                    raise ValueError(
+                        f"Cannot find {ref.datasetType.name} with id {ref.dataId} "
+                        f"in collections {butler.collections}."
+                    )
+                newRefsForDatasetType.append(resolvedRef)
+            else:
+                newRefsForDatasetType.append(ref)
+        refsForDatasetType[:] = newRefsForDatasetType
```
natelust, line 198: This changes too with my ticket, the race is on, maybe I should have held back these comments? :)

natelust (approved):
I think it mostly looks fine

python/lsst/pipe/base/graphBuilder.py
natelust: Nevermind, this is a QuantumScaffolding

python/lsst/pipe/base/butlerQuantumContext.py
```diff
@@ -71,11 +71,11 @@ def __init__(self, butler: Butler, quantum: Quantum):
         def _get(self, ref):
             if isinstance(ref, DeferredDatasetRef):
                 self._checkMembership(ref.datasetRef, self.allInputs)
-                return butler.getDeferred(ref.datasetRef)
+                return butler.getDirectDeferred(ref.datasetRef)
```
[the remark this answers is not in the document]
timj: getDirectDeferred will raise an exception if the ref.id is None.

python/lsst/pipe/base/butlerQuantumContext.py
```diff
@@ -71,11 +71,11 @@ def __init__(self, butler: Butler, quantum: Quantum):
...
 
             else:
                 self._checkMembership(ref, self.allInputs)
-                return butler.get(ref)
+                return butler.getDirect(ref)
```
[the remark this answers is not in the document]
timj: getDirect already gets upset if it's not a resolved ref.

python/lsst/pipe/base/butlerQuantumContext.py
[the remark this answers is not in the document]
TallJimbo: Yes, thanks to Tim's comment on the daf_butler PR.  As we discussed out-of-band, I'm also uncomfortable with this relying on logic _in ctrl_mpexec_, but I think the problem is where that code lives, not what it does or guarantees.  But in addition to that check for not-None ID (which we can defer to butler), I'll go add some documentation to the `ButlerQuantumContext` class docs and anywhere else I can find that resolved `DatasetRefs` are a precondition.

python/lsst/pipe/base/graphBuilder.py
TallJimbo: I agree that's a good idea, but this ticket is too big and cumbersome for any more scope, and I _definitely_ plan to be revisit this block in the next month anyway.

python/lsst/pipe/base/graphBuilder.py
TallJimbo: Yes, `findDataset` will raise in that case.  The certification process is supposed to guarantee that is impossible for any calibration I can think of, but it's not something we can guarantee at the level of database constraints.  So the exception basically says, "someone probably put together a malformed calibration repo".

python/lsst/pipe/base/testUtils.py
TallJimbo: At least (I imagine) it'll be an easy change in either case.

PR #151, opened 2020-10-05
DM-27060: pipe_base documentation fails to build due to BuildId

jdswinbank:
These types cause a documentation build failure:

  FileNotFoundError: [Errno 2] No such file or directory: '/j/ws/sqre/infra/documenteer/doc_template/_build/doctree/py-api/lsst.pipe.base.BuildId.doctree'

Presumably because Sphinx/autodoc doesn't understand that they are generated
by typing.NewType rather than being fully-fledged classes.

### DIFF
--- a/doc/lsst.pipe.base/index.rst
+++ b/doc/lsst.pipe.base/index.rst
@@ -70,6 +70,8 @@ Python API reference
 
 .. automodapi:: lsst.pipe.base
    :no-main-docstr:
+   :skip: BuildId
+   :skip: DatasetTypeName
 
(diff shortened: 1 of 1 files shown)


TallJimbo (approved):
Telling Sphinx to ignore these seems like the best fix for now.

But if these types are causing problems because they're being used to document (e.g.) function arguments, I might ask @natelust to open a ticket to change that, too, because having docs reference a pseudo-type that can only be understood by looking at the code doesn't seem like good practice.  Using `NewType` in type annotations is okay; using it in docstrings apparently isn't right now.

Thanks for tracking this down!

PR #153, opened 2020-10-09
DM-27131: Reorganize pickling to allow subclasses to add parameters

timj:
Use a factory unpickler in conjunction with a new method that
forms the keyword arguments to use in the constructor. This
allows subclasses to add additional parameters without requiring
they all implement a compatible reduce method.

### DIFF
--- a/python/lsst/pipe/base/task.py
+++ b/python/lsst/pipe/base/task.py
@@ -370,7 +370,30 @@ class Task:
         """
         return f"{self._fullName}.{name}"
 
+    @staticmethod
+    def _unpickle_via_factory(factory, args, kwargs):
+        """Unpickle something by calling a factory
+
+        Allows subclasses to unpickle using `__reduce__` with keyword
+        arguments as well as positional arguments.
+        """
+        return factory(*args, **kwargs)
+
+    def _reduce_kwargs(self):
+        """Returns a dict of the keyword arguments that should be used
+        by `__reduce__`.
+
+        Subclasses with additional arguments should always call the parent
+        class method to ensure that the standard parameters are included.
+
+        Returns
+        -------
+        kwargs : `dict`
+            Keyword arguments to be used when pickling.
+        """
+        return dict(config=self.config, name=self._name, parentTask=self._parentTask,)
+
     def __reduce__(self):
         """Pickler.
         """
-        return self.__class__, (self.config, self._name, self._parentTask, None)
+        return self._unpickle_via_factory, (self.__class__, [], self._reduce_kwargs())
--- a/python/lsst/pipe/base/task.py
+++ b/python/lsst/pipe/base/task.py
@@ -37,7 +37,8 @@ class TaskError(Exception):
     -----
     Examples of such errors:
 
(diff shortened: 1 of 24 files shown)


python/lsst/pipe/base/task.py
```diff
@@ -370,7 +370,30 @@ def _computeFullName(self, name):
...
+        kwargs : `dict`
+            Keyword arguments to be used when pickling.
+        """
+        return dict(config=self.config, name=self._name, parentTask=self._parentTask,)
+
     def __reduce__(self):
         """Pickler.
         """
-        return self.__class__, (self.config, self._name, self._parentTask, None)
+        return self._unpickle_via_factory, (self.__class__, [], self._reduce_kwargs())
```
natelust, line 399: I think you might want to look at `__getnewargs_ex__` instead of `__reduce__`.

timj: I continually try to look at it and then people always tell me they prefer reduce instead.

TallJimbo: I've been switching the classes where I had used `__getnewargs_ex__` to use `__reduce__` with this kind of trampoline, because I don't think it's worth it to implement `__new__` and explicitly disable `__getstate__` and `__setstate__` just to be able to use `__getnewargs_ex__`.  I'm not sure `__getstate__` and `__setstate__` are always in play, but I really don't like having to think about when they might be, and how many other invisible construction paths I have to guard against.  `__reduce__` (with or without a trampoline) basically guarantees `__init__` is called, and I think that simplicity is really underrated in the pickle docs' recommendations to prefer other approaches.

natelust: @TallJimbo: I see the difference here from where I have used this in the past. You want to explicitly want to re-trigger the behavior in __init__ rather than just save the state of something and restore that state (which is what you get with it just calling `object.__new__` (or your own `__new__`), `__init__` is never called a new object is just created with the specified state. FWIW, you dont need get and set state with with `__getnewargs_ex__`, its (yet) another way to do pickling.

timj: Yes, in this case the `__init__` is doing a lot of mangling of the input arguments and I really don't want to be trying to keep track of all the internal attributes that are set.

python/lsst/pipe/base/task.py
```diff
@@ -370,7 +370,30 @@ def _computeFullName(self, name):
...
+
+        Subclasses with additional arguments should always call the parent
+        class method to ensure that the standard parameters are included.
+
+        Returns
+        -------
+        kwargs : `dict`
+            Keyword arguments to be used when pickling.
+        """
+        return dict(config=self.config, name=self._name, parentTask=self._parentTask,)
```
ktlim, line 430: Doesn't `PipelineTask` need a `_reduce_kwargs()` override that includes `initInputs`?

ktlim (approved):
I'd like at least a ticket to scrub all other Tasks to make sure that they don't need a `_reduce_kwargs()` override.

PR #154, opened 2020-10-12
DM-27140 Add named subsets to pipelines

natelust:
Add a directive to pipeline documents that allows a subset of
labels in the pipeline to be declared with a name. This
modification allows portions of a pipeline to be run with using
this named subset instead of specifying a list of labels.

### DIFF
--- a/python/lsst/pipe/base/pipeline.py
+++ b/python/lsst/pipe/base/pipeline.py
@@ -34,7 +34,6 @@ from typing import Mapping, Set, Union, Generator, TYPE_CHECKING, Optional
 
 import copy
 import os
-import re
 
 # -----------------------------
 #  Imports for other modules --
@@ -223,52 +222,22 @@ class Pipeline:
         string based matching due to the nature of contracts and may prune more
         than it should.
         """
-
-        pipeline = copy.deepcopy(self)
-
-        def remove_contracts(label: str):
-            """Remove any contracts that contain the given label
-
-            String comparison used in this way is not the most elegant and may
-            have issues, but it is the only feasible way when users can specify
-            contracts with generic strings.
-            """
-            new_contracts = []
-            for contract in pipeline._pipelineIR.contracts:
-                # match a label that is not preceded by an ASCII identifier, or
-                # is the start of a line and is followed by a dot
-                if re.match(f".*([^A-Za-z0-9_]|^){label}[.]", contract.contract):
-                    continue
-                new_contracts.append(contract)
-            pipeline._pipelineIR.contracts = new_contracts
-
-        # Labels supplied as a set, explicitly remove any that are not in
-        # That list
+        # Labels supplied as a set
         if labelSpecifier.labels:
-            # verify all the labels are in the pipeline
-            if not labelSpecifier.labels.issubset(pipeline._pipelineIR.tasks.keys()):
-                difference = labelSpecifier.labels.difference(pipeline._pipelineIR.tasks.keys())
-                raise ValueError("Not all supplied labels are in the pipeline definition, extra labels:"
-                                 f"{difference}")
-            # copy needed so as to not modify while iterating
-            pipeline_labels = list(pipeline._pipelineIR.tasks.keys())
-            for label in pipeline_labels:
-                if label not in labelSpecifier.labels:
-                    pipeline.removeTask(label)
-                    remove_contracts(label)
+            labelSet = labelSpecifier.labels
         # Labels supplied as a range, first create a list of all the labels
         # in the pipeline sorted according to task dependency. Then only
         # keep labels that lie between the supplied bounds
         else:
-            # Use a dict for fast searching while preserving order
-            # save contracts and remove them so they do not fail in the
-            # expansion step, will be restored after. This is needed because
-            # a user may only configure the tasks they intend to run, which
-            # may cause some contracts to fail if they will later be dropped
-            contractSave = pipeline._pipelineIR.contracts
+            # Create a copy of the pipeline to use when assessing the label
+            # ordering. Use a dict for fast searching while preserving order.
+            # Remove contracts so they do not fail in the expansion step. This
+            # is needed because a user may only configure the tasks they intend
+            # to run, which may cause some contracts to fail if they will later
+            # be dropped
+            pipeline = copy.deepcopy(self)
             pipeline._pipelineIR.contracts = []
             labels = {taskdef.label: True for taskdef in pipeline.toExpandedPipeline()}
-            pipeline._pipelineIR.contracts = contractSave
 
             # Verify the bounds are in the labels
             if labelSpecifier.begin is not None:
@@ -280,28 +249,17 @@ class Pipeline:
                     raise ValueError(f"End of range subset, {labelSpecifier.end}, not found in pipeline "
                                      "definition")
 
-            closed = False
+            labelSet = set()
             for label in labels:
-                # if there is a begin label delete all labels until it is
-                # reached.
-                if labelSpecifier.begin:
+                if labelSpecifier.begin is not None:
                     if label != labelSpecifier.begin:
-                        pipeline.removeTask(label)
-                        remove_contracts(label)
                         continue
                     else:
                         labelSpecifier.begin = None
-                # if there is an end specifier, keep all tasks until the
-                # end specifier is reached, afterwards delete the labels
-                if labelSpecifier.end:
-                    if label != labelSpecifier.end:
-                        if closed:
-                            pipeline.removeTask(label)
-                            remove_contracts(label)
-                        continue
-                    else:
-                        closed = True
-        return pipeline
+                labelSet.add(label)
+                if labelSpecifier.end is not None and label == labelSpecifier.end:
+                    break
+        return Pipeline.fromIR(self._pipelineIR.subset_from_labels(labelSet))
 
     @staticmethod
     def _parseFileSpecifier(fileSpecifer):
--- a/python/lsst/pipe/base/pipelineIR.py
+++ b/python/lsst/pipe/base/pipelineIR.py
@@ -1,4 +1,3 @@
-__all__ = ("ConfigIR", "ContractError", "ContractIR", "InheritIR", "PipelineIR", "TaskIR")
 # This file is part of pipe_base.
 #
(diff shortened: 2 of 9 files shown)


python/lsst/pipe/base/pipeline.py
```diff
@@ -245,11 +245,21 @@ def remove_contracts(label: str):
         # Labels supplied as a set, explicitly remove any that are not in
         # That list
         if labelSpecifier.labels:
+            # update the label specifier to expand any named subsets
+            toRemove = set()
+            toAdd = set()
+            for label in labelSpecifier.labels:
+                if label in pipeline._pipelineIR.named_subsets:
+                    toRemove.add(label)
+                    toAdd.update(pipeline._pipelineIR.named_subsets[label].subset)
```
TallJimbo, line 254: Have you thought about whether this should be recursive (i.e. can subsets nest)?  I don't see a compelling reason why it should be, but it's probably something that should be a conscious choice rather than an implementation artifact.

TallJimbo (commented):
Looks good from a preliminary perspective; don't have much to say structurally.  I haven't done a close read for e.g. docstrings etc. yet.

python/lsst/pipe/base/pipeline.py
```diff
@@ -223,52 +222,22 @@ def subsetFromLabels(self, labelSpecifier: LabelSpecifier) -> Pipeline:
...
-            # a user may only configure the tasks they intend to run, which
-            # may cause some contracts to fail if they will later be dropped
-            contractSave = pipeline._pipelineIR.contracts
+            # Create a copy of the pipeline to use when assessing the label
+            # ordering. Use a dict for fast searching while preserving order.
+            # Remove contracts so they do not fail in the expansion step. This
+            # is needed because a user may only configure the tasks they intend
+            # to run, which may cause some contracts to fail if they will later
+            # be dropped
+            pipeline = copy.copy(self)
```
TallJimbo, line 238: Are you sure a one-level deep shallow copy is the right level of copying here?

(I always find `copy.copy` to be scary, and I might go so far as to call it an antipattern, because whether it is or is not the right amount of copying to do is really an implementation detail in general.)

python/lsst/pipe/base/pipeline.py
```diff
@@ -280,28 +249,17 @@ def remove_contracts(label: str):
...
-                    if label != labelSpecifier.end:
-                        if closed:
-                            pipeline.removeTask(label)
-                            remove_contracts(label)
-                        continue
-                    else:
-                        closed = True
-        return pipeline
+                labelSet.add(label)
+                if labelSpecifier.end and label == labelSpecifier.end:
```
TallJimbo, line 260: If the first test here could be `labelSpecifier.end is not None`, it probably should be.

python/lsst/pipe/base/pipelineIR.py
```diff
@@ -87,6 +92,66 @@ def __eq__(self, other: "ContractIR"):
...
+    a pipeline yaml file.
+    """
+    label: str
+    """The label used to identify the subset of task labels.
+    """
+    subset: Set[str]
+    """A set of task labels contained in this subset.
+    """
+    description: Optional[str]
+    """A description of what this sub set of tasks is intended to do
```
TallJimbo, line 107: "sub set" -> "subset"

python/lsst/pipe/base/pipelineIR.py
```diff
@@ -87,6 +92,66 @@ def __eq__(self, other: "ContractIR"):
...
+    """
+
+    @staticmethod
+    def from_yaml(label: str, value: Union[List[str], dict]) -> LabeledSubset:
+        """This static method generates `LabeledSubset` objects given a properly
+        formatted object that as been created by a yaml loader.
```
TallJimbo, line 113: Docstring should [start with the imperative voice](https://developer.lsst.io/python/numpydoc.html#writing-summaries-for-functions-and-methods).

python/lsst/pipe/base/pipelineIR.py
```diff
@@ -87,6 +92,66 @@ def __eq__(self, other: "ContractIR"):
...
+
+        Parameters
+        ----------
+        label : `str`
+            The label that will be used to identify this labeled subset.
+        value : `list` of str or `dict`
```
TallJimbo, line 119: Add backticks around `str`.

python/lsst/pipe/base/pipelineIR.py
```diff
@@ -87,6 +92,66 @@ def __eq__(self, other: "ContractIR"):
...
+            Object returned from loading a labeled subset section from a yaml
+            document.
+
+        Returns
+        -------
+        labeledSubset : `LabeledSubset`
+            A `LabeledSubset object build from the inputs.
```
TallJimbo, line 126: Need closing backtick.  Please run `package-docs build`, too, as that may catch something I've missed.

python/lsst/pipe/base/pipelineIR.py
```diff
@@ -87,6 +92,66 @@ def __eq__(self, other: "ContractIR"):
...
+
+        Raises
+        ------
+        ValueError
+            Raised if the value input is not properly formatted for parsing
+        """
+        if isinstance(value, Mapping):
+            subset = value.pop("subset", None)
```
TallJimbo, line 134: Can't `pop` a `Mapping`.  Maybe you mean `MutableMapping`?

python/lsst/pipe/base/pipelineIR.py
```diff
@@ -87,6 +92,66 @@ def __eq__(self, other: "ContractIR"):
...
+            if subset is None:
+                msg = "If a labeled subset is specified as a mapping, it must contain the key 'subset'"
+                raise ValueError(msg)
```
TallJimbo, line 137: Why define `msg` as a local variable here?

python/lsst/pipe/base/pipelineIR.py
```diff
@@ -87,6 +92,66 @@ def __eq__(self, other: "ContractIR"):
...
+            description = value.pop("description", None)
+        elif isinstance(value, abcIterable):
+            subset = value
+            description = None
+        else:
+            raise ValueError(f"There was a problem parsing the labeled subset {label}")
```
TallJimbo, line 143: This is not a very helpful error message.  Can you at least report what syntax was expected?

python/lsst/pipe/base/pipelineIR.py
```diff
@@ -345,6 +418,36 @@ def _read_contracts(self, loaded_yaml):
             if isinstance(contract, str):
                 self.contracts.append(ContractIR(contract=contract))
 
+    def _read_labeled_subsets(self, loaded_yaml: dict):
+        """Process the subsets portion of the loaded yaml document
+
+        Parameters
+        ----------
+        loaded_yaml: `Mapping`
```
TallJimbo, line 426: As above, I think this needs to be `MutableMapping` (and you should probably state below that it's modified in-place).

python/lsst/pipe/base/pipelineIR.py
```diff
@@ -87,6 +92,66 @@ def __eq__(self, other: "ContractIR"):
...
+                raise ValueError(msg)
+            description = value.pop("description", None)
+        elif isinstance(value, abcIterable):
+            subset = value
+            description = None
+        else:
+            raise ValueError(f"There was a problem parsing the labeled subset {label}")
+        return LabeledSubset(label, set(subset), description)
+
+    def to_primitives(self) -> dict:
```
TallJimbo, line 146: Why is the "to" method `to_primitives`, but the "from" method `from_yaml`?  It looks like "primitives" and "yaml" mean the same thing, given that `from_yaml` operates on a `dict` or `list` rather than a file or string.

python/lsst/pipe/base/pipelineIR.py
```diff
@@ -373,13 +476,31 @@ def process_args(argument: Union[str, dict]) -> dict:
...
             if accumulate_tasks.keys() & tmp_IR.tasks.keys():
                 raise ValueError("Task labels in the imported pipelines must "
                                  "be unique")
             accumulate_tasks.update(tmp_IR.tasks)
             self.contracts.extend(tmp_IR.contracts)
+            # verify that tmp_IR has unique labels for named subset among
+            # existing labeled subsets, and with existing task labels.
+            overlapping_subsets = accumulate_labeled_subsets.keys() & tmp_IR.labeled_subsets.keys()
+            task_subset_overlap = (accumulate_labeled_subsets.keys() | tmp_IR.labeled_subsets.keys()) \
+                & accumulate_tasks.keys()
```
TallJimbo, line 491: An extra parentheses for line continuation [is preferred](https://developer.lsst.io/python/style.html#python-s-implied-continuation-inside-parens-brackets-and-braces-should-be-used-for-wrapped-lines) over backslash.

python/lsst/pipe/base/pipelineIR.py
```diff
@@ -430,6 +551,84 @@ def _read_tasks(self, loaded_yaml):
...
+        # those labels (see docstring on _remove_contracts for why this may
+        # cause issues)
+        for label in pipeline_labels:
+            if label not in labelSpecifier:
+                pipeline.tasks.pop(label)
+                pipeline._remove_contracts(label)
+
+        # create a copy of the object to iterate over
+        labeled_subsets = copy.copy(pipeline.labeled_subsets)
+        # remove any labeled subsets that no longer have a complete set
```
TallJimbo, line 625: This sounds like the right behavior, but it's sufficiently non-obvious that it should probably be mentioned in the docstring as well.

TallJimbo (approved):
Looks good, but lots of little style nitpicks.

python/lsst/pipe/base/pipelineIR.py
natelust: I personally prefer this to breaks inside the ValueError constructor, but I switched to that style

python/lsst/pipe/base/pipeline.py
natelust: Yeah, I had deep copy then switched to copy because I knew the implementation details (there are only a few top level containers of things that nothing really modifies) but it probably is better to always deep copy to avoid any future issues that may arise, it certainly is worth the small amount of memory overhead.

PR #159, opened 2020-11-17
DM-27633 add a parameters section to Pipelines

natelust:


python/lsst/pipe/base/pipelineIR.py
```diff
@@ -195,6 +240,27 @@ def to_primitives(self) -> dict:
...
+        Returns
+        -------
+        config : ConfigIR
+            A new ConfigIR object formatted with the input parameters
+        """
+        new_config = copy.deepcopy(self)
+        for key, value in new_config.rest.items():
+            match = re.match("parameters[.](.*)", value)
+            if match and match.group(1) in parameters:
+                new_config.rest[key] = parameters[match.group(1)]
```
czwa, line 261: Should this warn or raise `if match and match.group(1) not in parameters`?  Or will that fail later in the pipeline construction process?

python/lsst/pipe/base/pipelineIR.py
```diff
@@ -320,14 +386,28 @@ class InheritIR:
...
+    def toPipelineIR(self, instrument=None) -> "PipelineIR":
         """Convert to a representation used in yaml serialization
+        Load in the Pipeline specified by this object, and turn it into a
+        PipelineIR instance.
+
+        Parameters
+        ----------
+        instrument : Optional `str`
+            A string giving the fully qualified path to an instrument object.
+            If a inherited pipeline defines the same instrument as defined in
```
czwa, line 398: Typo: "an inherited".

czwa (approved):
Other than the question and typo, it looks fine to me.

python/lsst/pipe/base/pipelineIR.py
natelust: I left this open so that if there is a parameter in the pipeline and it matches it is used. But if someone had a value that contained this string that was not a parameter it would allow it to continue in case there was a task somewhere that needed this as a string or something for some other reason. I didn't want to be any more restrictive than I needed to be. Do you think it should be some other way (I definitely am open to other viewpoints here, I just went with what seemed best to me)

czwa: I went back and forth, and I think I've settled on "a warning is good."  In the case where a parameter wasn't defined right, it can at least say "No parameter named 'typoe' found".  In the case where someone defines a `goodIdea: config.parameters.badIdea`, the warning will hopefully annoy them into not using that name.

PR #161, opened 2020-11-20
DM-27682: Add ability to save/load QuantumGraph with URI

timj:


python/lsst/pipe/base/graph/graph.py
```diff
@@ -549,8 +549,64 @@ def findCycle(self) -> List[Tuple[QuantumNode, QuantumNode]]:
...
+        initialization. To make sure that DimensionUniverse exists this method
+        accepts dummy DimensionUniverse argument.
+        """
+        uri = ButlerURI(uri)
+        # With ButlerURI we have the choice of always using a local file
+        # or reading in the bytes directly. Reading in bytes can be more
+        # efficient for reasonably-sized pickle files when the resource
+        # is remote. For now use the local file variant. For a local file
+        # as_local() does nothing.
+        with uri.as_local() as local:
```
natelust, line 600: I think this would be nicer written on one line as:
```python
with uri.as_local() as local, open(local.ospath, 'rb') as fd:
    qgraph = pickle.load
```

natelust (approved):

PR #162, opened 2020-11-23
Add Graph property for accessing build Id (DM-27667)

andy-slac:
For verifying graph build ID in pipetask we need a public access to
graph's build ID.


andy-slac:
Closing this, Nate implemented better solution on DM-27784.

PR #163, opened 2020-12-01
DM-27784 Introduce new save/load format for QuantumGraph

natelust:
This new on disk format, as well as new loading methods allow
partial loading of a QuantumGraph given a set of node number.

### DIFF
--- a/python/lsst/pipe/base/graph/_loadHelpers.py
+++ b/python/lsst/pipe/base/graph/_loadHelpers.py
@@ -1,0 +1,367 @@
+# This file is part of pipe_base.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (http://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+from __future__ import annotations
+
+__all__ = ("LoadHelper")
+
+from lsst.daf.butler import ButlerURI, Quantum
+from lsst.daf.butler.core._butlerUri.s3 import ButlerS3URI
+from lsst.daf.butler.core._butlerUri.file import ButlerFileURI
+
+from ..pipeline import TaskDef
+from .quantumNode import NodeId
+
+from dataclasses import dataclass
+import functools
+import io
+import lzma
+import pickle
+import struct
+
+from collections import defaultdict, UserDict
+from typing import (Optional, Iterable, DefaultDict, Set, Dict, TYPE_CHECKING, Type, Union)
+
+if TYPE_CHECKING:
+    from . import QuantumGraph
+
+
+# Create a custom dict that will return the desired default if a key is missing
+class RegistryDict(UserDict):
+    def __missing__(self, key):
+        return DefaultLoadHelper
+
+
+# Create a registry to hold all the load Helper classes
+HELPER_REGISTRY = RegistryDict()
+
+
+def register_helper(URICLass: Union[Type[ButlerURI], Type[io.IO[bytes]]]):
+    """Used to register classes as Load helpers
+
+    When decorating a class the parameter is the class of "handle type", i.e.
+    a ButlerURI type or open file handle that will be used to do the loading.
+    This is then associated with the decorated class such that when the
+    parameter type is used to load data, the appropriate helper to work with
+    that data type can be returned.
+
+    A decorator is used so that in theory someone could define another handler
+    in a different module and register it for use.
+
+    Parameters
+    ----------
+    URIClass : Type of `~lsst.daf.butler.ButlerURI` or `~io.IO` of bytes
+        type for which the decorated class should be mapped to
+    """
+    def wrapper(class_):
+        HELPER_REGISTRY[URICLass] = class_
+        return class_
+    return wrapper
+
+
+class DefaultLoadHelper:
+    """Default load helper for `QuantumGraph` save files
+
+    This class, and its subclasses, are used to unpack a quantum graph save
+    file. This file is a binary representation of the graph in a format that
+    allows individual nodes to be loaded without needing to load the entire
+    file.
+
+    This default implementation has the interface to load select nodes
+    from disk, but actually always loads the entire save file and simply
+    returns what nodes (or all) are requested. This is intended to serve for
+    all cases where there is a read method on the input parameter, but it is
+    unknown how to read select bytes of the stream. It is the responsibility of
+    sub classes to implement the method responsible for loading individual
+    bytes from the stream.
+
+    Parameters
+    ----------
+    uriObject : `~lsst.daf.butler.ButlerURI` or `io.IO` of bytes
+        This is the object that will be used to retrieve the raw bytes of the
+        save.
+
+    Raises
+    ------
+    ValueError
+        Raised if the specified file contains the wrong file signature and is
+        not a `QuantumGraph` save
+    """
+    def __init__(self, uriObject: Union[ButlerURI, io.IO[bytes]]):
+        self.uriObject = uriObject
+
+        preambleSize, taskDefSize, nodeSize = self._readSizes()
+
+        # Recode the total header size
+        self.headerSize = preambleSize + taskDefSize + nodeSize
+
+        self._readByteMappings(preambleSize, self.headerSize, taskDefSize)
+
+    def _readSizes(self):
+        # need to import here to avoid cyclic imports
+        from .graph import STRUCT_FMT_STRING, MAGIC_BYTES
+        # Read the first few bytes which correspond to the lengths of the
+        # magic identifier bytes, 2 byte version
+        # number and the two 8 bytes numbers that are the sizes of the byte
+        # maps
+        magicSize = len(MAGIC_BYTES)
+        fmt = STRUCT_FMT_STRING
+        fmtSize = struct.calcsize(fmt)
+        preambleSize = magicSize + fmtSize
+
+        headerBytes = self._readBytes(0, preambleSize)
+        magic = headerBytes[:magicSize]
+        sizeBytes = headerBytes[magicSize:]
+
+        if magic != MAGIC_BYTES:
+            raise ValueError("This file does not appear to be a quantum graph save got magic bytes "
+                             f"{magic}, expected {MAGIC_BYTES}")
+
+        # Turn they encode bytes back into a python int object
+        save_version, taskDefSize, nodeSize = struct.unpack('>HQQ', sizeBytes)
+
+        # Store the save version, so future read codes can make use of any
+        # format changes to the save protocol
+        self.save_version = save_version
+
+        return preambleSize, taskDefSize, nodeSize
+
+    def _readByteMappings(self, preambleSize, headerSize, taskDefSize):
+        # Take the header size explicitly so subclasses can modify before
+        # This task is called
+
+        # read the bytes of taskDef bytes and nodes skipping the size bytes
+        headerMaps = self._readBytes(preambleSize, headerSize)
+
+        # read the map of taskDef bytes back in skipping the size bytes
+        self.taskDefMap = pickle.loads(headerMaps[:taskDefSize])
+
+        # read back in the graph id
+        self._buildId = self.taskDefMap['__GraphBuildID']
+
+        # read the map of the node objects back in skipping bytes
+        # corresponding to the taskDef byte map
+        self.map = pickle.loads(headerMaps[taskDefSize:])
+
+    def load(self, nodes: Optional[Iterable[int]] = None, graphID: Optional[str] = None) -> QuantumGraph:
+        """Loads in the specified nodes from the graph
+
+        Load in the `QuantumGraph` containing only the nodes specified in the
+        ``nodes`` parameter from the graph specified at object creation. If
+        ``nodes`` is None (the default) the whole graph is loaded.
+
+        Parameters
+        ----------
+        nodes : `Iterable` of `int` or `None`
+            The nodes to load from the graph, loads all if value is None
+            (the default)
+        graphID : `str` or `None`
+            If specified this ID is verified against the loaded graph prior to
+            loading any Nodes. This defaults to None in which case no
+            validation is done.
+
+        Returns
+        -------
+        graph : `QuantumGraph`
+            The loaded `QuantumGraph` object
+
+        Raises
+        ------
+        ValueError
+            Raised if one or more of the nodes requested is not in the
+            `QuantumGraph` or if graphID parameter does not match the graph
+            being loaded.
+        """
+        # need to import here to avoid cyclic imports
+        from . import QuantumGraph
+        if graphID is not None and self._buildId != graphID:
+            raise ValueError('graphID does not match that of the graph being loaded')
+        # Read in specified nodes, or all the nodes
+        if nodes is None:
+            nodes = list(self.map.keys())
+            # if all nodes are to be read, force the reader from the base class
+            # that will read all they bytes in one go
+            _readBytes = functools.partial(DefaultLoadHelper._readBytes, self)
+        else:
+            # only some bytes are being read using the reader specialized for
+            # this class
+            # create a set to ensure nodes are only loaded once
+            nodes = set(nodes)
+            # verify that all nodes requested are in the graph
+            remainder = nodes - self.map.keys()
+            if remainder:
+                raise ValueError("Nodes {remainder} were requested, but could not be found in the input "
+                                 "graph")
+            _readBytes = self._readBytes
+        # create a container for loaded data
+        quanta: DefaultDict[TaskDef, Set[Quantum]] = defaultdict(set)
+        quantumToNodeId: Dict[Quantum, NodeId] = {}
+        loadedTaskDef = {}
+        # loop over the nodes specified above
+        for node in nodes:
+            # Get the bytes to read from the map
+            start, stop = self.map[node]
+            start += self.headerSize
+            stop += self.headerSize
+
+            # read the specified bytes, will be overloaded by subclasses
+            # bytes are compressed, so decompress them
+            dump = lzma.decompress(_readBytes(start, stop))
+
+            # reconstruct node
+            qNode = pickle.loads(dump)
+
+            # read the saved node, name. If it has been loaded, attach it, if
+            # not read in the taskDef first, and then load it
+            nodeTask = qNode.taskDef
+            if nodeTask not in loadedTaskDef:
+                # Get the byte ranges corresponding to this taskDef
+                start, stop = self.taskDefMap[nodeTask]
+                start += self.headerSize
+                stop += self.headerSize
+
+                # load the taskDef, this method call will be overloaded by
+                # subclasses.
+                # bytes are compressed, so decompress them
+                taskDef = pickle.loads(lzma.decompress(_readBytes(start, stop)))
+                loadedTaskDef[nodeTask] = taskDef
+            # Explicitly overload the "frozen-ness" of nodes to attach the
+            # taskDef back into the un-persisted node
+            object.__setattr__(qNode, 'taskDef', loadedTaskDef[nodeTask])
+            quanta[qNode.taskDef].add(qNode.quantum)
+
+            # record the node for later processing
+            quantumToNodeId[qNode.quantum] = qNode.nodeId
+
+        # construct an empty new QuantumGraph object, and run the associated
+        # creation method with the un-persisted data
+        qGraph = object.__new__(QuantumGraph)
+        qGraph._buildGraphs(quanta, _quantumToNodeId=quantumToNodeId, _buildId=self._buildId)
+        return qGraph
+
+    def _readBytes(self, start: int, stop: int) -> bytes:
+        """Loads the specified byte range from the ButlerURI object
+
+        In the base class, this actually will read all the bytes into a buffer
+        from the specified ButlerURI object. Then for each method call will
+        return the requested byte range. This is the most flexible
+        implementation, as no special read is required. This will not give a
+        speed up with any sub graph reads though.
+        """
+        if not hasattr(self, 'buffer'):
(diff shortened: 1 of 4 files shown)


python/lsst/pipe/base/graph/_loadHelpers.py
```diff
@@ -0,0 +1,310 @@
...
+        return self.buffer[start:stop]
+
+    def close(self):
+        """Cleans up an instance if needed. Base class does nothing
+        """
+        pass
+
+
+@register_helper(ButlerS3URI)
+class S3LoadHelper(DefaultLoadHelper):
```
timj, line 213: Not reviewing, but I'd rather we added byte range support to `ButlerURI.read()` directly -- it seems generically useful and if a URI scheme can't support it I guess we have to decided whether to complain or download the full file and do local read.  WebDAV is missing from this so I'm not sure if that means you didn't look at it or it can't support it.

natelust: I just have not done webdav yet, I was going to tackle it in another follow-up ticket so as to unblock the ticket andy is working on.

I agree that if possible it would be nice to get this back upstream, as a comment in the file says. However, it will fundamentally change ButlerURI, and I was unprepared to do that with you away, or without proper planning. So I wrote it is such a way that future updates to ButlerURI would not affect the public api.

Basically ButlerURI supports the api of a file handle, but doesn't keep internal state on the position of the stream (for instance no seek method) so it does not behave as one. This is evident in the File uri where a open call happens for every read, and always assume reading bytes happens from the start of the file. Changing this changes the expectation of how the objects can be used (and or potentially means keeping handles open and forcing ButlerURI to be used as a context manager (which I know it is in some ways already))

Im happy if we want to make those changes though.

timj: Ok, I misunderstood what you had done. I was more than happy for `ButlerURI.read()` to take an offset parameter. I'd need to think more about making it something that looks like an open filehandle that remembers seek position.

natelust: I am happy to write something related to expected behavior, trade offs etc.

And maybe you are right that an offset parameter is fine, but it will make a potentially confusing interface that is close to the io module but behaves differently. But it's something we should think through.

I also should have said above, I didn't specialize it for webdav, but it will work fine. It just won't have the benefit of loading select chunks, it will use the default loader and basically load the whole thing and subset, which is not exactly ideal from a network perspective. I don't think it will be hard, and maybe I will have time to do it while andy is reviewing, if not I'll file a follow up ticket.

python/lsst/pipe/base/graph/_loadHelpers.py
```diff
@@ -0,0 +1,310 @@
...
+        """Loads the specified byte range from the ButlerURI object
+
+        In the base class, this actually will read all the bytes into a buffer
+        from the specified ButlerURI object. Then for each method call will
+        return the requested byte range. This is the most flexible
+        implementation, as no special read is required. This will not give a
+        speed up with any sub graph reads though.
+        """
+        if not hasattr(self, 'buffer'):
+            self.buffer = self.uriObject.read()
```
andy-slac, line 203: I"m not sure `hasattr` is justified in this case, could you instead set `self.buffer = None` in constructor and just check for `is None`?

python/lsst/pipe/base/graph/_loadHelpers.py
```diff
@@ -0,0 +1,310 @@
...
+        # minus 1 in the stop range, because this header is inclusive rather
+        # than standard python where the end point is generally exclusive
+        args["Range"] = f"bytes={start}-{stop-1}"
+        try:
+            response = self.uriObject.client.get_object(Bucket=self.uriObject.netloc,
+                                                        Key=self.uriObject.relativeToPathRoot,
+                                                        **args)
+        except (self.uriObject.client.exceptions.NoSuchKey,
+                self.uriObject.client.exceptions.NoSuchBucket) as err:
+            raise FileNotFoundError(f"No such resource: {self}") from err
```
andy-slac, line 226: Do you want `{self.uriObject}` instead of `{self}`?

python/lsst/pipe/base/graph/_loadHelpers.py
```diff
@@ -0,0 +1,310 @@
...
+        response["Body"].close()
+        return body
+
+
+@register_helper(ButlerFileURI)
+class FileLoadHelper(DefaultLoadHelper):
+    # This subclass implements partial loading of a graph using a file uri
+    def _readBytes(self, start: int, stop: int) -> bytes:
+        if not hasattr(self, 'fileHandle'):
+            self.fileHandle = open(self.uriObject.ospath, 'rb')
```
andy-slac, line 237: Maybe also set fileHandle to None in constructor?

python/lsst/pipe/base/graph/_loadHelpers.py
```diff
@@ -0,0 +1,310 @@
...
+            self.fileHandle.close()
+
+
+@register_helper(io.IOBase)  # type: ignore
+class OpenFileHandleHelper(DefaultLoadHelper):
+    # This handler is special in that it does not get initialized with a
+    # ButlerURI, but an open file handle.
+
+    # Most everything stays the same, the variable is even stored as uriObject,
+    # because the interface needed can stay the same. Unfortunately because we
```
andy-slac, line 252: needed can stay the same?

python/lsst/pipe/base/graph/_loadHelpers.py
```diff
@@ -0,0 +1,310 @@
...
+
+        # Store the save version, so future read codes can make use of any
+        # format changes to the save protocol
+        self.save_version = save_version
+
+        # Recode the total header size
+        self.headerSize = 18+taskDefSize+nodeSize
+
+        # read the bytes of taskDef bytes and nodes skipping the size bytes
+        headerMaps = self.uriObject.read(self.headerSize)[18:]
```
andy-slac, line 120: Why not `_readBytes()` here (and above)?

python/lsst/pipe/base/graph/_loadHelpers.py
```diff
@@ -0,0 +1,310 @@
...
+    def __init__(self, uriObject: io.IO[bytes]):
+        # This is actually a io handle, but is named uri because the follow the
+        # same api, and this can be used in place of a URI object in this
+        # class.
+        self.uriObject = uriObject
+
+        sizeBytes = self.uriObject.read(18)
+        save_version, taskDefSize, nodeSize = struct.unpack('>HQQ', sizeBytes)
+        self.save_version = save_version
+        self.headerSize = 18+taskDefSize+nodeSize
```
andy-slac, line 266: This code duplication and that you don't call base class constructor suggests that some refactoring may be in order.

andy-slac (approved):
Looks OK in general, bunch of minor comments.

tests/test_quantumGraph.py
```diff
@@ -334,30 +345,56 @@ def testFindCycle(self):
...
-        with tempfile.TemporaryFile() as tmpFile:
+        with tempfile.TemporaryFile(suffix='.qgraph') as tmpFile:
             self.qGraph.save(tmpFile)
             tmpFile.seek(0)
             restore = QuantumGraph.load(tmpFile, self.universe)
             self._cleanGraphs(self.qGraph, restore)
             self.assertEqual(self.qGraph, restore)
+            # Load in just one node
+            tmpFile.seek(0)
+            restoreSub = QuantumGraph.load(tmpFile, self.universe, nodes=(0,))
```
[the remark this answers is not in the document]
natelust: knowing the implementation, we would always have a node zero, it is just not guaranteed to be the same node. But you are right, it would be better to not depend on that so we can change it at some point in the future.

python/lsst/pipe/base/graph/graph.py
```diff
@@ -614,19 +637,75 @@ def save(self, file):
...
+        for taskDef in self.taskGraph:
+            # compressing has very little impact on saving or load time, but
+            # a large impact on on disk size, so it is worth doing
+            dump = lzma.compress(pickle.dumps(taskDef, protocol=protocol))
+            taskDefMap[taskDef.label] = (count, count+len(dump))
+            count += len(dump)
+            pickleData.append(dump)
+
+        # serialize the nodes, recording the start and end bytes of each node
+        self_list = list(self)
```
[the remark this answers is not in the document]
natelust: I think in a past version I was modifying during iteration, so I copied. You are right there is no need for that now.

python/lsst/pipe/base/graph/graph.py
```diff
@@ -614,19 +637,75 @@ def save(self, file):
...
+        for node in self_list:
+            node = copy.copy(node)
+            taskDef = node.taskDef
+            object.__setattr__(node, 'taskDef', taskDef.label)
```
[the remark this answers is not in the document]
natelust: This is because in normal operation, QuantumNodes are frozen dataclasses. However, because of the way things are being serialized for partial reads, there is no longer one big pickle file. Pickle is really efficient in that it would not serialize the same object multiple times, which we now were, and it was huge and slow. This is why the TaskDef objects were normalized out to their own save block.

We could have a different structure for this, but it seemed like a lot of overhead for something that is created to be dumped to disk, and then read back in and turned into something else. Especially as the structure of the class would be exactly the same, just with a different type declared. If you insist I will do it, but it seems to me to be one of those cases for which python flexibility is designed.

On a side note, I put in a comment around one of those blocks, but I guess I forgot to put it where it is saved as well. I will for sure add that.

python/lsst/pipe/base/graph/graph.py
```diff
@@ -614,19 +637,75 @@ def save(self, file):
...
+        map_lengths = struct.pack('>HQQ', SAVE_VERSION, len(taskDef_pickle), len(map_pickle))
+
+        # write each component of the save out in a deterministic order
+        buffer = io.BytesIO()
+        buffer.write(map_lengths)
+        buffer.write(taskDef_pickle)
+        buffer.write(map_pickle)
+        for dump in pickleData:
+            buffer.write(dump)
+        return buffer
```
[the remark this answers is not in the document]
natelust: Are you talking about the fact that there is the bytes of the object and then it written out to the buffer, or are you taking about the data structures inside the nodes?

python/lsst/pipe/base/graph/graph.py
```diff
@@ -614,19 +637,75 @@ def save(self, file):
...
+
+        # pickle the taskDef byte map
+        taskDef_pickle = pickle.dumps(taskDefMap, protocol=protocol)
+
+        # pickle the node byte map
+        map_pickle = pickle.dumps(nodeMap, protocol=protocol)
+
+        # record the sizes as 2 unsigned long long numbers for a total of 16
+        # bytes
+        map_lengths = struct.pack('>HQQ', SAVE_VERSION, len(taskDef_pickle), len(map_pickle))
```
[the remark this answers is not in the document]
natelust: I considered the secondary pickle, but it seemed like a lot of overhead for 3 numbers, and this is exactly what the struct module is designed for. There was also an issue where the unpickler seems to want 3 methods guaranteed on the handle to support multiple pickles in a file, and I didn't want to exclude the type of handles we could use.

python/lsst/pipe/base/graph/_loadHelpers.py
```diff
@@ -0,0 +1,310 @@
...
+
+
+# Create a custom dict that will return the desired default if a key is missing
+class RegistryDict(UserDict):
+    def __missing__(self, key):
+        return DefaultLoadHelper
+
+
+# Create a registry to hold all the load Helper classes
+HELPER_REGISTRY = RegistryDict()
```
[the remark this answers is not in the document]
natelust: I can be, I actually did it this way because I thought people would complain about the lambda :)

python/lsst/pipe/base/graph/_loadHelpers.py
```diff
@@ -0,0 +1,310 @@
...
+        # number and the two 8 bytes numbers that are the sizes of the byte
+        # maps
+        sizeBytes = self.uriObject.read(18)
+
+        # Turn they encode bytes back into a python int object
+        save_version, taskDefSize, nodeSize = struct.unpack('>HQQ', sizeBytes)
+
+        # Store the save version, so future read codes can make use of any
+        # format changes to the save protocol
+        self.save_version = save_version
```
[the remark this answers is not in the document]
natelust: Im not sure what you mean about "magic string", and which other comment are you referring to?

python/lsst/pipe/base/graph/_loadHelpers.py
```diff
@@ -0,0 +1,310 @@
...
+
+        # Recode the total header size
+        self.headerSize = 18+taskDefSize+nodeSize
```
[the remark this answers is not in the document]
natelust: strangely no.... not on my system, or on the github checks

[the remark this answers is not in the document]
timj: The style guide requires spaces around `+` but not around `*` and flake8 can't distinguish those so I think we have to disable the flake8 test. I'd much rather we had it enabled and changed the style guide.

python/lsst/pipe/base/graph/_loadHelpers.py
natelust: I did this because each subclass is going to be doing its own thing, and I didnt really want variables that were not important to the current class in the to be in the current class. I can think of a few other ways to do it, but I am not sure what you dislike about this approach, it is writing out exactly what I am trying to ask, and is not really much overhead compared to a None check.

python/lsst/pipe/base/graph/_loadHelpers.py
natelust: Yep, I didnt catch that when I was looking at the daf_butler implementation.

python/lsst/pipe/base/graph/_loadHelpers.py
natelust: See comment above

python/lsst/pipe/base/graph/_loadHelpers.py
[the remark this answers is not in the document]
natelust: After changing the code anyway I think I understand what you meant by this, and well its changed to stay read the string from the graph module now.

python/lsst/pipe/base/graph/graph.py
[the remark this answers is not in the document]
andy-slac: I'm taking about `dump` having all bytes that you want to dump and then you copy those into `buffer` too. It's a temporary spike for sure but it can be quite big spike. One way to remedy that is to remove pieces from `dump` as you write them into `buffer` but it may not help much depending on how memory allocation works. Also filling `buffer` will likely to cause re-allocations and fragmentation too. I don't know how to avoid all this mess if you want to construct whole buffer in memory before writing (or if it's worth spending time trying to improve it).

python/lsst/pipe/base/graph/graph.py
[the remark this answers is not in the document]
andy-slac: I agree that pickling is probably too much, having a fixed-size header should work better for your use case.

python/lsst/pipe/base/graph/_loadHelpers.py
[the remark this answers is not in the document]
andy-slac: Don't show it to those people 😉 I think lambda is a reasonable use for cases like this.

python/lsst/pipe/base/graph/_loadHelpers.py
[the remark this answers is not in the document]
andy-slac: "magic" can be any string that can help you uniquely identify file format. Right now you expect that file header is just a three numbers, but if by mistake someone gives you a random file that starts with a sequence `\0x00\0x01` then you will still try to interpret as a valid file. It will fail in pickle somewhere but it may be nicer to be able to identify that file format is wrong by choosing relatively unique bytes for header, e.g. `"qgraph\0x00\0x01...."`

python/lsst/pipe/base/graph/_loadHelpers.py
andy-slac: Having "dynamic" attributes that can be added or removed is a bit too much dynamic to my taste, probably due to my C++ background. If subclasses are doing something very different then you probably want a different base class or interface for them? This code sort of assumes that subclasses do not add "buffer" attribute or if they do add it then they know what they do. This sort of dependency may be better expressed as some (abstract) method in this class.

python/lsst/pipe/base/pipeline.py
```diff
@@ -136,7 +136,10 @@ def __str__(self):
     def __eq__(self, other: object) -> bool:
         if not isinstance(other, TaskDef):
             return False
-        return self.config == other.config and\
+        configDict = self.config.toDict()
+        otherDict = other.config.toDict()
+        configEq = configDict == otherDict or str(configDict) == str(otherDict)
```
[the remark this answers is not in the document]
andy-slac: Following up slack conversation - I think `Config.__eq__` should already handle `nan` comparison adequately: https://github.com/lsst/pex_config/blob/master/python/lsst/pex/config/config.py#L1368-L1379. There is also [`Config.compare()`](https://github.com/lsst/pex_config/blob/master/python/lsst/pex/config/config.py#L1393) method which does precision-base comparison but in this case I think we want bitwise comparison instead. If `__eq__` is not working then I agree that it needs to be fixed in`pex_config`.

python/lsst/pipe/base/graph/graph.py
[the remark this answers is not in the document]
natelust: Its a bit of a chicken and egg here, you cant dump directly, as the ordering that things is written is important here, and `ButlerURI` does not support subsequent write in append mode for all of the backends. I think there are tricks I can play with the buffer though so I will look into it.

python/lsst/pipe/base/pipeline.py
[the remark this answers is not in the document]
natelust: Somewhere in the hierarchy `__eq__` was not working right because it was attempting to look at the id of the two objects, even though their contents were the same. I will need to try and recreate that, the details are a bit fuzzy in my mind.

python/lsst/pipe/base/graph/graph.py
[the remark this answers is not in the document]
natelust: I think I might be able to avoid some of the copies using bytearrays, I will try and do some size tests.

python/lsst/pipe/base/graph/_loadHelpers.py
```diff
@@ -113,24 +125,59 @@ def __init__(self, uriObject: Union[ButlerURI, io.IO[bytes]]):
...
+        ------
+        ValueError
+            Raised if one or more of the nodes requested is not in the
+            `QuantumGraph` or if graphID parameter does not match the graph
+            being loaded.
+        """
         # need to import here to avoid cyclic imports
         from . import QuantumGraph
+        if graphID is not None and self._buildId != graphID:
+            raise ValueError('graphID does not match that of the graph being loaded')
```
andy-slac, line 179: Maybe also print both buildIds to simplify debugging?

python/lsst/pipe/base/graph/graph.py
```diff
@@ -680,19 +713,38 @@ def _buildSaveObject(self) -> io.BytesIO:
...
+        # from from the container.
+        # This must iterate over a range object even
+        # though the index is not used, because the container is modified with
+        # popleft and it is not safe to both modify and iterate over a
+        # container.
+        #
+        # Only this section needs to worry about memory pressue because
+        # everything else written to the buffer prior to this pickle data is
+        # only on the order of kilobytes to low numbers of megabytes.
+        for _ in range(len(pickleData)):
```
andy-slac, line 739: Why not just `while pickleData:`?

python/lsst/pipe/base/graph/graph.py
```diff
@@ -680,19 +713,38 @@ def _buildSaveObject(self) -> io.BytesIO:
...
+            buffer.extend(pickleData.popleft())
         return buffer
 
     @classmethod
-    def load(cls, file: io.IO[bytes], universe: DimensionUniverse, nodes: Optional[Iterable[int]] = None):
+    def load(cls, file: io.IO[bytes], universe: DimensionUniverse,
+             nodes: Optional[Iterable[int]] = None,
+             graphID: Optional[str] = None
```
andy-slac, line 746: MAybe `BuildId` instead of `str`?

tests/test_quantumGraph.py
```diff
@@ -367,10 +368,30 @@ def testSaveLoadUri(self):
...
+                # verify that more than one node works
+                nodeNumber2 = random.randint(0, len(self.qGraph)-1)
+                # ensure it is a different node number
+                while nodeNumber2 == nodeNumber:
+                    nodeNumber2 = random.randint(0, len(self.qGraph)-1)
+                restoreSub = QuantumGraph.loadUri(uri, self.universe, nodes=(nodeNumber, nodeNumber2))
+                self.assertEqual(len(restoreSub), 2)
+                self.assertEqual(set(restoreSub),
+                                 set((restore.getQuantumNodeByNodeId(NodeId(nodeNumber, restore._buildId)),
+                                     restore.getQuantumNodeByNodeId(NodeId(nodeNumber2, restore._buildId)))))
```
andy-slac, line 386: This might need re-indentation.

python/lsst/pipe/base/graph/graph.py
```diff
@@ -680,19 +713,38 @@ def _buildSaveObject(self) -> io.BytesIO:
...
+        # buffer = io.BytesIO()
+        # buffer.write(map_lengths)
+        # buffer.write(taskDef_pickle)
+        # buffer.write(map_pickle)
+        buffer = bytearray()
+        buffer.extend(map_lengths)
+        buffer.extend(taskDef_pickle)
+        buffer.extend(map_pickle)
+        # iterate over the length of pickleData, and for each element pop the
+        # leftmost element off the deque and write it out. This is to save
```
andy-slac, line 728: Trailing space on this line.

python/lsst/pipe/base/graph/graph.py
```diff
@@ -680,19 +713,38 @@ def _buildSaveObject(self) -> io.BytesIO:
...
+        buffer.extend(taskDef_pickle)
+        buffer.extend(map_pickle)
+        # iterate over the length of pickleData, and for each element pop the
+        # leftmost element off the deque and write it out. This is to save 
+        # memory, as the memory is added to the buffer object, it is removed
+        # from from the container.
+        # This must iterate over a range object even
+        # though the index is not used, because the container is modified with
+        # popleft and it is not safe to both modify and iterate over a
+        # container.
```
andy-slac, line 734: Maybe re-flow this comment, it looks jagged.

andy-slac (approved):
I checked last commit, looks good, few minor comments.

python/lsst/pipe/base/graph/graph.py
natelust: I like it

PR #167, opened 2020-12-17
DM-28036 Support URI in pipelines

natelust:
Support loading and saving files with a URI. This change necessitates
The label specifier on the URI path change to # from :.

### DIFF
--- a/python/lsst/pipe/base/pipeline.py
+++ b/python/lsst/pipe/base/pipeline.py
@@ -30,14 +30,17 @@ __all__ = ["Pipeline", "TaskDef", "TaskDatasetTypes", "PipelineDatasetTypes", "L
 # -------------------------------
 from dataclasses import dataclass
 from types import MappingProxyType
-from typing import Mapping, Set, Union, Generator, TYPE_CHECKING, Optional
+from typing import Mapping, Set, Union, Generator, TYPE_CHECKING, Optional, Tuple
 
 import copy
+import re
 import os
+import urllib.parse
+import warnings
 
 # -----------------------------
 #  Imports for other modules --
-from lsst.daf.butler import DatasetType, NamedValueSet, Registry, SkyPixDimension
+from lsst.daf.butler import DatasetType, NamedValueSet, Registry, SkyPixDimension, ButlerURI
 from lsst.utils import doImport
 from .configOverrides import ConfigOverrides
 from .connections import iterConnections
@@ -168,7 +171,7 @@ class Pipeline:
            A path that points to a pipeline defined in yaml format. This
            filename may also supply additional labels to be used in
            subsetting the loaded Pipeline. These labels are separated from
-           the path by a colon, and may be specified as a comma separated
+           the path by a \\#, and may be specified as a comma separated
            list, or a range denoted as beginning..end. Beginning or end may
            be empty, in which case the range will be a half open interval.
            Unlike python iteration bounds, end bounds are *INCLUDED*. Note
@@ -189,13 +192,49 @@ class Pipeline:
         string based matching due to the nature of contracts and may prune more
         than it should.
         """
-        # Split up the filename and any labels that were supplied
-        filename, labelSpecifier = cls._parseFileSpecifier(filename)
-        pipeline: Pipeline = cls.fromIR(pipelineIR.PipelineIR.from_file(filename))
+        return cls.from_uri(filename)
+
+    @classmethod
+    def from_uri(cls, uri: Union[str, ButlerURI]):
+        """Load a pipeline defined in a pipeline yaml file at a location
+        specified by a URI.
+
+        Parameters
+        ----------
+        uri: `str` or `ButlerURI`
+           If a string is supplied this should be a URI path that points to a
+           pipeline defined in yaml format. This uri may also supply
+           additional labels to be used in subsetting the loaded Pipeline.
+           These labels are separated from the path by a \\#, and may be
+           specified as a comma separated list, or a range denoted as
+           beginning..end. Beginning or end may be empty, in which case the
+           range will be a half open interval. Unlike python iteration
+           bounds, end bounds are *INCLUDED*. Note that range based selection
+           is not well defined for pipelines that are not linear in nature,
+           and correct behavior is not guaranteed, or may vary from run to
+           run. The same specifiers can be used with a ButlerURI object, by
+           being the sole contents in the fragments attribute.
+
+        Returns
+        -------
+        pipeline: `Pipeline`
+            The pipeline loaded from specified location with appropriate (if
+            any) subsetting
+
+        Notes
+        -----
+        This method attempts to prune any contracts that contain labels which
+        are not in the declared subset of labels. This pruning is done using a
+        string based matching due to the nature of contracts and may prune more
+        than it should.
+        """
+        # Split up the uri and any labels that were supplied
+        uri, label_specifier = cls._parse_file_specifier(uri)
+        pipeline: Pipeline = cls.fromIR(pipelineIR.PipelineIR.from_uri(uri))
 
         # If there are labels supplied, only keep those
-        if labelSpecifier is not None:
-            pipeline = pipeline.subsetFromLabels(labelSpecifier)
+        if label_specifier is not None:
+            pipeline = pipeline.subsetFromLabels(label_specifier)
         return pipeline
 
     def subsetFromLabels(self, labelSpecifier: LabelSpecifier) -> Pipeline:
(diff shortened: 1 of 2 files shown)


python/lsst/pipe/base/pipeline.py
```diff
@@ -189,9 +191,43 @@ def fromFile(cls, filename: str) -> Pipeline:
...
+        return cls.fromUri(filename)
+
+    @classmethod
+    def fromUri(cls, uri):
+        """Load a pipeline defined in a pipeline yaml file at a location
+        specified by a URI.
+
+        Parameters
+        ----------
+        uri: `str`
```
TallJimbo, line 203: I think most other middleware methods that take a URI actually take a `ButlerURI` instance instead of or in addition to a `str`.  And I _bet_ `ButlerURI`'s constructor will happily accept a `ButlerURI` instance, so this may just be a matter of updating the annotations and docs (but please don't take my word for it).

python/lsst/pipe/base/pipeline.py
```diff
@@ -263,16 +299,22 @@ def subsetFromLabels(self, labelSpecifier: LabelSpecifier) -> Pipeline:
...
-    def _parseFileSpecifier(fileSpecifer):
+    def _parseFileSpecifier(fileSpecifer) -> Tuple[str, Optional[LabelSpecifier]]:
         """Split appart a filename path from label subsets
         """
-        split = fileSpecifer.split(':')
+        # This is to support legacy pipelines during transition
+        split = re.split("[:](?!\\/\\/)", fileSpecifer)
+        if len(split) > 1:
+            warnings.warn("The fileSpecifier seems to use the legacy : to separate labels, this is "
+                          "deprecated and will be removed in June 2021, please use # instead.")
```
TallJimbo, line 309: A user that sees this warning is unlikely to know what "fileSpecifier" means, because that's sort of specific to this code context.   Maybe "The pipeline file {fileSpecifier!r} seems to use ..." instead?

You should probably also make this a `DeprecationWarning` or `FutureWarning`, rather than the default `UserWarning` (I don't recall the exact syntax, or which of those the dev guide recommends).

python/lsst/pipe/base/pipeline.py
```diff
@@ -484,6 +526,9 @@ def _addConfigImpl(self, label: str, newConfig: pipelineIR.ConfigIR):
     def toFile(self, filename: str):
         self._pipelineIR.to_file(filename)
 
+    def toUri(self, uri: str):
+        self._pipelineIR.to_uri(uri)
```
TallJimbo, line 530: Is this supposed to return something?  (Same for the preexisting `toFile`.)

If not, is it too late to rename these to something that more strongly suggests an in-place operation.

Finally, MyPy at least is going to complain if you don't add `-> None` to methods that don't return anything while annotating the arguments.

python/lsst/pipe/base/pipeline.py
```diff
@@ -484,6 +526,9 @@ def _addConfigImpl(self, label: str, newConfig: pipelineIR.ConfigIR):
...
+
```
TallJimbo, line 531: Pre-existing line below: `typing.Generator` requires three type parameters, not just one (the other two are `None` in this case).  FWIW, this is one of the reasons I  prefer `Iterator` when the user shouldn't care how the lazy iterator is implemented.

python/lsst/pipe/base/pipelineIR.py
```diff
@@ -791,18 +793,52 @@ def from_string(cls, pipeline_string: str):
...
             Location of document to use in creating a `PipelineIR` object.
+
+        Returns
+        -------
+        pipelineIR : `PipelineIR`
+            The loaded pipeline
+
+        Note
+        ----
+        This method is deprecated, please use from_uri
```
TallJimbo, line 812: Add a deprecated decorator or warning, then?

python/lsst/pipe/base/pipelineIR.py
```diff
@@ -791,18 +793,52 @@ def from_string(cls, pipeline_string: str):
...
+        return cls.from_uri(filename)
+
+    @classmethod
+    def from_uri(cls, uri: str) -> PipelineIR:
+        """Create a `PipelineIR` object from the document specified by the
+        input uri.
+
+        Parameters
+        ----------
+        uri: `str`
```
TallJimbo, line 823: As with `Pipeline.fromUri`, ideally this would probably take `ButlerURI` instances as well as `str` (and probably already actually does).

python/lsst/pipe/base/pipelineIR.py
```diff
@@ -791,18 +793,52 @@ def from_string(cls, pipeline_string: str):
...
+            Location of document to use in creating a `PipelineIR` object.
+        Returns
```
TallJimbo, line 825: Need a newline before this heading, I think.

python/lsst/pipe/base/pipelineIR.py
```diff
@@ -791,18 +793,52 @@ def from_string(cls, pipeline_string: str):
...
+        -------
+        pipelineIR : `PipelineIR`
+            The loaded pipeline
+        """
+        loadedUri = ButlerURI(uri)
```
TallJimbo, line 830: Really nitpicky, but I'll just note that you're using snake_case for method names here, camelCase for similar method names in Pipeline, camelCase with Biglittle acronyms for "Uri" (while @timj used BigBig for "URI"), and camelCase with BigBig acronyms for "IR".  You could at least use snake_case for variable names, too, now, and avoid at least _some_ of that inconsistency without any API changes.

Ok, so maybe it's not _that_ nitpicky, but also not something we can solve just on this ticket: I am in particular worried about us adding `fromUri`, `fromURI`, or `from_uri` (or similar) to ton of new things over the next few months with absolutely no consistency as we ButlerURI All The Things, and that actually is a real problem for users.

And if we are going to rebel against the dev guide's [_existing_ mandate](https://developer.lsst.io/cpp/style.html?highlight=acronym#id39) that acronyms should Biglittle (`Psf`, `Wcs`, even `Lsst`) - a rebellion I would support, but don't want to lead - we really should RFC it, not just ignore it.

python/lsst/pipe/base/pipelineIR.py
```diff
@@ -813,8 +849,19 @@ def to_file(self, filename: str):
         filename : `str`
             Location of document to write a `PipelineIR` object.
         """
-        with open(filename, 'w') as f:
-            yaml.dump(self.to_primitives(), f, sort_keys=False)
+        self.to_uri(filename)
+
+    def to_uri(self, uri: Union[ButlerURI, str]):
```
TallJimbo, line 854: I would rename this with "write" in the name somewhere ("write_to_uri"?); the name right now feels like it's supposed to somehow _transform_ the object into a URI and then return the URI (which obviously doesn't make sense, but it's still confusing).

Should also accept a `ButlerURI` instance as well as a string.

TallJimbo (approved):
Lots of nitpicky naming/typing comments, but nothing serious, except maybe for a naming inconsistency problem that is bigger than just this ticket (or just you).

python/lsst/pipe/base/pipeline.py
natelust: I'm happy to change the new method name, but technically I think I would have to go through a process to change the other, which does not seem worth it as at some point I would say we should just drop the old one.

python/lsst/pipe/base/pipeline.py
natelust: The typing "guidelines" I have read have been "Be liberal in what you accept, be specific with what you return" so I try to adhere to that as much as possible. It does not hurt to have more info about what is being returned, but it is a pain to get more specific later (especially in a case where it suddenly does matter)

python/lsst/pipe/base/pipelineIR.py
natelust: So this is a bit of a quirk of how things evolved over time. The snake vs camel came from pipelineIR when it was going to be more of a stand alone thing, and I thought we were moving toward that for new stuff. Then later this got more integrated with Pipeline.py, but there the convention was already camelCase, and so it was not ok to mix the two. So I am not quite sure what we want to do here. pipelineIR is basically all internal, so we could change this around w/o a lot of disruption. If we go this way I would prefer to do the other interface changes on a different ticket to do that. I am so conditioned to camelCase that I keep falling back to it on variable names. I will address any new ones here.

As far as URI vs Uri I tried to emulate the methods that Tim added in Quantum with the mixed case (and I think I missed that with one fromURI) So I guess the question remains, what would you prefer to see done? Maybe we could come up with a plan from there.

TallJimbo: @timj, do you want to make a call on how we capitalize and separate words in methods involving "URI" going forward?

Once we've settled where we're going, I think I'm basically happy for this PR to choose that or leave existing methods the same as @natelust thinks is best.

timj: We are at the confluence of competing naming forces. I'm fine if pipe_base switches to `xxx_uri` naming convention. I've always found `xxxUri` to be a bit odd but I realize that's the old convention.

natelust: For files that are already camelCase, would you prefer xxxUri or xxxURI? Or would you two be happy if I migrated some interfaces to snake case on this ticket, and others at a different time?

timj: I'm fine with incremental migration if it doesn't have much impact or take a lot of extra time. `xxxURI` should be used for classes that are related to `ButlerURI` otherwise things get odd but you are going to say they are already odd.

natelust: This is more about method names, but if you two are fine with starting migration to snake cake I am fine with that.

TallJimbo: Yeah, if there isn't an obvious best way to stick with an old style, better to jump all the way to snake case as the new style than land on something awkwardly in-between.

python/lsst/pipe/base/pipeline.py
```diff
@@ -263,42 +301,47 @@ def subsetFromLabels(self, labelSpecifier: LabelSpecifier) -> Pipeline:
         return Pipeline.fromIR(self._pipelineIR.subset_from_labels(labelSet))
 
     @staticmethod
-    def _parseFileSpecifier(fileSpecifer):
-        """Split appart a filename path from label subsets
+    def _parse_file_specifier(uri: Union[str, ButlerURI]
+                              ) -> Tuple[ButlerURI, Optional[LabelSpecifier]]:
+        """Split appart a uri and any possible label subsets
```
TallJimbo, line 306: appart -> apart

python/lsst/pipe/base/pipeline.py
```diff
@@ -263,42 +301,47 @@ def subsetFromLabels(self, labelSpecifier: LabelSpecifier) -> Pipeline:
...
-            return fileSpecifer, None
-        # More than one specifier provided, bail out
-        if len(split) > 2:
-            raise ValueError("Only one : is allowed when specifying a pipeline to load")
-        else:
-            labelSubset: str
-            filename: str
-            filename, labelSubset = split[0], split[1]
+        if isinstance(uri, str):
+            # This is to support legacy pipelines during transition
```
TallJimbo, line 309: Adding an example of what this matches to the comment would be nice; I've already forgotten.

TallJimbo (approved):

PR #173, opened 2021-03-16
DM-29255 Modify applying config overrides for flexibility

natelust:
These changes allow greater flexibility when doing config overrieds
in areas such as not needing string escaping when working with the
command line, and using values from python statements in other
parts of pipeline config blocks.

### DIFF
--- a/python/lsst/pipe/base/configOverrides.py
+++ b/python/lsst/pipe/base/configOverrides.py
@@ -25,13 +25,98 @@
 __all__ = ["ConfigOverrides"]
 
 import ast
+from operator import attrgetter
 
 import lsst.pex.exceptions as pexExceptions
 from lsst.utils import doImport
 
+import inspect
 from enum import Enum
 
 OverrideTypes = Enum("OverrideTypes", "Value File Python Instrument")
+
+
+class ConfigExpressionParser(ast.NodeVisitor):
+    """An expression parser that will be used to transform configuration
+    strings supplied from the command line or a pipeline into a python
+    object.
+
+    This is roughly equivalent to ast.literal_parser, but with the ability to
+    transform strings that are valid variable names into the value associated
+    with the name. Variables that should be considered valid are supplied to
+    the constructor as a dictionary that maps a string to its corresponding
+    value.
+
+    This class in an internal implementation detail, and should not be exposed
+    outside this module.
+
+    Parameters
+    ----------
+    namespace : `dict` of `str` to variable
+        This is a mapping of strings corresponding to variable names, to the
+        object that is associated with that name
+    """
+
+    def __init__(self, namespace):
+        self.variables = namespace
+
+    def visit_Name(self, node):
+        """This method gets called when the parser has determined a node
+        corresponds to a variable name.
+        """
+        # If the id (name) of the variable is in the dictionary of valid names,
+        # load and return the corresponding variable.
+        if node.id in self.variables:
+            return self.variables[node.id]
+        # If the node does not correspond to a valid variable, turn the name
+        # into a string, as the user likely intended it as such.
+        return f"{node.id}"
+
+    def visit_List(self, node):
+        """This method is visited if the node is a list. Constructs a list out
+        of the sub nodes.
+        """
+        return [self.visit(elm) for elm in node.elts]
+
+    def visit_Tuple(self, node):
+        """This method is visited if the node is a tuple. Constructs a list out
+        of the sub nodes, and then turns it into a tuple.
+        """
+        return tuple(self.visit_List(node))
+
+    def visit_Constant(self, node):
+        """This method is visited if the node is a constant
+        """
+        return node.value
+
+    def visit_Dict(self, node):
+        """This method is visited if the node is a dict. It builds a dict out
+        of the component nodes.
+        """
+        return {self.visit(key): self.visit(value) for key, value in zip(node.keys, node.values)}
+
+    def visit_Set(self, node):
+        """This method is visited if the node is a set. It builds a set out
+        of the component nodes.
+        """
+        return {self.visit(el) for el in node.elts}
+
+    def visit_UnaryOp(self, node):
+        """This method is visited if the node is a unary operator. Currently
+        The only operator we support is the negative (-) operator, all others
+        are passed to generic_visit method.
+        """
+        if isinstance(node.op, ast.USub):
+            value = self.visit(node.operand)
+            return -1*value
+        self.generic_visit(node)
+
+    def generic_visit(self, node):
+        """This method is called for all other node types. It will just raise
+        a value error, because this is a type of expression that we do not
+        support.
+        """
+        raise ValueError("Unable to parse string into literal expression")
 
 
 class ConfigOverrides:
@@ -114,6 +199,15 @@ class ConfigOverrides:
         instrument_lib = doImport(instrument)()
         self._overrides.append((OverrideTypes.Instrument, (instrument_lib, task_name)))
 
+    def _parser(self, value, configParser):
+        try:
+            value = configParser.visit(ast.parse(value, mode='eval').body)
+        except ValueError:
+            # eval failed, wrap exception with more user-friendly
+            # message
+            raise pexExceptions.RuntimeError(f"Unable to parse `{value}' into a Python object") from None
+        return value
+
     def applyTo(self, config):
         """Apply all overrides to a task configuration object.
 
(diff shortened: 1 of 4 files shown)


python/lsst/pipe/base/configOverrides.py
```diff
@@ -125,35 +227,77 @@ def applyTo(self, config):
         ------
         `Exception` is raised if operations on configuration object fail.
         """
+        # Look up a stack of variables people may be useing when setting
```
andy-slac, line 230: typo: useing

python/lsst/pipe/base/configOverrides.py
```diff
@@ -25,15 +25,108 @@
...
+        """This method gets called when the parser has determined a node
+        corresponds to a variable name.
+        """
+        # If the id (name) of the variable is in the dictionary of valid names,
+        # load and return the corresponding variable.
+        if node.id in self.variables:
+            return self.variables[node.id]
+        # If the node does not correspond to a valid variable, turn the name
+        # into a string, as the user likely intended it as such.
+        return f"{node.id}"
```
andy-slac, line 71: I have not looked into `ast` details, is not `node.id` a string already if `node` is a `Name`?

python/lsst/pipe/base/configOverrides.py
```diff
@@ -25,15 +25,108 @@
...
+
+    def visit_List(self, node):
+        """This method is visited if the node is a list. Constructs a list out
+        of the sub nodes.
+        """
+        result = []
+        for elm in node.elts:
+            result.append(self.visit(elm))
+        return result
```
andy-slac, line 80: `return [self.visit(elm) for elm in node.elts]` should be shorter?

andy-slac: And again, I know nothing about `ast` but would it be possible to skip implementing `visit_List` if you inherit from `NodeTransformer` instead of `NodeVisitor`?

python/lsst/pipe/base/configOverrides.py
```diff
@@ -25,15 +25,108 @@
...
+        for el in node.elts:
+            result.add(self.visit(el))
+        return result
+
+    def visit_UnaryOp(self, node):
+        """This method is visited if the node is a unary operator. Currently
+        The only operator we support is the negative (-) operator, all others
+        are passed to generic_visit method.
+        """
+        if isinstance(node.op, ast.USub):
```
andy-slac, line 117: `UAdd` may also be trivial to support here?

python/lsst/pipe/base/configOverrides.py
```diff
@@ -114,6 +207,15 @@ def addInstrumentOverride(self, instrument: str, task_name: str):
         instrument_lib = doImport(instrument)()
         self._overrides.append((OverrideTypes.Instrument, (instrument_lib, task_name)))
 
+    def _parser(self, value, configParser):
+        try:
+            value = configParser.visit(ast.parse(value, mode='eval').body)
+        except ValueError:
+            # eval failed, wrap exception with more user-friendly
+            # message
+            raise pexExceptions.RuntimeError(f"Unable to parse `{value}' into a Python object")
```
andy-slac, line 216: Add `from None` to `raise`?

python/lsst/pipe/base/configOverrides.py
```diff
@@ -125,35 +227,77 @@ def applyTo(self, config):
         ------
         `Exception` is raised if operations on configuration object fail.
         """
+        # Look up a stack of variables people may be useing when setting
+        # configs. Create a dictionary that will be used akin to a namespace
+        # for the duration of this function.
```
andy-slac, line 232: This is a cool feature, though I'm not sure I understand how it is going to work or whether it's going to pick some names unintentionally from higher stack levels. I think some documentation and rationale would be useful (in user-facing docs).

python/lsst/pipe/base/configOverrides.py
```diff
@@ -125,35 +227,77 @@ def applyTo(self, config):
...
+                # command line, and is handled above.
+                if isinstance(value, dict):
+                    new = {}
+                    for k, v in value.items():
+                        if isinstance(v, str):
+                            new[k] = self._parser(v, configParser)
+                        else:
+                            new[k] = v
+                    value = new
+                if isinstance(value, list):
```
andy-slac, line 265: nitpicking: `elif` instead of `if`?

python/lsst/pipe/base/configOverrides.py
```diff
@@ -125,35 +227,77 @@ def applyTo(self, config):
...
+                    for v in value:
+                        if isinstance(v, str):
+                            new.append(self._parser(v, configParser))
+                        else:
+                            new.append(v)
+                    value = new
+                # The field might be a string corresponding to a attribute
+                # hierarchy, attempt to split off the last field which
+                # will then be set.
+                parent, *child = field.rsplit(".", maxsplit=1)
```
andy-slac, line 276: `rpartition` may be slightly more readable, but that is a personal preference

python/lsst/pipe/base/configOverrides.py
```diff
@@ -125,35 +227,77 @@ def applyTo(self, config):
...
+                if child:
+                    # This branch means there was a hierarchy, get the
+                    # field to set, and look up the sub config for which
+                    # it is to be set
+                    finalField = child[0]
+                    tmpConfig = attrgetter(parent)(config)
```
andy-slac, line 282: Is not this the same as `getattr(config, parent)`?

python/lsst/pipe/base/configOverrides.py
```diff
@@ -125,35 +227,77 @@ def applyTo(self, config):
...
+                    tmpConfig = config
+                # set the specified config
+                setattr(tmpConfig, finalField, value)
+
             elif otype is OverrideTypes.Python:
-                exec(override, None, {"config": config})
+                # exec python string with the context of all vars known. This
+                # both lets people use a var they know about (maybe a bit
+                # dangerous, but not really more so than arbitrary python exec,
+                # and there are so many other places to worry about security
```
andy-slac, line 295: Sounds like "let someone else worry about it" 😄

tests/test_configOverrides.py
```diff
@@ -255,9 +239,23 @@ def testChoiceValueInt(self):
...
+        test_choice_value = 1  # noqa: F841
+        self.checkSingleFieldOverride(field, "test_choice_value", 1)
+
+        # Verify That python importing a variable works
+        config = ConfigTest()
+        overrides = ConfigOverrides()
+        overrides.addPythonOverride("from numpy import pi")
+        overrides.addValueOverride("fFloat", "pi")
+        overrides.applyTo(config)
+        from numpy import pi
```
andy-slac, line 257: I know `numpy` is practically required dependency now but maybe test can be done with `math.pi` instead to avoid this import?

andy-slac (approved):
Looks good, small bunch of minor comments/questions.

python/lsst/pipe/base/configOverrides.py
natelust: `NodeTransformer` is designed to transform an `ast`, into another `ast`, that will later be consumed by some other visitor system. In our case we actually want to take in the `ast` and emit actual python processing.

The methods that were implemented were chosen to represent the range of things we expect to handle processing snippets from pipelines etc. The advantage of this approach is that we can expand to handle more specific syntax in the future if we choose to.

python/lsst/pipe/base/configOverrides.py
natelust: It is, I just didnt expect someone to write something like `+17.2`, though I guess for symmetry it might be nice?

python/lsst/pipe/base/configOverrides.py
natelust: This feature exists basically to support (though it has other nice side effects) a few configuration level syntax being introduced on another ticket. The idea is something like this:

```python
label: TaskA
  class: some.PipelineTask
  config:
    action: A_Functor_Like_Action
    action.threshold: 17
    python: from action_library import A_Functor_Like_Action
```
This code would also allow variables to be found outside of an explicit import in certain cases. I was on the wall about letting people do imports prior to running `toExpandedPipeline` (where this code is triggered) and that is why the frame lookup is there. But like you say there might be unintended side effects, and so I think I will fall back to only allowing variable lookups in the python block, or in the module where the config was defined. This would still allow someone to import several standard actions that people might want to use, while likely not being too inclusive. We could always add more later if we choose to.

As far as documentation goes, I agree, and it will land either in Pipeline documentation or documentation on the new actions stuff when it comes out.

python/lsst/pipe/base/configOverrides.py
natelust: sort of, but with `getattr` you can only get a single attribute from a string. `attrgetter` supports a dotted notation of attributes to fetch; i.e. 'fieldA.fieldB.fieldC.fieldD`. This allows us to go through an arbitrarily deep config structure concisely without loops.

python/lsst/pipe/base/configOverrides.py
natelust: Yeah kind of, that is how it has been explained to me. If we are using pex config to defend against code attacks, so many other layers of computer infrastructure have failed, and people would likely be able to do anything they want anyway.

tests/test_configOverrides.py
natelust: sure, that would help import times too I guess.

python/lsst/pipe/base/configOverrides.py
andy-slac: Nice, I did not know about that.

PR #175, opened 2021-03-22
tickets/DM-29345 Add method to get QuantumNodes by TaskDef

natelust:
Add a method to the QuantumGraph object that returns all the
QuantumNodes associated with a TaskDef.

### DIFF
--- a/python/lsst/pipe/base/graph/graph.py
+++ b/python/lsst/pipe/base/graph/graph.py
@@ -103,6 +103,7 @@ class QuantumGraph:
 
         nodeNumberGenerator = count()
         self._nodeIdMap: Dict[NodeId, QuantumNode] = {}
+        self._taskToQuantumNode: DefaultDict[TaskDef, Set[QuantumNode]] = defaultdict(set)
         self._count = 0
         for taskDef, quantumSet in self._quanta.items():
             connections = taskDef.connections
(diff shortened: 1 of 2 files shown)


python/lsst/pipe/base/graph/graph.py
```diff
@@ -103,6 +103,7 @@ def _buildGraphs(self,
 
         nodeNumberGenerator = count()
         self._nodeIdMap: Dict[NodeId, QuantumNode] = {}
+        self._taskToQuantumNode: DefaultDict[TaskDef, Set[QuantumNode]] = defaultdict(set)
```
kfindeisen, line 106: I wonder if this is redundant with `self._quanta`, given that you can get the quantum of a `QuantumNode`.

kfindeisen (approved):
Looks good!

python/lsst/pipe/base/graph/graph.py
natelust: of course, this could be done the other way, of looking up all the QuantumNodes, extracting the Quantum from then and returning that, but I already had to have the data structure contained in memory of TaskDef to Set[Quantum] to construct the graph, so it seems wasteful to not just use that (again unless we want to free up a dict of pointers)

kfindeisen: Yes, clearly there's several ways of doing it. I wasn't actually thinking about memory usage, but about complexity -- redundant data structures are harder to keep in a self-consistent state, especially in the presence of exceptions.

PR #177, opened 2021-03-24
DM-29388 Treat strings that cant be parsed as strings

natelust:
When trying to apply config overrides, if a string cant be turned
into some other python expression, or object, leave it as a string.
If this assumption is bad, it will allow the config field to raise
its own exception.

### DIFF
--- a/python/lsst/pipe/base/configOverrides.py
+++ b/python/lsst/pipe/base/configOverrides.py
@@ -27,7 +27,6 @@ __all__ = ["ConfigOverrides"]
 import ast
 from operator import attrgetter
 
-import lsst.pex.exceptions as pexExceptions
 from lsst.utils import doImport
 
 import inspect
@@ -202,10 +201,13 @@ class ConfigOverrides:
     def _parser(self, value, configParser):
         try:
             value = configParser.visit(ast.parse(value, mode='eval').body)
-        except ValueError:
-            # eval failed, wrap exception with more user-friendly
-            # message
-            raise pexExceptions.RuntimeError(f"Unable to parse `{value}' into a Python object") from None
+        except Exception:
+            # This probably means it is a specific user string such as a URI.
+            # Let the value return as a string to attempt to continue to
+            # process as a string, another error will be raised in downstream
+            # code if that assumption is wrong
+            pass
+
         return value
 
     def applyTo(self, config):


python/lsst/pipe/base/configOverrides.py
```diff
@@ -202,10 +201,13 @@ def addInstrumentOverride(self, instrument: str, task_name: str):
     def _parser(self, value, configParser):
         try:
             value = configParser.visit(ast.parse(value, mode='eval').body)
-        except ValueError:
-            # eval failed, wrap exception with more user-friendly
-            # message
-            raise pexExceptions.RuntimeError(f"Unable to parse `{value}' into a Python object") from None
+        except Exception:
```
andy-slac, line 204: I am worried a bit about catching/ignoring too generic exception here, this could lead to some surprises (e.g. SyntaxError in the code above). Would it be possible to make it more specific?

andy-slac (approved):
Looks OK, one minor comment.

python/lsst/pipe/base/configOverrides.py
natelust: The issue ended up being, that if the user supplies a string that is "sqlite:///some_path_here", that is a syntax error when trying to parse the string as a python expression. So I could catch just ValueError and SyntaxError, but I was worried that some other string a user comes up with would throw some other kind of exception and we would be playing wack-a-mole. But I am happy to just pass on SyntaxError if you think that is more appropriate.

andy-slac: I can certainly imagine other errors generated when pretending that arbitrary string is a Python code.  Could parser catch all exceptions when parsing and raise something specific to distinguish errors in user string from any other errors?

natelust: I think the worst case is that something will be left as a string, and a config field will raise a validationError. Or I guess if it is a string field and something is really really weird, a config will be set to a weird value. But I'm going to leave that to another tickets work as this is a quick fix ticket.

timj: Does this mean if their string is "5+3" they will get a value of "8" and nothing will complain?

natelust: it means that the value with be int 8, and a string field will raise an exception that it expected a configuration to be supplied as a string and not an int. If the user really wanted the config field to be the string "3+5", then they would need to double quote it: `"'5+3'"`. This may be annoying, but it is much less likely to come up than the annoyances people were having about needing to double quote normal strings passed from the command line.

kfindeisen: Could this cause problems with config files (as opposed to command-line arguments, which is what the original DM-29388 bug was) not being able to be compatible with both Gen 2 and Gen 3 simultaneously? Or is quoting unchanged in that context?

natelust: Config files (through normal config file override use) should not have issues. There is a seperate handler method for that (`addFileOverride`) that does not use this bit of logic. This is only comes up for individual config field overrides, which come from the command line and Pipeline Files.

PR #178, opened 2021-04-06
DM-19634: Handle --longlog when seen.

ktlim:
Since this Gen2 code will be going away, I took a more expeditious route of handling `--longlog` when seen in the command line, so it will take effect before `--loglevel` options that come after (but still override those that come before).

### DIFF
--- a/python/lsst/pipe/base/argumentParser.py
+++ b/python/lsst/pipe/base/argumentParser.py
@@ -469,7 +469,7 @@ class ArgumentParser(argparse.ArgumentParser):
         self.add_argument("-L", "--loglevel", nargs="*", action=LogLevelAction,
                           help="logging level; supported levels are [trace|debug|info|warn|error|fatal]",
                           metavar="LEVEL|COMPONENT=LEVEL")
-        self.add_argument("--longlog", action="store_true", help="use a more verbose format for the logging")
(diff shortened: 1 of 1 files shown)


python/lsst/pipe/base/argumentParser.py
```diff
@@ -1252,6 +1243,35 @@ def __call__(self, parser, namespace, values, option_string):
         ident.idList += idDictList
 
 
+class LongLogAction(argparse.Action):
+    """argparse action to make logs verbose.
+
+    An action so that it can take effect before log level options.
```
andy-slac, line 1249: Does this actually mean "take effect before log level options that appear after --longlog"? I think this still does not handle `--loglevel=debug --longlog` ordering as expected. Is this what you meant by "good-enough fix"?

andy-slac (approved):
Looks OK, just one question.

python/lsst/pipe/base/argumentParser.py
ktlim: Yes, it does not handle `--loglevel` options that are placed before.  It's not great, but it's simpler than arranging to save all the loglevel arguments and play them after configuring longlog (which is what I think you do in Gen3).

PR #179, opened 2021-04-16
DM-28646 Introduce factory for creating lightweight butlers

natelust:
Introduce a factory class that is initialized with a butler that
can be used to convert QuantumGraphs into lightweight sqlite
backed butlers containing only the info from the graph.

### DIFF
--- a/python/lsst/pipe/base/makeLightWeightButler.py
+++ b/python/lsst/pipe/base/makeLightWeightButler.py
@@ -1,0 +1,246 @@
+# This file is part of pipe_base.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (http://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+from __future__ import annotations
+
+__all__ = ("buildLightweightButler", )
+
+import io
+
+from collections import defaultdict
+from typing import Callable, DefaultDict, Mapping, Optional, Set, Tuple, Iterable, List, Union
+import os
+import shutil
+
+from lsst.daf.butler import (DatasetRef, DatasetType, Butler, ButlerConfig, Registry, DataCoordinate,
+                             RegistryConfig)
+from lsst.daf.butler.core.utils import getClassOf
+from lsst.daf.butler.transfers import RepoExportContext
+
+
+from . import QuantumGraph, QuantumNode
+
+DataSetTypeMap = Mapping[DatasetType, Set[DataCoordinate]]
+
+
+def _accumulate(graph: QuantumGraph) -> Tuple[Set[DatasetRef], DataSetTypeMap]:
+    # accumulate the dataIds that will be transferred to the lightweight
+    # registry
+
+    # exports holds all the existing data that will be migrated to the
+    # lightweight butler
+    exports: Set[DatasetRef] = set()
+
+    # inserts is the mapping of DatasetType to dataIds for what is to be
+    # inserted into the registry. These are the products that are expected
+    # to be produced during processing of the QuantumGraph
+    inserts: DefaultDict[DatasetType, Set[DataCoordinate]] = defaultdict(set)
+
+    n: QuantumNode
+    for quantum in (n.quantum for n in graph):
+        for attrName in ("initInputs", "inputs", "outputs"):
+            attr: Mapping[DatasetType, Union[DatasetRef, List[DatasetRef]]] = getattr(quantum, attrName)
+
+            for type, refs in attr.items():
+                # This if block is because init inputs has a different
+                # signature for its items
+                if not isinstance(refs, list):
+                    refs = [refs]
+                # iterate over all the references, if it has an id, it
+                # means it exists and should be exported, if not it should
+                # be inserted into the new registry
+                for ref in refs:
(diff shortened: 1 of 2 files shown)


python/lsst/pipe/base/makeLightWeightButler.py
```diff
@@ -0,0 +1,234 @@
...
+from lsst.daf.butler.core.utils import getClassOf
+from lsst.daf.butler.transfers import RepoExportContext
+
+
+from . import QuantumGraph, QuantumNode
+
+DataSetTypeMap = Mapping[DatasetType, Set[DataCoordinate]]
+
+
+@dataclass
```
andy-slac, line 24: Does it need to be a dataclass? I guess the only useful thing that you get from it is `__init__`, but that is also trivial to write manually. 

And more generally I'm not sure that it needs to be a class, that probably depends on use cases, but my impression that all of it's functionaliy is a single method and maybe things can be made simpler by defining it as a standalone method?

andy-slac (approved):
Looks good, few minor comments.

python/lsst/pipe/base/makeLightWeightButler.py
natelust: So I like Dataclass when it there is no reason not to. I think it is more readable, and you get things like `__repr__` for free as well.

That said I agree with you on the class nature, I actually went back and forth a few times. I agree it comes down to use case. If we ever have the case where multiple graphs are converted it could be useful. Otherwise I am fine with it being a free function with 2 more arguments that calls private functions.

Do you have any preference?

python/lsst/pipe/base/makeLightWeightButler.py
```diff
@@ -0,0 +1,234 @@
...
+        # Recuse though any discovered collections to make sure all collections
+        # are exported. This exists because I ran into a situation where some
+        # collections were not properly being discovered and exported. This
+        # method may be able to be removed in the future if collection export
+        # logic changes
+        collections = set(collections)
+        while True:
+            discoveredCollections = set(self.butler.registry.queryCollections(collections, flattenChains=True,
+                                                                              includeChains=True))
+            if discoveredCollections - collections:
```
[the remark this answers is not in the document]
natelust: I'm sorry I am not getting what you are saying with this comment.

python/lsst/pipe/base/makeLightWeightButler.py
andy-slac: I think main utility of `dataclass` is still a collection of its data attributes (plus few standard dunder methods). In this case I don't think anyone cares that this class has `butler` and `collections` attributes, as the main and only purpose of this class is its `build` method. 

I have no strong preference for replacing class with a method as I don't know what all possible use case for it would be, so I'm OK with keeping it as it is now.

python/lsst/pipe/base/makeLightWeightButler.py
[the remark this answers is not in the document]
andy-slac: Sorry for confusion, I was just saying that `discoveredCollections - collections` creates new `set` which you do not really use so it feels a bit wasteful to create a whole `set` just to test that it's non-empty. Maybe another method to check that returned set has no extra elements will be more efficient? I guess it does not really matter in this particular case and I'm just being more efficiency maniac than I usually am 🙂

python/lsst/pipe/base/makeLightWeightButler.py
```diff
@@ -0,0 +1,246 @@
...
+
+def _export(butler: Butler, collections: Optional[Iterable[str]], exports: Set[DatasetRef]) -> io.StringIO:
+    # This exports the datasets that exist in the input butler using
+    # daf butler objects, however it reaches in deep and does not use the
+    # public methods so that it can export it to a string buffer and skip
+    # disk access.
+    yamlBuffer = io.StringIO()
+    # Yaml is hard coded, since the class controls both ends of the
+    # export/import
+    BackendClass = getClassOf(butler._config["repo_transfer_formats"]["yaml"]["export"])
```
andy-slac, line 108: I think this can be simplified or made more efficient as `_config["repo_transfer_formats", "yaml", "export"]`

python/lsst/pipe/base/makeLightWeightButler.py
```diff
@@ -0,0 +1,246 @@
...
+    backend = BackendClass(yamlBuffer)
+    exporter = RepoExportContext(butler.registry, butler.datastore, backend, directory=None, transfer=None)
+    exporter.saveDatasets(exports)
+
+    # Look for any defined collection, if not get the defaults
+    if (collections := collections) is None:
```
andy-slac, line 114: `collections := collections`? Could just check for `None`?

python/lsst/pipe/base/makeLightWeightButler.py
```diff
@@ -0,0 +1,246 @@
...
+            inserts: DataSetTypeMap,
+            run: str,
+            butlerModifier: Optional[Callable[[Butler], Butler]]
+            ) -> Butler:
+    # This method takes the exports from the existing butler, imports
+    # them into the newly created butler, and then inserts the datasets
+    # that are expected to be produced.
+
+    # import the existing datasets
+    newButler.import_(filename=yamlBuffer, format='yaml', reuseIds=True)
```
andy-slac, line 165: `'yaml'` - I think we prefer double quotes everywhere.

python/lsst/pipe/base/makeLightWeightButler.py
```diff
@@ -0,0 +1,246 @@
...
+def buildLightweightButler(butler: Butler,
+                           graph: QuantumGraph,
+                           outputLocation: str,
+                           run: str,
+                           *,
+                           clobber: bool = False,
+                           butlerModifier: Optional[Callable[[Butler], Butler]] = None,
+                           collections: Optional[Iterable[str]] = None
+                           ) -> None:
+    r"""buildLightweightButler is an object that is responsible for exporting
```
andy-slac, line 189: `an object` -> `a method`?

andy-slac (approved):
Re-reviewing, looks good, couple of minor comments.

python/lsst/pipe/base/makeLightWeightButler.py
natelust: Good catch, that was some version archeology

python/lsst/pipe/base/makeLightWeightButler.py
natelust: `a function` even

PR #181, opened 2021-05-14
DM-28649: Make changes to the execution butler builder

natelust:


python/lsst/pipe/base/executionButlerBuilder.py
```diff
@@ -126,26 +124,30 @@ def _export(butler: Butler, collections: Optional[Iterable[str]], exports: Set[D
...
+    config["registry", "db"] = "sqlite:///<butlerRoot>/gen3.sqlite3"
+    # record the current root of the datastore if it is specified relative
+    # to the butler root
+    if config.get(("datastore", "root")) == BUTLER_ROOT_TAG:
+        config['datastore', 'root'] = butler._config.configDir.geturl()
 
-    # Create the new registry which will create and populate the sqlite
-    # file.
-    Registry.createFromConfig(RegistryConfig(config))
+    Butler.makeRepo(root=outputLocation, config=config, overwrite=True, forceConfigRoot=False)
```
timj, line 150: It's slightly safer to pass the config returned by makeRepo to the Butler constructor rather than passing your constructed config from above. This ensures that everything is consistent.

python/lsst/pipe/base/executionButlerBuilder.py
```diff
@@ -243,4 +249,3 @@ def buildLightweightButler(butler: Butler,
     newButler = _setupNewButler(butler, outputLocation, dirExists)
 
     newButler = _import(yamlBuffer, newButler, inserts, run, butlerModifier)
```
timj, line 251: Why is this returning a Butler? It doesn't create one.

timj (approved):
Looks good. A couple of minor quibbles.

python/lsst/pipe/base/executionButlerBuilder.py
```diff
@@ -126,26 +124,30 @@ def _export(butler: Butler, collections: Optional[Iterable[str]], exports: Set[D
...
-    config = ButlerConfig(butler._config)
-    config["registry", "db"] = f"sqlite:///{outputLocation}/gen3.sqlite3"
-    config["root"] = butler._config.configDir.ospath
+    config = Config(butler._config)
+    config["root"] = outputLocation.geturl()
+    config["registry", "db"] = "sqlite:///<butlerRoot>/gen3.sqlite3"
+    # record the current root of the datastore if it is specified relative
+    # to the butler root
+    if config.get(("datastore", "root")) == BUTLER_ROOT_TAG:
+        config['datastore', 'root'] = butler._config.configDir.geturl()
```
timj, line 148: Just remembered. I think we want this execution butler to also enable 
```python
config["datastore", "trust_get_request"] = True
```
otherwise the distributed workflow won't function.

timj: PS the string quoting is inconsistent on this line relative to the others.

python/lsst/pipe/base/executionButlerBuilder.py
```diff
@@ -233,8 +237,10 @@ def buildLightweightButler(butler: Butler,
         Raise if something exists in the filesystem at the specified output
         location and clobber is `False`
     """
+    outputLocation = ButlerURI(outputLocation)
+
     # Do this first to Fail Fast if the output exists
-    if (dirExists := os.path.exists(outputLocation)) and not clobber:
+    if (dirExists := outputLocation.exists()) and not clobber:
```
timj, line 243: You might want to add a check here as well for `outputLocation.isdir()` so you can stop early if they've given you a file URI.

python/lsst/pipe/base/executionButlerBuilder.py
natelust: Good catch, this was sort of twisted up in a refactor, I intended to have it return the newly built butler in case someone (perhaps bps) wanted to make use of it. I will fix that up

PR #182, opened 2021-05-17
DM-28668: PipelineTask unit test framework bypasses dimensions checks

kfindeisen:
This PR tries to catch connections dimension errors by checking that supplied data IDs in testing match the connections (on the assumption that a mismatch might mean the connections aren't what the test author expected) and by providing a separate function for detecting common errors.

### DIFF
--- a/doc/lsst.pipe.base/testing-a-pipeline-task.rst
+++ b/doc/lsst.pipe.base/testing-a-pipeline-task.rst
@@ -234,3 +234,29 @@ There is currently no test framework for the use of init-inputs in task construc
    task = OrTask(config=config)
    run = testUtils.runTestQuantum(task, butler, quantum)
    run.assert_called_once_with(cat=testCatalog)
+
+.. _testing-a-pipeline-task-static-analysis:
+
+Analyzing Connections Classes
+=============================
+
+Mistakes in creating pipeline connections classes can lead to hard-to-debug errors at run time.
+The `lsst.pipe.base.testUtils.lintConnections` function analyzes a connections class for common errors.
+The only errors currently tested are those involving inconsistencies between connection and quantum dimensions.
+
+All tests done by `lintConnections` are heuristic, looking for common patterns of misuse.
+Advanced users who are *deliberately* bending the usual rules can use keywords to turn off specific tests.
+
+.. code-block:: py
+   :emphasize-lines: 9-10
+
+   class ListConnections(PipelineTaskConnections,
+                         dimensions=("instrument", "visit", "detector")):
+       cat = connectionTypes.Input(
+           name="src",
+           storageClass="SourceCatalog",
+           dimensions=("instrument", "visit", "detector"),
+           multiple=True)  # force a list of one catalog
+
+   lintConnections(ListConnections)  # warns that cat always has one input
+   lintConnections(ListConnections, checkUnnecessaryMultiple=False)  # passes
--- a/python/lsst/pipe/base/testUtils.py
+++ b/python/lsst/pipe/base/testUtils.py
@@ -33,7 +33,8 @@ import collections.abc
 import itertools
 import unittest.mock
 
-from lsst.daf.butler import DataCoordinate, DatasetRef, DatasetType, Quantum, StorageClassFactory
+from lsst.daf.butler import DataCoordinate, DatasetRef, DatasetType, Quantum, StorageClassFactory, \
+    SkyPixDimension
 from lsst.pipe.base import ButlerQuantumContext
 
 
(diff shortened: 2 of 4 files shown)


python/lsst/pipe/base/testUtils.py
```diff
@@ -362,3 +426,47 @@ def getInitInputs(butler, config):
...
+    """Inspect a connections class for common errors.
+
+    These tests are designed to detect misuse of connections features in
+    standard designs. An unusually designed connections class may trigger
+    alerts despite being correctly written; specific checks can be turned off
+    using keywords.
+
+    Parameters
+    ----------
+    connections : `lsst.pipe.base.PipelineTaskConnections`-type
```
taranu, line 444: What does the `-type` signify here? I checked the dev guide and numpydoc guide and saw similar usages but no explanation.

python/lsst/pipe/base/testUtils.py
```diff
@@ -362,3 +426,47 @@ def getInitInputs(butler, config):
...
+    # Since all comparisons are inside the class, don't bother
+    # normalizing skypix.
+    quantumDimensions = connections.dimensions
+
+    # connectionTypes.DimensionedConnection is implementation detail,
+    # don't use it.
+    for name in itertools.chain(connections.inputs, connections.prerequisiteInputs, connections.outputs):
+        connection = connections.allConnections[name]
+        connDimensions = set(connection.dimensions)
+        if checkMissingMultiple and not connection.multiple and connDimensions > quantumDimensions:
```
taranu, line 467: Consider having a single error message string, appending to it and asserting after every condition is checked rather than immediately after the first failure.

doc/lsst.pipe.base/testing-a-pipeline-task.rst
```diff
@@ -234,3 +234,29 @@ There is currently no test framework for the use of init-inputs in task construc
...
+
+   class ListConnections(PipelineTaskConnections,
+                         dimensions=("instrument", "visit", "detector")):
+       cat = connectionTypes.Input(
+           name="src",
+           storageClass="SourceCatalog",
+           dimensions=("instrument", "visit", "detector"),
+           multiple=True)  # force a list of one catalog
+
+   lintConnections(ListConnections)  # warns that cat always has one input
```
taranu, line 261: Is this the pattern that users should follow? i.e. should we be adding `lintConnections(xxConnections)` to existing and/or new `Connections` classes?

tests/test_testUtils.py
```diff
@@ -574,6 +598,38 @@ def testSkypixHandling(self):
...
+                dimensions={"tract", "patch", "band", "skymap"},
+            )
+
+        with self.assertRaises(AssertionError):
+            lintConnections(BadConnections)
+        lintConnections(BadConnections, checkMissingMultiple=False)
+
+    def testLintConnectionsExtraMultiple(self):
+        class BadConnections(PipelineTaskConnections,
+                             dimensions=["tract", "patch", "band", "skymap"]):
```
taranu, line 621: I'm presuming that making this a list instead of a set is what makes this a bad `Connections`, but it's not immediately obvious to me why, so maybe add a brief explanatory comment?

taranu (commented):
The changes seem fine; I just had a couple of questions and some suggestions that you can feel free to ignore if they're not helpful.

python/lsst/pipe/base/testUtils.py
```diff
@@ -61,29 +63,89 @@ def makeQuantum(task, butler, dataId, ioDataIds):
...
+        dimensions simplified to ``skypix``.
+    """
+    simplified = set()
+    for dimension in dimensions:
+        # skypix not a real Dimension, handle it first
+        if dimension == "skypix":
+            simplified.add(dimension)
+        else:
+            # Need a Dimension to test spatialness
+            fullDimension = universe[dimension] if isinstance(dimension, str) else dimension
```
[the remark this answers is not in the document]
kfindeisen: I think at the very least that would fail formal type-checking (`DimensionUniverse.get` expects a string), and it might cause bugs depending on how `get` is implemented.

python/lsst/pipe/base/testUtils.py
kfindeisen: It means a `type` object representing that type (or a subtype thereof).

doc/lsst.pipe.base/testing-a-pipeline-task.rst
kfindeisen: I'm not sure I would say "should"; it's a tool that may be useful as part of a broader unit testing strategy.

tests/test_testUtils.py
kfindeisen: It should be irrelevant. I'll add a comment that I'm trying to test the linter's robustness to different collection types.

taranu (approved):

PR #183, opened 2021-05-20
DM-29761: start saving per-quantum provenance and propagating nothing-to-do cases

TallJimbo:


python/lsst/pipe/base/connections.py
```diff
@@ -547,6 +547,24 @@ def translateAdjustQuantumInputs(
             )
         return results
 
+    def hasPostWriteLogic(self):
```
natelust, line 572: Why is this a method?

TallJimbo: :man_shrugging: , I just never know what to do with boolean flag getters; I feel like "starts with verb" implies "it's a method", but there's pretty much no other way to name boolean properties.  I'm not even consistent about that within this commit.

Everybody else in the world, please just agree on what we should do in this case and I'll happily follow along.

timj: Is the implied rest of the question "and not a property"?

natelust: Or just an attribute, That's the great thing about properties is that they can be migrated or substituted in place of an attribute if needed

TallJimbo: This one shouldn't be an attribute, because I think the right specialization pattern is "override method [or property]".

natelust: Why is that better than setting a class level attribute? I really don't understand why this is such an uncommon thing. Are you trying to protect it from changing, and if so other code would be needed here.

PR #185, opened 2021-05-25
DM-30335: Minor fixes to execution butler

timj:


python/lsst/pipe/base/executionButlerBuilder.py
```diff
@@ -108,6 +109,11 @@ def _export(butler: Butler, collections: Optional[Iterable[str]], exports: Set[D
     exporter = RepoExportContext(butler.registry, butler.datastore, backend, directory=None, transfer=None)
     exporter.saveDatasets(exports)
 
+    # Need to ensure that the dimension records for outputs are
+    # transferred.
+    for _, dataIds in inserts.items():
+        exporter.saveDataIds(dataIds)
```
TallJimbo, line 115: I _think_ this should be fine, but we should be on the lookout for it running into the problem @SimonKrughoff reported [here](https://lsstc.slack.com/archives/C01FBUGM2CV/p1621983930190800) where we export visit records but not visit definition records.  I'm not sure exactly what he did differently.

TallJimbo (approved):

python/lsst/pipe/base/executionButlerBuilder.py
```diff
@@ -141,14 +147,19 @@ def _setupNewButler(butler: Butler, outputLocation: ButlerURI, dirExists: bool)
...
     # to the butler root
     if config.get(("datastore", "root")) == BUTLER_ROOT_TAG:
         config["datastore", "root"] = butler._config.configDir.geturl()
     config["datastore", "trust_get_request"] = True
 
-    config = Butler.makeRepo(root=outputLocation, config=config, overwrite=True, forceConfigRoot=False)
+    # Requires that we use the dimension configuration from the original
+    # butler and not use the defaults.
+    config = Butler.makeRepo(root=outputLocation, config=config,
+                             dimensionConfig=butler.registry.dimensions.dimensionConfig,
```
natelust, line 161: I wonder if this is the difference between how I was doing things before with initializing the registry explicity vs switching to `butler.makeRepo` to make the registry in terms of dimensions being transfered.

timj: We haven't changed dimensions config in a long time so I don't think this would have mattered. It was going to be a potential problem in the future though.

PR #186, opened 2021-06-03
DM-30564: Improve task label uniqueness exception message

parejkoj:


python/lsst/pipe/base/pipelineIR.py
```diff
@@ -619,10 +619,14 @@ def process_args(argument: Union[str, dict]) -> dict:
             if self.instrument is None:
                 self.instrument = tmp_IR.instrument
             elif self.instrument != tmp_IR.instrument and tmp_IR.instrument is not None:
-                raise ValueError("Only one instrument can be declared in a pipeline or it's imports")
-            if accumulate_tasks.keys() & tmp_IR.tasks.keys():
-                raise ValueError("Task labels in the imported pipelines must "
-                                 "be unique")
+                msg = ("Only one instrument can be declared in a pipeline or its imports. "
+                       f"These instruments were found: {self.instrument}, {tmp_IR.instrument}")
```
natelust, line 623: If you are expanding the exception message, might as well also say something like:
```
"Top level pipeline defines {self.instrument} but {other_pipeline.location} defines {tmp_IR.instrument}"
```
To help people track down where the other declaration was.

tests/test_pipelineIR.py
```diff
@@ -352,14 +352,15 @@ def testImportingInstrument(self):
...
+        # Test that multiple instruments can't be defined,
+        # and that the error message tells you what instruments were found.
         pipeline_str = textwrap.dedent("""
         description: Test Pipeline
         instrument: new.instrument
         imports:
           - location: $PIPE_BASE_DIR/tests/testPipeline1.yaml
         """)
-        with self.assertRaises(ValueError):
+        with self.assertRaisesRegex(ValueError, "new.instrument, test.instrument"):
```
natelust, line 363: If you take the advice above, make sure this still works, though I suspect it will

python/lsst/pipe/base/pipelineIR.py
```diff
@@ -619,10 +619,14 @@ def process_args(argument: Union[str, dict]) -> dict:
...
+                raise ValueError(msg)
+            duplicate_labels = accumulate_tasks.keys() & tmp_IR.tasks.keys()
+            if duplicate_labels:
```
natelust, line 626: You are free to leave it as is, but this is exactly the kind of place warlrus was designed for, variable is only used in one context:
```
if duplicate_labels := accumulate_tasks.keys() & tmp_IR.tasks.keys():
    msg = ("Task labels in the imported pipelines must be unique. "
           f"These labels appear multiple times: {duplicate_labels}")
```

natelust (approved):

python/lsst/pipe/base/pipelineIR.py
parejkoj: Thank you, that's a very useful addition.

PR #189, opened 2021-06-15
DM- 30702 Store Metadata in QuantumGraph

natelust:
## Checklist

- [ x ] ran Jenkins - https://ci.lsst.codes/job/stack-os-matrix/34456/display/redirect
- [ x ] added a release note for user-visible changes to `doc/changes`

### DIFF
--- a/python/lsst/pipe/base/graph/_loadHelpers.py
+++ b/python/lsst/pipe/base/graph/_loadHelpers.py
@@ -20,7 +20,7 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 from __future__ import annotations
 
-__all__ = ("LoadHelper")
+__all__ = ("LoadHelper", )
 
 from lsst.daf.butler import ButlerURI, Quantum
 from lsst.daf.butler.core._butlerUri.s3 import ButlerS3URI
@@ -32,12 +32,13 @@ from .quantumNode import NodeId
 from dataclasses import dataclass
 import functools
 import io
+import json
 import lzma
 import pickle
 import struct
 
 from collections import defaultdict, UserDict
-from typing import (Optional, Iterable, DefaultDict, Set, Dict, TYPE_CHECKING, Type, Union)
+from typing import (Optional, Iterable, DefaultDict, Set, Dict, TYPE_CHECKING, Tuple, Type, Union)
 
 if TYPE_CHECKING:
     from . import QuantumGraph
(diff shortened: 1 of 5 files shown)


timj (commented):
Is it possible to add a little test that creates a quantum graph with a metadata header and shows that you can read the metadata back in?

python/lsst/pipe/base/graph/graph.py
```diff
@@ -51,12 +53,16 @@
 
 # modify this constant any time the on disk representation of the save file
 # changes, and update the load helpers to behave properly for each version.
-SAVE_VERSION = 1
+SAVE_VERSION = 2
 
 # String used to describe the format for the preamble bytes in a file save
 # This marks a Big endian encoded format with an unsigned short, an unsigned
 # long long, and an unsigned long long in the byte stream
```
andy-slac, line 60: This comment probably needs small update to describe versions 1 and 2 differences.

python/lsst/pipe/base/graph/graph.py
```diff
@@ -696,15 +726,11 @@ def _buildSaveObject(self) -> bytearray:
             # compressing has very little impact on saving or load time, but
             # a large impact on on disk size, so it is worth doing
             dump = lzma.compress(pickle.dumps(taskDef, protocol=protocol))
-            taskDefMap[taskDef.label] = (count, count+len(dump))
+            taskDefMap[taskDef.label] = {"bytes": (count, count+len(dump))}
```
andy-slac, line 729: Is there only one key ("bytes") in this dict? I guess this is to make things more extensible in the future but this also add additional incompatibility between versions 1 and 2. Same question for `nodeMap` below.

python/lsst/pipe/base/graph/_loadHelpers.py
```diff
@@ -107,58 +108,87 @@ class DefaultLoadHelper:
...
+            # read the map of the node objects back in skipping bytes
+            # corresponding to the taskDef byte map
+            self.map = pickle.loads(headerMaps[taskDefSize:])
 
-        # read the map of the node objects back in skipping bytes
-        # corresponding to the taskDef byte map
-        self.map = pickle.loads(headerMaps[taskDefSize:])
+            # There is no metadata for old versions
+            self.metadata = None
+        elif self.save_version >= 2:
```
andy-slac, line 185: I would probably be more defensive here and only check for `self.save_version == 2` and add `else: raise ...` to force explicit verification when new version is added. Same for similar check in `__init__`.

andy-slac (approved):
Looks good, few minor comments/questions.

natelust:
@timj That is a good catch, I had intended to, and then completely forgot by the time I finished fixing a failure from an existing test.

python/lsst/pipe/base/graph/graph.py
natelust: I am setting this up for another upcoming ticket which I should be able to speed up load times. I wanted to change the format now, instead of having a larger change in version 3 of the file format.

This also plays into some possible future changes if/when QuantumDirectories come into play.

python/lsst/pipe/base/graph/_loadHelpers.py
natelust: I went back and forth on this. I ended up settling on this pattern for 2 reasons.
1) People would be editing the branch anyway
2) I could not predict where someone would want to branch the logic, i.e.
```
elif self.save_version >= 2:
    <some shared logic>
    if self.save_version == 3:
        <do some logic extra for version 3>
```
or
```
elif self.save_version == 2:
    <some version 2 logic>
elif self.save_version >= 3:
    <some new logic all together>
```
I normally would not assume that we should force someone to review various parts of code that might not be related to their change, but in this case I can see your point.

PR #190, opened 2021-06-18
DM-30767: include initOutputs in execution butler creation

TallJimbo:
## Checklist

- [ ] ran Jenkins

### DIFF
--- a/python/lsst/pipe/base/executionButlerBuilder.py
+++ b/python/lsst/pipe/base/executionButlerBuilder.py
@@ -25,6 +25,7 @@ __all__ = ("buildExecutionButler", )
 import io
 
 from collections import defaultdict
+import itertools
 from typing import Callable, DefaultDict, Mapping, Optional, Set, Tuple, Iterable, List, Union
 
 from lsst.daf.butler import (DatasetRef, DatasetType, Butler, DataCoordinate, ButlerURI, Config)
@@ -33,12 +34,16 @@ from lsst.daf.butler.transfers import RepoExportContext
 from lsst.daf.butler.core.repoRelocation import BUTLER_ROOT_TAG
 
 from .graph import QuantumGraph, QuantumNode
+from .pipeline import PipelineDatasetTypes
 
 DataSetTypeMap = Mapping[DatasetType, Set[DataCoordinate]]
 
 
-def _accumulate(graph: QuantumGraph) -> Tuple[Set[DatasetRef], DataSetTypeMap]:
-    # accumulate the dataIds that will be transferred to the execution
+def _accumulate(
+    graph: QuantumGraph,
+    dataset_types: PipelineDatasetTypes,
+) -> Tuple[Set[DatasetRef], DataSetTypeMap]:
+    # accumulate the DatasetRefs that will be transferred to the execution
     # registry
 
     # exports holds all the existing data that will be migrated to the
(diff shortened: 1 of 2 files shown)


python/lsst/pipe/base/executionButlerBuilder.py
```diff
@@ -265,7 +276,11 @@ def buildExecutionButler(butler: Butler,
     if not outputLocation.isdir():
         raise NotADirectoryError("The specified output URI does not appear to correspond to a directory")
 
-    exports, inserts = _accumulate(graph)
+    # Gather all DatasetTypes from the Python and check any that already exist
+    # in the registry for consistency.
```
natelust, line 280: Are you relying on the registry get datasettype method to raise if it is not found in the registry for this comments behavior?

python/lsst/pipe/base/pipeline.py
```diff
@@ -747,6 +756,18 @@ def makeDatasetTypesSet(connectionType: str, freeze: bool = True) -> NamedValueS
                 datasetTypes.freeze()
             return datasetTypes
 
+        # optionally add initOutput dataset for config
+        initOutputs = makeDatasetTypesSet("initOutputs", freeze=False)
+        if include_configs:
```
natelust, line 761: Should this logic be here or inside the makeDatasetTypeSet method? Do we only expect this to be used here?

natelust (approved):

python/lsst/pipe/base/executionButlerBuilder.py
TallJimbo: Well, it does call `Registry.getDatasetType`, but it actually catches that exception and proceeds if the missing dataset type is an output, because we want to allow the user to only register the dataset types in the main repo later.

python/lsst/pipe/base/pipeline.py
TallJimbo: It's only appropriate for `initOutputs`, so I don't think we want it in `makeDatasetType`, since that's called for all of the connection types.

python/lsst/pipe/base/executionButlerBuilder.py
natelust: This seems inconsistent with the comment that is here. I saw the exception catching, and was thinking there must be something else going on based on this comment. This also leads to the question, how are we propagating the info if --register-dataset-types is passed to know if they should be inserted at the end.

I personally think we should check and throw if they are not in the registry and --register-dataset-types is not in args, but not actually do the registration until the end. This will prevent the case where someone mistypes a name, and they find out at the beginning instead of the end when they have a weird type.

TallJimbo: Adjusted the comment to clarify and created https://jira.lsstcorp.org/browse/DM-30845.

timj: I'm pretty sure that if someone is creating an execution butler they want the "register-dataset-types" to apply to the execution butler and not the upstream butler -- the question is whether having both options is fatal, or does the right thing and ignores main registry, or is really taken to create in main registry. Are we explicitly saying that if you have the execution butler option set that we are guaranteeing that main registry will never be touched (so writeable=False for that butler)

TallJimbo: The user cannot have a choice about whether to register dataset types in the execution butler - if they don't, it will fail.  So there's no option for that at present, and that's reasonable.

And yes, we do guarantee that if you're *just* making an execution butler, the main registry will never be touched (via `writeable=False`).  That's true more generally of `pipetask qgraph`, which is the subcommand that BPS should be using to do that.  I think it's possible to ask `pipetask run` to make an execution butler but that feels like a mistake in the click stuff, even if it even works.

PR #191, opened 2021-07-09
DM-30301: Change Task logging from lsst.log to python logging

timj:
## Checklist

- [ ] ran Jenkins

### DIFF
--- a/python/lsst/pipe/base/argumentParser.py
+++ b/python/lsst/pipe/base/argumentParser.py
@@ -632,7 +632,9 @@ log4j.appender.A1.layout.ConversionPattern=%c %p: %m%n
             self.error(f"Error: input={namespace.input!r} not found")
 
         namespace.config = config
-        namespace.log = log if log is not None else getTaskLogger()
+        # Ensure that the external logger is converted to the expected
+        # logger class.
+        namespace.log = getTaskLogger(log.name) if log is not None else getTaskLogger()
         mapperClass = dafPersist.Butler.getMapperClass(namespace.input)
         if mapperClass is None:
             self.error(f"Error: no mapper specified for input repo {namespace.input!r}")
--- a/python/lsst/pipe/base/argumentParser.py
+++ b/python/lsst/pipe/base/argumentParser.py
@@ -1318,7 +1318,7 @@ class LogLevelAction(argparse.Action):
                 pyLevel = lsstLog.LevelTranslator.lsstLog2logging(logLevel)
                 logging.getLogger(component or None).setLevel(pyLevel)
             else:
-                # Need to see lsstLog level
+                # Need to set lsstLog level
                 lsstLogLevel = lsstLog.LevelTranslator.logging2lsstLog(logLevel)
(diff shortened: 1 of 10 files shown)


python/lsst/pipe/base/argumentParser.py
```diff
@@ -1299,17 +1300,27 @@ def __call__(self, parser, namespace, values, option_string):
...
+            # Set logging level for whatever logger this wasn't.
+            if isinstance(logger, lsstLog.Log):
+                pyLevel = lsstLog.LevelTranslator.lsstLog2logging(logLevel)
+                logging.getLogger(component or None).setLevel(pyLevel)
             else:
-                lsstLog.Log.getLogger(component).setLevel(logLevel)
-            # set logging level for Python logging
-            pyLevel = lsstLog.LevelTranslator.lsstLog2logging(logLevel)
-            logging.getLogger(component).setLevel(pyLevel)
+                # Need to see lsstLog level
```
ktlim, line 1321:
```suggestion
                # Need to set lsstLog level
```

python/lsst/pipe/base/task.py
```diff
@@ -155,12 +160,16 @@ def __init__(self, config=None, name=None, parentTask=None, log=None):
                 config = self.ConfigClass()
             self._taskDict = dict()
             loggerName = self._fullName
-            if log is not None and log.getName():
-                loggerName = log.getName() + '.' + loggerName
+            if log is not None and (name := log.name):
```
ktlim, line 163: Is the walrus really worth it here?  Seems like it adds 10 characters here to save 4 characters on the next line; the extra property lookup shouldn't be terrible.

python/lsst/pipe/base/task_logging.py
```diff
@@ -0,0 +1,244 @@
...
+    @property
+    def TRACE(self):
+        return LOG_TRACE
+
+    @property
+    def VERBOSE(self):
+        return LOG_VERBOSE
+
+    @property
+    @deprecated(reason="Use logging.DEBUG. Will be removed after v23.",
```
ktlim, line 94: If we're going to need to use `logger.VERBOSE`, I think we should continue to allow `logger.DEBUG`.

ktlim (approved):

python/lsst/pipe/base/task_logging.py
timj: I've changed all the deprecations if you would like to take a look. I now am allowing all the constants to exist without deprecation.  I've also made TaskLogAdapter public (mostly so it gets documented in the sphinx docs) but I'm okay with reversing that and instead adding an explicit line in the sphinx.

python/lsst/pipe/base/task.py
timj: In fact I've realized we should be doing this:

```python
if log is not None and log.name:
    loggerName = log.getChild(loggerName).name
```
and may be simpler to just do:
```python
self.log = getTaskLogger(name, log)
```

PR #192, opened 2021-07-14
DM-31100 Introduce a cleaner helper module for QuantumGraph

natelust:

### DIFF
--- a/python/lsst/pipe/base/graph/_implDetails.py
+++ b/python/lsst/pipe/base/graph/_implDetails.py
@@ -23,9 +23,8 @@ from collections import defaultdict
 
 __all__ = ("_DatasetTracker", "DatasetTypeName")
 
-from dataclasses import dataclass, field
 import networkx as nx
-from typing import (DefaultDict, Generic, Optional, Set, TypeVar, Generator, Tuple, NewType,)
+from typing import (DefaultDict, Generic, Optional, Set, TypeVar, NewType, Dict)
 
 from lsst.daf.butler import DatasetRef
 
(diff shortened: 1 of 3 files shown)


python/lsst/pipe/base/graph/_implDetails.py
```diff
@@ -40,58 +39,61 @@
...
         graph = nx.DiGraph()
-        graph.add_edges_from(self._datasetDictToEdgeIterator())
-        if None in graph.nodes():
-            graph.remove_node(None)
+        for entry in self._producers.keys() | self._consumers.keys():
+            producer = self.getProducer(entry)
+            consumers = self.getConsumers(entry)
+            # This block is for tasks that consume existing inputs
+            if producer is None and consumers:
+                for c in consumers:
```
erykoff, line 83: Can you make this `for consumer in consumers`?

python/lsst/pipe/base/graph/_implDetails.py
```diff
@@ -40,58 +39,61 @@
 _U = TypeVar("_U", TaskDef, QuantumNode)
 
 
-@dataclass
-class _DatasetTrackerElement(Generic[_U]):
-    inputs: Set[_U] = field(default_factory=set)
-    output: Optional[_U] = None
-
-
 class _DatasetTracker(Generic[_T, _U]):
```
erykoff, line 42: The class and the methods are missing docstrings.  I guess because this is internal that might be okay, but I think it'll make it clearer when you look at this code in 6 months.

python/lsst/pipe/base/graph/_implDetails.py
```diff
@@ -40,58 +39,61 @@
...
-            # be removed later
-            inputs = entry.inputs or (None,)
-            for inpt in inputs:
-                yield (entry.output, inpt)
-
     def keys(self) -> Set[_T]:
-        return set(self._container.keys())
+        return self._producers.keys() | self._consumers.keys()
+
+    def __contains__(self, value: _T) -> bool:
```
erykoff, line 98: I think `__contains__` is looking for a `key` not a `value` (which is what is being checked).  So this is just a comment on the variable name.

erykoff (approved):
Looks good, I have some small comments on variable names and docstrings.

python/lsst/pipe/base/graph/_implDetails.py
natelust: As of now I still remember this from more than 6 months ago, but you are right more documentation is better, especially if it is not me that looks at this next time.

erykoff (approved):

PR #199, opened 2021-08-24
DM-31507 Modify data-structures to aid ref counting

natelust:
It is desirable that variables get cleaned up when they go out of
scope, meaning that their ref count must go to zero. Some objects
kept closures or circular references which means they must be
cleaned by the garbage collector which can happen at unpredictable
times. Make these objects able to be dropped with reference counting.

## Checklist

- [ ] ran Jenkins


python/lsst/pipe/base/butlerQuantumContext.py
```diff
@@ -79,30 +77,32 @@ def __init__(self, butler: Butler, quantum: Quantum):
         for refs in quantum.outputs.values():
             for ref in refs:
                 self.allOutputs.add((ref.datasetType, ref.dataId))
+        self._butler_ = butler
```
andy-slac, line 80: Cannot you do `self.__butler` instead and avoid `__getattribute__`?

python/lsst/pipe/base/task.py
```diff
@@ -170,13 +174,17 @@ def __init__(self, config=None, name=None, parentTask=None, log=None):
             self._display = lsstDebug.Info(self.__module__).display
         else:
             self._display = None
-        self._taskDict[self._fullName] = self
+        self._taskDict[self._fullName] = weakref.ref(self)
```
andy-slac, line 177: Would it be better to use `WeakValueDictionary` for `self._taskDict`?

andy-slac (approved):
Looks good.

PR #204, opened 2021-09-26
DM-31769 Control graph creation from dataset existence

natelust:
This change allows the graph builder to take a string with specifies
how dataset existence will be used to control the graph query.

## Checklist

- [ ] ran Jenkins


python/lsst/pipe/base/graph/_implDetails.py
```diff
@@ -181,3 +208,77 @@ def __contains__(self, key: _T) -> bool:
...
+        nodes = set(datasetRefDict.getConsumers(ref))
+        for node in nodes:
+            # This node will never be associated with this ref
+            datasetRefDict.removeConsumer(ref, node)
+            if node in prunes:
+                continue
+            # find the connection corresponding to the input ref
+            connectionRefs = node.quantum.inputs.get(ref.datasetType)
+            # Ref must be an initInput ref and we want to ignore those
+            if connectionRefs is None:
```
TallJimbo, line 230: How about:
```
if connectionRefs is None:
    raise RuntimeError(f"Cannot prune on non-Input dataset type {ref.datasetType.name}")
    
```
to be a little more explicit and helpful to users?  (I know we talked about doing nothing here, but it seems just as easy to tell users they screwed up, which is better.)

python/lsst/pipe/base/graph/_implDetails.py
```diff
@@ -181,3 +208,77 @@ def __contains__(self, key: _T) -> bool:
...
+                continue
+            tmpRefs = set(connectionRefs).difference((ref,))
+            tmpConnections = NamedKeyDict(node.quantum.inputs.items())
+            tmpConnections[ref.datasetType] = list(tmpRefs)
+            helper = AdjustQuantumHelper(inputs=tmpConnections, outputs=node.quantum.outputs)
+            # assert to make the type checker happy, it should not actually be
+            # possible to not have dataId set to None at this point
```
TallJimbo, line 237: This is exactly the kind of string to put in the `assert` itself.

python/lsst/pipe/base/graph/_implDetails.py
```diff
@@ -181,3 +208,77 @@ def __contains__(self, key: _T) -> bool:
...
+            # If it cant because NoWorkFound is raise, that means a connection
+            # is no longer present, and the node should be removed from the
+            # graph
+            try:
+                helper.adjust_in_place(node.taskDef.connections, node.taskDef.label, node.quantum.dataId)
+                newQuantum = Quantum(taskName=node.quantum.taskName, taskClass=node.quantum.taskClass,
+                                     dataId=node.quantum.dataId, initInputs=node.quantum.initInputs,
+                                     inputs=helper.inputs, outputs=helper.outputs)
+                # If the inputs or outputs were adjusted than the graph builder
+                # thought, dissassociate node from those refs, and if they are
```
TallJimbo, line 251: Missing word or something in this sentence.

python/lsst/pipe/base/graph/_implDetails.py
```diff
@@ -181,3 +208,77 @@ def __contains__(self, key: _T) -> bool:
...
+                # output refs, prune them from downstream tasks
+                for condition, existingMapping, newMapping, remover in (
+                        (helper.inputs_adjusted, node.quantum.inputs, helper.inputs,
+                         datasetRefDict.removeConsumer),
+                        (helper.outputs_adjusted, node.quantum.outputs, helper.outputs,
+                         datasetRefDict.removeProducer)):
```
TallJimbo, line 257: I bet this would overall be a lot easier to read if you instead modified `AdjustQuantumHelper` to remember the datasets that were removed directly (it already has a similar loop to compute its `bool` flags; it just throws what you want away).

Fine to defer this to another ticket if you've already done a lot of one-off testing that you don't think you have time to repeat.

python/lsst/pipe/base/graph/_implDetails.py
```diff
@@ -181,3 +208,77 @@ def __contains__(self, key: _T) -> bool:
...
+                            notNeeded |= set(existingMapping[key]) - set(newMapping[key])
+                        if notNeeded:
+                            for ref in notNeeded:
+                                remover(ref, node)
+                            if remover is datasetRefDict.removeProducer:
+                                _pruner(datasetRefDict, notNeeded, prunes=prunes)
+                object.__setattr__(node, 'quantum', newQuantum)
+
+            except NoWorkFound:
+                # This will throw if the length is less than the minimum number
```
TallJimbo, line 273: Might be simpler to treat this the same as `helper.outputs_adjusted == node.quantum.outputs` (i.e. all outputs have gone away).  I think they're the same.  Might be another thing `AdjustQuantumHelper` should be doing, at least optionally.

python/lsst/pipe/base/graph/graph.py
```diff
@@ -170,11 +180,44 @@ def _buildGraphs(self,
...
+            prunes = set()
+            _pruner(self._datasetRefDict, pruneRefs, prunes=prunes)
+
+            # recreate the taskToQuantumNode dict removing nodes that have been
+            # pruned. Keep track of task defs that now have no QuantumNodes
+            emptyTasks: Set[TaskDef] = set()
+            newTaskToQuantumNode: DefaultDict[TaskDef, Set[QuantumNode]] = defaultdict(set)
+            # accumulate all types
+            types_ = set()
+            culprates = set()
```
TallJimbo, line 194:
```suggestion
            culprits = set()
```

python/lsst/pipe/base/graphBuilder.py
```diff
@@ -508,6 +535,12 @@ def connectDataIds(self, registry, collections, userQuery, externalDataId):
             results, just as if these constraints had been included via ``AND``
             in ``userQuery``.  This includes (at least) any instrument named
             in the pipeline definition.
+        datasetQueryQconstraint : `str`
+            Constrain construction of a quantum graph by pre-existance of
```
TallJimbo, line 539:
```suggestion
            Constrain construction of a quantum graph by pre-existence of
```

python/lsst/pipe/base/graphBuilder.py
```diff
@@ -508,6 +535,12 @@ def connectDataIds(self, registry, collections, userQuery, externalDataId):
             results, just as if these constraints had been included via ``AND``
             in ``userQuery``.  This includes (at least) any instrument named
             in the pipeline definition.
+        datasetQueryQconstraint : `str`
```
TallJimbo, line 538: How hard would it be to make this accept a list of `str`, and do the string splitting in `ctrl_mpexec`, with enum sentinals or something for `all` and `off`?  (Oh, if we only had stateful enums...)

python/lsst/pipe/base/graphBuilder.py
```diff
@@ -586,7 +631,7 @@ def connectDataIds(self, registry, collections, userQuery, externalDataId):
             yield commonDataIds
 
     def resolveDatasetRefs(self, registry, collections, run, commonDataIds, *, skipExistingIn=None,
-                           clobberOutputs=True):
+                           clobberOutputs=True, constrainedByDAllatasets: bool = True):
```
TallJimbo, line 634:
```suggestion
                           clobberOutputs=True, constrainedByAllDatasets: bool = True):
```

python/lsst/pipe/base/graph/_implDetails.py
```diff
@@ -95,6 +104,15 @@ def addConsumer(self, key: _T, value: _U):
         if self._createInverse:
             self._itemsDict[value].add(key)
 
+    def removeConsumer(self, key: _T, value: _U):
+        self._consumers[key].discard(value)
+        if self._createInverse:
+            self._itemsDict[value].discard(key)
+
+    def removeBoth(self, key: _T, value: _U):
```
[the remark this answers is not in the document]
natelust: It was in one iteration, but not now, but I figured it was low bar to keep it in, and might be useful at some point, so I left it in

python/lsst/pipe/base/graph/_implDetails.py
```diff
@@ -81,6 +85,11 @@ def addProducer(self, key: _T, value: _U):
         if self._createInverse:
             self._itemsDict[value].add(key)
 
+    def removeProducer(self, key: _T, value: _U):
```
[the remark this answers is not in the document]
natelust: This is one of the things I added while you were working, I think it explains things to your satisfaction, but might make most sense in the context of understanding the container, which is generic, so its hard to be too specific.. I like the naming because it works well with existing add and remove/consumer.

python/lsst/pipe/base/graph/_implDetails.py
natelust: Indeed I have, I think this can stand until another ticket or some of the future reworks. This really should not be triggered right now, and even when it does it will be really rare, so no rush.

python/lsst/pipe/base/graph/_implDetails.py
natelust: I am not quite sure what you are trying to say here. im not sure the helper will ever do anything, because an exception was thrown

python/lsst/pipe/base/graphBuilder.py
natelust: I figured the work had to happen somewhere and putting the logic here centralized it from potentially other places potentially doing it repeatedly, but I dont have a really strong feeling, I don't like it anywhere, and really I dont have the kind of data structures I wanted to use anyway.... If I have time I will move it if you want

python/lsst/pipe/base/graph/_implDetails.py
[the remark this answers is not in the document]
TallJimbo: I would still like a clearer answer to the bulleted question; right now it reads like it fully removes (e.g.) the `QuantumNode` that produces a `DatasetRef`, which can't be right since it doesn't automatically remove that node's input-dataset links.

python/lsst/pipe/base/graph/_implDetails.py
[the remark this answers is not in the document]
TallJimbo: I'd prefer to see it removed.  It's trivial to add back in, but nearly impossible for anyone else later to recognize as removable.

python/lsst/pipe/base/graph/_implDetails.py
TallJimbo: Right; I was just thinking of adding a `try/except` block inside `AdjustQuantumHelper`.  But as per other comment thread, that's not going to happen on this ticket, so I'm just resolving this comment.

python/lsst/pipe/base/graphBuilder.py
TallJimbo: I think the logic that converts from a single `str` is really just for command-line interfaces, and is fine to leave in ctrl_mpexec (or maybe the CLI utilities in daf_butler); any Python interface (even a JSON interface someday) should really be using something more structured.

But I could also easily believe that adding something that works to the CLI utilities is actually a lot harder, if only because it requires loading Click stuff into brain-memory from cold storage, so not a priority.

TallJimbo (approved):
Finished my second pass; I've replied on a few very points where I'd still like to see some very minor changes, but happy to marked Approved and Reviewed now to get out of your way.  I think I've resolved most of the comments, or at least the ones GitHub is still showing me.

python/lsst/pipe/base/graph/_implDetails.py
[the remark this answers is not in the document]
natelust: I will need to think about how to reword this, because it does exactly what is says, it removes a value from being marked as a producer of the associated key. Same with the other method. I will try a different wording.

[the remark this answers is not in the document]
TallJimbo: Examples for each of the generic terms would go a long ways in making it more understandable, e.g.
```
Remove a value (e.g. QuantumNode or Task) from being considered
a producer of the corresponding key (e.g. DatasetRef or DatasetType).
```

python/lsst/pipe/base/graph/_implDetails.py
```diff
@@ -181,3 +233,119 @@ def __contains__(self, key: _T) -> bool:
             Boolean of the presence of the supplied key
         """
         return key in self._producers or key in self._consumers
+
+
+def _pruner(datasetRefDict: _DatasetTracker[DatasetRef, QuantumNode], refsToRemove: Iterable[DatasetRef], *,
+            alreadyPruned: Optional[Set[QuantumNode]] = None):
+    r"""Prunes supplied dataset refs out of datasetRefDict container, recursing
```
TallJimbo, line 240:
```suggestion
    r"""Prune supplied dataset refs out of datasetRefDict container, recursing
```
(docstrings should be in imperative voice)

python/lsst/pipe/base/graph/_implDetails.py
```diff
@@ -181,3 +233,119 @@ def __contains__(self, key: _T) -> bool:
...
+            tmpConnections = NamedKeyDict(node.quantum.inputs.items())
+            tmpConnections[toRemove.datasetType] = list(tmpRefs)
+            helper = AdjustQuantumHelper(inputs=tmpConnections, outputs=node.quantum.outputs)
+            msg = ("assert to make the type checker happy, it should not actually be possible to not "
+                   "have dataId set to None at this point")
+            assert node.quantum.dataId is not None, msg
+            # Try to adjust the quantum with the reduced refs to make sure the
+            # node will still satisfy all its conditions.
+            #
+            # If it cant because NoWorkFound is raise, that means a connection
```
TallJimbo, line 300:
```suggestion
            # If it can't because NoWorkFound is raised, that means a connection
```

python/lsst/pipe/base/graph/_implDetails.py
```diff
@@ -181,3 +233,119 @@ def __contains__(self, key: _T) -> bool:
...
+            else:
+                toRemove = ref
+
+            tmpRefs = set(connectionRefs).difference((toRemove,))
+            tmpConnections = NamedKeyDict(node.quantum.inputs.items())
+            tmpConnections[toRemove.datasetType] = list(tmpRefs)
+            helper = AdjustQuantumHelper(inputs=tmpConnections, outputs=node.quantum.outputs)
+            msg = ("assert to make the type checker happy, it should not actually be possible to not "
+                   "have dataId set to None at this point")
+            assert node.quantum.dataId is not None, msg
```
TallJimbo, line 296: Assigning the assert message to a variable outside the assert statement seems a bit against the spirit of asserts and being able to "compile them out".

python/lsst/pipe/base/graph/_implDetails.py
```diff
@@ -181,3 +233,119 @@ def __contains__(self, key: _T) -> bool:
...
+                        tmpRef = tmpRef.makeCompositeRef()
+                    datasetRefDict.removeConsumer(tmpRef, node)
+                alreadyPruned.add(node)
+                # prune all outputs produced by this node
+                # mark that none of these will be produced
+                forwardPrunes = set()
+                for forwardRef in chain.from_iterable(node.quantum.outputs.values()):
+                    datasetRefDict.removeProducer(forwardRef, node)
+                    forwardPrunes.add(forwardRef)
+                _pruner(datasetRefDict, forwardPrunes, alreadyPruned=alreadyPruned)
```
TallJimbo, line 351: I'm a little wary of recursing (as well as the logic earlier) inside the `except` block, as it seems like a way to get ugly "while handling an exception, another exception was raised" tracebacks when something goes wrong.  Could we put a `return` at the end of `try` and just de-indent most of the `except` block?

python/lsst/pipe/base/graphBuilder.py
```diff
@@ -530,12 +564,27 @@ def connectDataIds(self, registry, collections, userQuery, externalDataId):
...
-                                   dataId=externalDataId,
-                                   ).materialize() as commonDataIds:
+        queryArgs = {'dimensions': self.dimensions, 'where': userQuery, 'dataId': externalDataId}
+        if datasetQueryConstraint == DatasetQueryConstraintVariant.ALL:
+            _LOG.debug("Constraining graph query using all datasets in pipeline.")
+            queryArgs['datasets'] = list(self.inputs)
+            queryArgs['collections'] = collections
+        elif datasetQueryConstraint == DatasetQueryConstraintVariant.OFF:
+            _LOG.debug("Not using dataset existance to constrain query.")
+        elif datasetQueryConstraint == DatasetQueryConstraintVariant.LIST:
```
TallJimbo, line 574: Feels weird to me that this variant instance satisfies an equality check against what (seems to be) its type. 

Feel free to defer fixing this until later, as I know the variant is going to change later anyway.  But it's something I think those fixes should address.

python/lsst/pipe/base/pipeline.py
```diff
@@ -709,13 +714,20 @@ def fromTaskDef(
         include_configs : `bool`, optional
             If `True` (default) include config dataset types as
             ``initOutputs``.
+        usePlaceHolder : bool
+            If true a placeholder storageclass will be used for any
+            component DatasetType who's parent storage class cannot be
```
TallJimbo, line 719:
```suggestion
            component DatasetType whose parent storage class cannot be
```

python/lsst/pipe/base/pipeline.py
```diff
@@ -725,6 +737,12 @@ def makeDatasetTypesSet(connectionType: str, freeze: bool = True) -> NamedValueS
                 to an attribute of type `list` on the connection class instance
             freeze : `bool`, optional
                 If `True`, call `NamedValueSet.freeze` on the object returned.
+            usePlaceHolder : bool
+                If true a placeholder storageclass will be used for any
+                component DatasetType who's parent storage class cannot be
```
TallJimbo, line 742:
```suggestion
                component DatasetType whose parent storage class cannot be
```

python/lsst/pipe/base/tests/simpleQGraph.py
```diff
@@ -343,6 +349,10 @@ def makeSimpleQGraph(nQuanta=5, pipeline=None, butler=None, root=None, run="test
         Dictionary whose keys are collection names and values are lists of
         dataset type names. By default a single dataset of type "add_dataset0"
         is added to a ``butler.run`` collection.
+    datasetQueryQconstraint : `DatasetQueryConstraintVariant`
+        The query constraint variant that should be used to constraint the
```
TallJimbo, line 353:
```suggestion
        The query constraint variant that should be used to constrain the
```

TallJimbo (approved):
Looks good; only remaining comments are minor style/doc things that you can address without another look from me.

In the long term, it occurs to me that we might want to make all DatasetRefs in the graph parent refs, and leave it to ButlerQuantumContext and QuantizedConnection to pull out components at the same time they convert multiple=False connections from lists to scalars.  Just wanted to mention that to someone else now so it's less likely to be forgotten.

python/lsst/pipe/base/pipeline.py
```diff
@@ -768,9 +794,15 @@ def makeDatasetTypesSet(connectionType: str, freeze: bool = True) -> NamedValueS
...
-                        compositeName, componentName = DatasetType.splitDatasetTypeName(c.name)
-                        parentStorageClass = DatasetType.PlaceholderParentStorageClass \
-                            if componentName else None
+                        _, componentName = DatasetType.splitDatasetTypeName(c.name)
+                        if componentName:
+                            if not usePlaceHolder:
+                                raise ValueError("Component parent class cannot be determined, and "
+                                                 "placeholder was not supplied")
+                            else:
+                                parentStorageClass = PlaceHolder
```
timj, line 803: I'm a bit confused. I know I've been out the loop but why have we replaced an explicit placeholder definition in daf butler by a completely generic place holder that now has no special meaning at all?

python/lsst/pipe/base/pipeline.py
```diff
@@ -997,6 +1030,23 @@ def fromPipeline(
...
+                    # component ref, these will need to have their parent
+                    # storage class set.
+                    toUpdate = []
+                    for other in byTask.values():
+                        # only need to look at inputs, because that is the only
+                        # valid place intermediates can be
+                        if dsType in other.inputs:
+                            toUpdate.append([x for x in other.inputs if x == dsType][0])
+                    # update the parent storage class for thie composit ref
+                    dsType.replaceParentStorageClass(composite.storageClass)
```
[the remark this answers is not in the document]
timj: I thought that was because you said you didn't need a placeholder at all because the rewrite of the graph builder code meant you always knew the parent ahead of time. If you are still using a placeholder I feel we should use the actual placeholder that `DatasetType` knows about so you can have the protection of that code only allowing a true placeholder to be replaced.

python/lsst/pipe/base/pipeline.py
natelust: This is exactly what you asked me to do in a weekly meeting, removing this special value from daf_butler. I have no strong opinion on it, but you previously indicated you preferred it gone

python/lsst/pipe/base/pipeline.py
```diff
@@ -714,20 +717,28 @@ def fromTaskDef(
...
             ``initOutputs``.
-        usePlaceHolder : bool
-            If true a placeholder storageclass will be used for any
-            component DatasetType who's parent storage class cannot be
-            determined.  If a StorageClass cannot be determined and this
-            parameter set to False, this method will raise a ValueError.
-            Defaults to False.
+        storage_class_mapping : `Mapping` of `str` to `StorageClass`, optional
+            If a taskdef contains a component dataset type that is unknown
+            to the registry, it's parent StorageClass will be looked up in this
```
TallJimbo, line 728: `it's` -> `its`

python/lsst/pipe/base/pipeline.py
```diff
@@ -714,20 +717,28 @@ def fromTaskDef(
...
             The dataset types used by this task.
+
+        Raises
+        ------
+        ValueError
+            Raised if dataset type connection definition differs from
+            registry definition.
+            Raised if component parent StorageClass could not be determined
+            and storage_class_mapping does not contain the composite type, or
+            is set to None.
```
TallJimbo, line 745: Maybe making this a `LookupError` instead?

TallJimbo (approved):

PR #206, opened 2021-10-05
DM-31966: make QuantumGraph-building and pipeline-analysis interfaces slightly more flexible

TallJimbo:
## Checklist

- [ ] ran Jenkins

### DIFF
--- a/python/lsst/pipe/base/graphBuilder.py
+++ b/python/lsst/pipe/base/graphBuilder.py
@@ -368,7 +368,7 @@ class _PipelineScaffolding:
 
     Parameters
     ----------
-    pipeline : `Pipeline`
+    pipeline : `Pipeline` or `Iterable` [ `TaskDef` ]
         Sequence of tasks from which a graph is to be constructed.  Must
         have nested task classes already imported.
     universe : `DimensionUniverse`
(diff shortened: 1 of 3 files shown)


python/lsst/pipe/base/graphBuilder.py
```diff
@@ -926,9 +926,12 @@ def makeGraph(self, pipeline, collections, run, userQuery,
...
             raise ValueError("Pipeline requires input datasets but no input collections provided.")
-        instrument = pipeline.getInstrument()
-        if isinstance(instrument, str):
-            instrument = doImport(instrument)
+        instrument = None
+        if isinstance(pipeline, Pipeline):
+            instrument = pipeline.getInstrument()
+            if isinstance(instrument, str):
+                instrument = doImport(instrument)
+            pipeline = list(pipeline.toExpandedPipeline())
```
erykoff, line 934: Is it a problem that the `instrument` is always `None` if we pass an iterable of `TaskDef` instead of a `pipeline`?  Or does creating those taskdefs already do some magic with the instrument?

erykoff (approved):
Looks good, though I do have a question about the new code path and instruments.

PR #211, opened 2021-11-23
DM-32682: Create specialist TaskMetadata 

timj:
This replaces PropertySet and PropertyList but the default is still to use PropertyX.

## Checklist

- [x] ran Jenkins


python/lsst/pipe/base/_task_metadata.py
```diff
@@ -0,0 +1,271 @@
...
+def _isListLike(v):
+    return isinstance(v, Sequence) and not isinstance(v, str)
+
+
+class TaskMetadata(BaseModel):
+    """Dict-like object for storing task metadata.
+    Metadata can be stored at two levels: single task or task plus subtasks.
+    The later is called full metadata of a task and has a form
+
+        topLevelTaskName:subtaskName:subsubtaskName.itemName
```
ktlim, line 44: I'm not exactly sure why we ended up with this weird dual hierarchy, but I guess we can live with it.

python/lsst/pipe/base/_task_metadata.py
```diff
@@ -0,0 +1,271 @@
...
+        Parameters
+        ----------
+        name : `str`
+            Name of the metadata property.
+        value
+            Metadata property value.
+        """
+        if self.__contains__(name):
+            v = self.__getitem__(name)
+            if isinstance(v, TaskMetadata):
```
ktlim, line 71: Is there any way to get rid of the very large number of `isinstance()` calls in this class?  Perhaps maintaining separate top-level and second-level dictionaries, so an `in` check suffices?

Also maybe separate value and list storage so that can also avoid `isinstance()`?

ktlim (approved):
While this is workable, I'm concerned about the implementation of the two-level hierarchy and the scalar/list duality.

doc/changes/DM-32682.feature.rst
```diff
@@ -0,0 +1,3 @@
+`lsst.pipe.base.TaskMetadata` can now be used as metadata for `~lsst.pipe.base.Task`.
```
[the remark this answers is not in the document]
timj: This depends entirely on how you've set up butler. At the moment butler will give you back what you put in -- we don't have any code for casting the result of a butler.get to a new type. The plan was to try to see if the butler can cast a PropertySet to a TaskMetadata on get but I don't think that affects this ticket. Making butler handle older metadata is a different ticket.

python/lsst/pipe/base/_task_metadata.py
timj: I assume we can't change it because we have code that's full of it. I assume we couldn't use a `.` because that would look like hierarchy.

timj:
The scalar/list duality exists here entirely because PropertySet has scalar/list duality and all the existing code with `.add()` and `.getScalar()` still has to work -- especially during a transition period. I'm not sure how we can avoid it.

python/lsst/pipe/base/_task_metadata.py
```diff
@@ -0,0 +1,271 @@
...
+        self.__setitem__(key, item)
+
+    @deprecated(reason="Use standard del dict syntax. " + _DEPRECATION_REASON,
+                version=_DEPRECATION_VERSION, category=FutureWarning)
+    def remove(self, key):
+        self.__delitem__(key)
+
+    @staticmethod
+    def _getKeys(key):
+        if not isinstance(key, str):
```
timj, line 204: I can replace this instance check with a try/except on the split.

python/lsst/pipe/base/_task_metadata.py
timj: One motivation for this class was to try to make it as close to a Mapping as possible. Is performance an issue given what we are actually doing with this class (which is mostly calling `add()` and only having tens of keys.

If we don't care about the serialized JSON explicitly showing the scalar/list/TaskMetadata keys as distinct parts of the model then it probably will make it more efficient to do and I'll remove the `__root__` from the model.

doc/changes/DM-32682.feature.rst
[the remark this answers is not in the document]
timj: Also, since the `*_metadata` dataset type is defined as a `PropertySet` you can't even run a pipeline with `TaskMetadata` as the storage class without first changing the dataset type definition.

python/lsst/pipe/base/_task_metadata.py
```diff
@@ -0,0 +1,427 @@
...
+        """
+        if topLevelOnly:
+            warnings.warn("Use keys() instead. " + _DEPRECATION_REASON, FutureWarning)
+            return set(self.keys())
+        else:
+            names = set()
+            for k, v in self.items():
+                names.add(k)  # Always include the current level
+                if isinstance(v, TaskMetadata):
+                    names.update({k + '.' + item for item in v.names(topLevelOnly=topLevelOnly)})
```
ktlim, line 221: This is a tiny bit confusing since you know that `topLevelOnly` is `False`.

python/lsst/pipe/base/_task_metadata.py
```diff
@@ -0,0 +1,427 @@
...
+            part of a hierarchy. If `False` also returns fully-qualified
+            names from the hierarchy. Keys associated with the top
+            of a hierarchy are never returned.
+        """
+        # Currently used by the verify package.
+        paramNames = set()
+        for k, v in self.items():
+            if isinstance(v, TaskMetadata):
+                if not topLevelOnly:
+                    paramNames.update({k + "." + item for item in v.paramNames(topLevelOnly=topLevelOnly)})
```
ktlim, line 246: Same here.

python/lsst/pipe/base/_task_metadata.py
```diff
@@ -0,0 +1,427 @@
...
+
+
+class TaskMetadata(BaseModel):
+    """Dict-like object for storing task metadata.
+    Metadata can be stored at two levels: single task or task plus subtasks.
+    The later is called full metadata of a task and has a form
+
+        topLevelTaskName:subtaskName:subsubtaskName.itemName
+
+    Metadata item key of a task (`itemName` above) must not contain `.`,
```
ktlim, line 52: Hasn't this changed?

python/lsst/pipe/base/_task_metadata.py
```diff
@@ -0,0 +1,427 @@
...
+        if key0 in self.metadata:
+            # And forward request to that metadata
+            return self.metadata[key0][".".join(keys)]
+        raise KeyError(f"'{key}' not found")
+
+    def __setitem__(self, key, item):
+        keys = self._getKeys(key)
+        key0 = keys.pop(0)
+        if len(keys) == 0:
+            # Currently no type validation.
```
ktlim, line 312: Isn't this now false?

ktlim (approved):

python/lsst/pipe/base/_task_metadata.py
timj: I used the given value mostly because if I rearranged the code at some point I wouldn't need to change the value there.

python/lsst/pipe/base/_task_metadata.py
timj: Right. Before this most recent change only a single `.` was allowed.

PR #212, opened 2021-11-24
DM-32451 Sort Pipelines in deterministic way

natelust:
When serializing to disk or string make sure the pipeline is ordered
in a deterministic way.

## Checklist

- [ ] ran Jenkins

### DIFF
--- a/tests/test_pipeline.py
+++ b/tests/test_pipeline.py
@@ -22,21 +22,26 @@
 """Simple unit test for Pipeline.
 """
 
+import tempfile
 import textwrap
 import unittest
 
 import lsst.pex.config as pexConfig
 from lsst.pipe.base import (Struct, PipelineTask, PipelineTaskConfig, Pipeline, TaskDef,
-                            PipelineTaskConnections, PipelineDatasetTypes)
+                            PipelineTaskConnections, PipelineDatasetTypes, connectionTypes)
 from lsst.pipe.base.tests.simpleQGraph import makeSimplePipeline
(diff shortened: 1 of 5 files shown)


tests/test_pipeline.py
```diff
@@ -150,12 +155,33 @@ def testParameters(self):
 
     def testSerialization(self):
         pipeline = Pipeline("test")
-        pipeline.addTask(AddTask, "add")
         pipeline.addTask(MultTask, "mult")
+        pipeline.addTask(AddTask, "add")
+
+        # verify the keys are in insertion order
+        self.assertEqual(list(pipeline._pipelineIR.tasks.keys()), ['mult', 'add'])
```
TallJimbo, line 162: I assume you're using internal APIs in these tests because the public way of getting the keys is going to sort them automatically?  If so, it might be better to remove these "before" checks and just verify that the public order is topological, since that should be the only thing the interface actually guarantees.

TallJimbo (approved):
This still isn't fully-deterministic sorting, right, because the tiebreaker for topological ties is the original definition order?  If so, I think that's fine for now, but we should adjust the commit message and possible add something to the `Pipeline` class docstring with that caveat.

tests/test_pipeline.py
natelust: Yeah, that is exactly why I was doing it, I wanted to definitively check that they were unsorted before, to guard against some future change where someone makes a change and suddenly the test is passing because of ordering, but where the sorting is broken.

PR #223, opened 2022-01-07
DM-33155: Change how task metadata storage class is chosen

timj:
* Add `TaskMetadata.to_dict()` method (this is now used by the `PropertySet.from_mapping()` method and triggered by the butler if type conversion is needed).
* Use the existing metadata storage class definition if one already exists.
* Switch `Task` to use `TaskMetadata` rather than `PropertySet`

Requires lsst/daf_butler#626 and lsst/daf_base#72

## Checklist

- [x] ran Jenkins


TallJimbo (approved):

timj:
@TallJimbo I realized that there is no need to continue to support the PropertySet switch at the top of the file so I removed that code.

PR #222, opened 2021-12-21
DM-33034: tools for building Sphinx docs from Pipelines

TallJimbo:
## Checklist

- [ ] ran Jenkins
- [ ] added a release note for user-visible changes to `doc/changes`

Note that this PR is against a temporary branch, because merging this ticket (but not its review) is blocked by DM-33027.  After that lands I'll rebase.

PR #224, opened 2022-01-11
DM-33204: Use ResourcePath instead of ButlerURI throughout.

timj:
Also replaces a few Union[str, ButlerURI] and similar with the new
ResourcePathExpression type alias.

This is essentially a cherry pick of commit b014171b from #222.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/pipe/base/pipelineIR.py
+++ b/python/lsst/pipe/base/pipelineIR.py
@@ -33,7 +33,7 @@ from typing import Any, Dict, Generator, List, Mapping, MutableMapping, Optional
 
 import yaml
 from deprecated.sphinx import deprecated
-from lsst.daf.butler import ButlerURI
+from lsst.resources import ResourcePath, ResourcePathExpression
 
 
 class KeepInstrument:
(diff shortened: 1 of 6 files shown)


python/lsst/pipe/base/pipelineIR.py
```diff
@@ -902,17 +897,20 @@ def to_file(self, filename: str):
...
 
         Parameters
         ----------
-        uri: `str` or `ButlerURI`
+        uri: convertible to `ResourcePath`
             Location of document to write a `PipelineIR` object.
         """
-        butlerUri = ButlerURI(uri)
-        butlerUri.write(yaml.dump(self.to_primitives(), sort_keys=False).encode())
+        with ResourcePath(uri).open("w") as buffer:
```
timj, line 912: @TallJimbo I was torn whether to keep the I/O code as it was or adopt the new form that you had written on the branch (also the `from_uri` method above).

TallJimbo: No opinion from me.

python/lsst/pipe/base/configOverrides.py
```diff
@@ -234,7 +236,8 @@ def applyTo(self, config):
 
         for otype, override in self._overrides:
             if otype is OverrideTypes.File:
-                config.load(override)
+                with override.open("r") as buffer:
```
timj, line 239: Now that `ResourcePath` is a standalone package we should consider using it in `pex_config` directly.

PR #225, opened 2022-01-13
DM-33260: Add support for GitHub action to run tests

timj:
Gen2 will not be testable.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/pipe/base/graph/graph.py
+++ b/python/lsst/pipe/base/graph/graph.py
@@ -54,6 +54,7 @@ from typing import (
 import networkx as nx
 from lsst.daf.butler import DatasetRef, DatasetType, DimensionRecordsAccumulator, DimensionUniverse, Quantum
 from lsst.resources import ResourcePath, ResourcePathExpression
+from lsst.utils.introspection import get_full_type_name
 from networkx.drawing.nx_agraph import write_dot
 
 from ..connections import iterConnections
@@ -895,7 +896,7 @@ class QuantumGraph:
             taskDescription = {}
             # save the fully qualified name, as TaskDef not not require this,
             # but by doing so can save space and is easier to transport
-            taskDescription["taskName"] = f"{taskDef.taskClass.__module__}.{taskDef.taskClass.__qualname__}"
+            taskDescription["taskName"] = get_full_type_name(taskDef.taskClass)
             # save the config as a text stream that will be un-persisted on the
             # other end
             stream = io.StringIO()
--- a/python/lsst/pipe/base/graph/graph.py
+++ b/python/lsst/pipe/base/graph/graph.py
@@ -20,8 +20,6 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 from __future__ import annotations
 
-from lsst.daf.butler.core.datasets.type import DatasetType
-
 __all__ = ("QuantumGraph", "IncompatibleGraphError")
 
 import io
@@ -54,7 +52,7 @@ from typing import (
 )
 
 import networkx as nx
-from lsst.daf.butler import DatasetRef, DimensionRecordsAccumulator, DimensionUniverse, Quantum
+from lsst.daf.butler import DatasetRef, DatasetType, DimensionRecordsAccumulator, DimensionUniverse, Quantum
 from lsst.resources import ResourcePath, ResourcePathExpression
 from networkx.drawing.nx_agraph import write_dot
 
--- a/python/lsst/pipe/base/graph/graph.py
+++ b/python/lsst/pipe/base/graph/graph.py
@@ -894,8 +894,7 @@ class QuantumGraph:
             # compressing has very little impact on saving or load time, but
             # a large impact on on disk size, so it is worth doing
             taskDescription = {}
-            # save the fully qualified name, as TaskDef not not require this,
-            # but by doing so can save space and is easier to transport
+            # save the fully qualified name.
             taskDescription["taskName"] = get_full_type_name(taskDef.taskClass)
             # save the config as a text stream that will be un-persisted on the
             # other end
--- a/python/lsst/pipe/base/tests/simpleQGraph.py
+++ b/python/lsst/pipe/base/tests/simpleQGraph.py
@@ -30,7 +30,11 @@ import logging
 import lsst.daf.butler.tests as butlerTests
 import lsst.pex.config as pexConfig
 import numpy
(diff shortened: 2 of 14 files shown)


python/lsst/pipe/base/graph/graph.py
```diff
@@ -897,7 +896,7 @@ def _buildSaveObject(self, returnHeader: bool = False) -> Union[bytearray, Tuple
             taskDescription = {}
             # save the fully qualified name, as TaskDef not not require this,
             # but by doing so can save space and is easier to transport
-            taskDescription["taskName"] = f"{taskDef.taskClass.__module__}.{taskDef.taskClass.__qualname__}"
+            taskDescription["taskName"] = get_full_type_name(taskDef.taskClass)
```
gpdf, line 899: Maybe the "TaskDef not not require" should be fixed at this time?  I'm really not sure what it's trying to say.

gpdf (commented):
Thank you very much for this!  @tgoldina and I really appreciate it - and would have been hard pressed to have figured out everything that ended up in this PR.

python/lsst/pipe/base/graph/graph.py
natelust: Does not require the fully qualified task name

TallJimbo: FWIW, on DM-32027, I've changed the docs to say that a fully-qualified task name is required, as there is code in various places that assumes that it is fully-qualified by calling `doImport` on it, and all of our pipeline definition files do use fully-qualified names.  Passing around a Python class name without the module and expecting magic to resolve it was always a bad idea, I think.

natelust: Yeah I think TaskDef should have required it, and we should make it so if you didn't do it in that ticket, because it is implicitly required

tests/test_quantumGraph.py
```diff
@@ -316,7 +307,14 @@ def testIsConnected(self):
...
         filteredNodes = [n for n in self.qGraph if n.taskDef.label != "U"]
-        subset = self.qGraph.subset((filteredNodes[0], filteredNodes[1]))
+        label0 = filteredNodes[0].taskDef.label
+        node1 = None
+        for n in filteredNodes[1:]:
+            if n.taskDef.label != label0:
+                node1 = n
+                break
+        self.assertIsNotNone(node1, f"Could not find node with label other than {label0}")
+        subset = self.qGraph.subset((filteredNodes[0], node1))
```
natelust, line 317: OK, so my comment on this is that this test should just be taken out completely.

Move `self.assertFalse(self.qGraph.isConnected)` to the beginning of the `testSubsetToConnected` test. All that is being tested in `testIsConnected` is tested in `testSubsetToConnected` anyway, so no need to hit the same methods more than once, and `subsetToConnected` is the proper way to split up a graph anyway.

python/lsst/pipe/base/graph/graph.py
```diff
@@ -54,8 +52,9 @@
 )
 
 import networkx as nx
-from lsst.daf.butler import DatasetRef, DimensionRecordsAccumulator, DimensionUniverse, Quantum
+from lsst.daf.butler import DatasetRef, DatasetType, DimensionRecordsAccumulator, DimensionUniverse, Quantum
```
natelust, line 55: Was this import broken before somehow, or implicit, I don't see anywhere else in the diff for this file where it has been added.

tests/test_dataIdContainer.py
```diff
@@ -24,15 +24,22 @@
...
 import lsst.utils.tests
 
 
+@unittest.skipIf(DataIdContainer is None, "Gen2 infrastructure is not available")
 class DataIdContainerTestCase(lsst.utils.tests.TestCase):
     def setUp(self):
-        self.container = pipeBase.DataIdContainer()
-        self.butler = unittest.mock.MagicMock(spec=lsst.daf.persistence.Butler)
+        self.container = DataIdContainer()
+        self.butler = unittest.mock.MagicMock(spec=dafPersistence.Butler)
```
natelust, line 42: Wont this also fail if dafPersistence is None?

natelust (approved):

python/lsst/pipe/base/graph/graph.py
timj: The code already used the module so was seemingly fully-qualified. I can fix the typo in the comment and look at @TallJimbo 's commit.

python/lsst/pipe/base/graph/graph.py
timj: Look a few lines above. DatasetType was being explicitly imported right at the top of the file using lsst.daf.butler.core.types.

tests/test_dataIdContainer.py
timj: yes, but the test is skipped because the gen2 code can't be imported because dafPersistence is missing. I can add an explicit `or` to the skipIf if you like.

python/lsst/pipe/base/graph/graph.py
natelust: It was qualified because it is the right thing to do, the comment was basically saying `TaskDef` does not require this for construction, but we are going to do it because it is better to.

python/lsst/pipe/base/graph/graph.py
natelust: You are right, sorry I could have sworn I didn't see that in the diff before, maybe in a different github view or something. And thanks for the fix, sometimes if I use tab complete on VSCode it will auto add the import for me, and if I don't catch it you end up with the above.

python/lsst/pipe/base/tests/simpleQGraph.py
```diff
@@ -166,8 +170,9 @@ def registerDatasetTypes(registry, pipeline):
         configDatasetType = DatasetType(
             taskDef.configDatasetName, {}, storageClass="Config", universe=registry.dimensions
         )
+        storageClass = "Packages" if Packages is not None else "StructuredDataDict"
```
timj, line 173: I had to tweak this so that the `Packages` storage class is used if the corresponding class is available. The reason is that even though I can do storage class conversions (see lsst/ctrl_mpexec#160) the `ctrl_mpexec` tests require that `butler.get()` returns `Packages` because it then uses methods to check for changes. It won't be a problem if the RFC is approved to let us migrate the `Packages` code to `utils`.

PR #227, opened 2022-01-19
DM-33260 hotfix: make lsstimport optional

tgoldina:
Hotfix for missed lsstimport in [DM-33260](https://jira.lsstcorp.org/browse/DM-33260).

## Checklist

- [ ] ran Jenkins

### DIFF
--- a/python/lsst/__init__.py
+++ b/python/lsst/__init__.py
@@ -1,5 +1,9 @@
 import pkgutil
 
-import lsstimport
+try:
+    import lsstimport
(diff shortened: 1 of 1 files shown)


timj (approved):
Looks good. I guess this worked before because the `lsst/__init__.py` can be picked up from `daf_butler` or `utils` but depending on import order you can hit the `pipe_base` one instead.

tgoldina:
Yes, I was installing pipe_base last.

PR #228, opened 2022-01-20
DM-33303: Allow pipelines to have dataset types with differing storage classes

timj:
If a pipeline uses a dataset type with a storage class that differs from the one in use in the butler registry, this can be okay if the storage classes are convertible.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/pipe/base/pipeline.py
+++ b/python/lsst/pipe/base/pipeline.py
@@ -901,7 +901,7 @@ class TaskDatasetTypes:
                             # the registry definition.
                             if not is_input:
                                 datasetType = registryDatasetType
-                            _LOG.warning(
+                            _LOG.debug(
                                 "Dataset types differ (task %s != registry %s) but are compatible"
                                 " for %s in %s.",
                                 datasetType,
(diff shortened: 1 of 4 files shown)


python/lsst/pipe/base/pipeline.py
```diff
@@ -879,27 +886,51 @@ def makeDatasetTypesSet(connectionType: str, freeze: bool = True) -> NamedValueS
...
+                            # Has to be able to be converted to expect type
+                            # on put.
+                            is_compatible = registryDatasetType.is_compatible_with(datasetType)
+                        if is_compatible:
+                            # For inputs we want the pipeline to use the
+                            # pipeline definition, for outputs it should use
+                            # the registry definition.
+                            if not is_input:
+                                datasetType = registryDatasetType
+                            _LOG.warning(
```
andy-slac, line 904: Should it be `debug`?

andy-slac (approved):
Looks good, one minor comment.

python/lsst/pipe/base/pipeline.py
timj: Yes. I meant to do that -- I only had them warning to make it easy to spot them in debugging.

PR #229, opened 2022-01-22
DM-3334: Use transfer_from to transfer datasets to execution butler

timj:
Transferring the datasets using export/import is slow
because each record has to be handled separately on re-ingest.
This requires file system checks.

Use transfer_from the datastore side is handled by bulk
queries of the file_datastore_records tables and the
files are assumed to exist and so are not checked.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/doc/changes/DM-33345.misc.rst
+++ b/doc/changes/DM-33345.misc.rst
@@ -1,0 +1,1 @@
+Execution butler creation time has been reduced significantly by avoiding unnecessary checks for existence of files in the datastore.
--- a/python/lsst/pipe/base/executionButlerBuilder.py
+++ b/python/lsst/pipe/base/executionButlerBuilder.py
@@ -78,6 +78,10 @@ def _accumulate(
                 # be inserted into the new registry
                 for ref in refs:
                     if ref.id is not None:
+                        # If this is a component we want the composite to be
+                        # exported.
+                        if ref.isComponent():
+                            ref = ref.makeCompositeRef()
                         exports.add(ref)
                     else:
                         if ref.isComponent():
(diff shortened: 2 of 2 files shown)


timj:
@TallJimbo it's entirely possible that this minor change will fix the problem. It's now doing a direct transfer of records from one datastore to the other and not involving the file system at all. `pipelines_check` works, with my only worry being that the execution registry that gets created is now 21 lines shorter (when doing `sqlite3 gen3.sqlite3 .dump`). I'm not entirely sure why it has shrunk and because the tags tables get renumbered between old code and new it's hard to compare the dumps (the datastore records do match).

TallJimbo:
Looking at what changed in the code, I would guess it's TAGGED and CALIBRATION  membership that's not being saved now.  But I'm not sure how `pipelines_check` is still working if that's the case.

timj:
Ah yes. So I need to do the queryDatasetAssociations stuff that is triggered previously in the export with each dataset ref that is to be exported. I could add the logic to `Butler.transfer_from()` as a boolean parameter and copy some of the code over from the export/import logic -- it doesn't seem wildly ridiculous to do that since it's directly related to the DatasetRefs that are being transferred.

TallJimbo:
:+1:, but I think you'll want an `Iterable[str]` parameter that lists the `TAGGED` and `CALIBRATION` collections to export, instead of a boolean, unless I've misinterpreted where you planned to do this.

timj:
Maybe I misread how the exporter code works. It looked at first glance that it finds all the associations attached to the datasets that are being exported. Does execution butler need the associations if the graph builder has already determined which calibrations and tagged datasets would be relevant?

TallJimbo:
The exporter only exports TAGGED and CALIBRATION memberships when the dataset is passed to `saveDatasets` _and_ the collection is passed to `saveCollection`.  But that's a very good point - the searches in collections are all done in `QuantumGraph` generation, and execution only uses `getDirect`, so we don't actually need the collection memberships anyway.  Sorry for the distraction!

TallJimbo (approved):

PR #235, opened 2022-02-19
DM-32250: add mypy configuration and typing fixes

TallJimbo:
## Checklist

- [x] ran Jenkins


mypy.ini
```diff
@@ -0,0 +1,82 @@
...
+
+[mypy-lsst.*]
+ignore_missing_imports = True
+ignore_errors = True
+
+# Do expect to import type annotations from LSST middleware dependencies that
+# should have them, but don't re-check them as they have their own mypy.ini
+# files with their own exceptions.
+
+[mypy-lsst.utils.*]
```
timj, line 21: How do we know it's importing the types from utils and resources? Also, daf_butler doesn't declare that it is types and doesn't set MYPYPATH in the table file.

timj: Ha. I see that's in lsst/daf_butler#647 ...

python/lsst/pipe/base/graph/_versionDeserializers.py
```diff
@@ -513,8 +526,8 @@ def constructGraph(
 
                 # bytes are compressed, so decompress them
                 taskDefDump = json.loads(lzma.decompress(_readBytes(start, stop)))
-                taskClass: PipelineTask = doImport(taskDefDump["taskName"])  # type: ignore
-                config: Config = taskClass.ConfigClass()  # type: ignore
+                taskClass: Type[PipelineTask] = doImport(taskDefDump["taskName"])  # type: ignore
```
timj, line 529: `doImportType` + `assert` is your friend.

python/lsst/pipe/base/pipeline.py
```diff
@@ -520,8 +530,8 @@ def addTask(self, task: Union[PipelineTask, str], label: str) -> None:
             # be defined without label which is not acceptable, use task
             # _DefaultName in that case
             if isinstance(task, str):
-                task = doImport(task)
-            label = task._DefaultName
+                task_class = doImport(task)
```
timj, line 533: `doImportType` but also, how does mypy know that `task_class._DefaultName` going to work without an assert?

python/lsst/pipe/base/pipeline.py
```diff
@@ -654,7 +664,7 @@ def _toExpandedPipelineImpl(self, checkContracts=True) -> Iterable[TaskDef]:
     def _buildTaskDef(self, label: str) -> TaskDef:
         if (taskIR := self._pipelineIR.tasks.get(label)) is None:
             raise NameError(f"Label {label} does not appear in this pipeline")
-        taskClass = doImport(taskIR.klass)
+        taskClass: Type[PipelineTask] = doImport(taskIR.klass)  # type: ignore
```
timj, line 667: `doImportType`

python/lsst/pipe/base/pipelineIR.py
```diff
@@ -837,7 +837,7 @@ def from_string(cls, pipeline_string: str):
         reason="This has been replaced with `from_uri`. will be removed after v23",
         version="v21.0,",
         category=FutureWarning,
-    )  # type: ignore
+    )
```
timj, line 840: In theory we can now remove this since v23 is out...

python/lsst/pipe/base/task.py
```diff
@@ -191,24 +218,30 @@ def __init__(self, config=None, name=None, parentTask=None, log=None):
                     loggerName = module_root + loggerName
 
         # Get a logger (that might be a subclass of logging.Logger).
-        self.log = lsst.utils.logging.getLogger(loggerName)
-        self.config = config
+        self.log: lsst.utils.logging.LsstLogAdapter = lsst.utils.logging.getLogger(loggerName)
```
timj, line 221: Why can't mypy work that out from the `utils` type annotations? If one says `x = 5` it knows that `x` is an `int`.

requirements.txt
```diff
@@ -2,9 +2,9 @@ pyyaml >= 5.1
 pydantic
 numpy >= 1.17
 networkx
-git+https://github.com/lsst/daf_butler@main#egg=daf_butler
+git+https://github.com/lsst/daf_butler@tickets/DM-32250#egg=daf_butler
```
timj, line 5: Note to remove before merging.

timj (approved):
Looks great. Biggest comment is to use `doImportType` + `assert` to check that `doImport` is returning the right type.

mypy.ini
```diff
@@ -0,0 +1,86 @@
...
+# worth adding new annotations to them.
+[mypy-lsst.pipe.base.argumentParser.*]
+disallow_untyped_defs = False
+[mypy-lsst.pipe.base.cmdLineTask.*]
+disallow_untyped_defs = False
+[mypy-lsst.pipe.base.shims.*]
+disallow_untyped_defs = False
+[mypy-lsst.pipe.base.task_logging.*]
+disallow_untyped_defs = False
+[mypy-lsst.pipe.base.timer.*]
```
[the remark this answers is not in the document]
TallJimbo: :+1: I'll move these to another little informal section with a comment to that effect.

python/lsst/pipe/base/_task_metadata.py
```diff
@@ -149,14 +149,14 @@ def to_dict(self) -> Dict[str, Any]:
         used when a simple dictionary is needed.  Use
         `TaskMetadata.from_dict()` to convert it back.
         """
-        d = {}
+        d: Dict[str, Any] = {}
```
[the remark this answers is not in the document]
TallJimbo: I assume you mean
```
Union[TaskMetadata, str, StrictFloat, StrictInt, StrictBool, StrictStr, List[StrictFloat], List[StrictInt], List[StrictBool], List[StrictStr]
```?

Can the `str | StrictStr` within that be collapsed to just `str`?

I think I'll try this but bail out if it doesn't just work.

File: python/lsst/pipe/base/_task_metadata.py [hunk 10]
```diff
@@ -392,26 +395,30 @@ def _getKeys(key):
             raise KeyError(f"Invalid key '{key}': only string keys are allowed") from None
         return keys
 
-    def keys(self):
+    def keys(self) -> Tuple[str, ...]:
```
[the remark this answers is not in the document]
TallJimbo: Probably, but `items` is actually much more problematic because it's an iterator, not a multi-pass sized iterable.

I think you can just return `KeysView(self)`, since it seems you implement a compatible `__iter__`, and returning `ValuesView(self)` and `ItemsView(self)` should work, too, if your `__getitem__` isn't too weird.

I'm not currently planning to do that on this ticket, though.

python/lsst/pipe/base/_task_metadata.py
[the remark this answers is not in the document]
timj: Oh, that first `str` is a mistake and yes it was meant to be union of all of them -- copy and paste error. These are all the types defined in the TaskMetadata pydantic model. I'm not sure if `str`, `float` etc work in the type annotations but the model itself needs StrictStr -- pydantic didn't do the right thing if it was not StrictStr -- it would cast the numbers to strings because the order of types matters to it.

python/lsst/pipe/base/pipelineIR.py
TallJimbo: I'll save this for another ticket I already have open (the one on expanded pipelines).

python/lsst/pipe/base/task.py
TallJimbo: Ok, tried - the problem appears to be the order in which MyPy processes the annotations (I think), and the fact that there's a reference to `parentTask.log` earlier in this method that MyPy can't figure out without the explicit annotation.

python/lsst/pipe/base/_task_metadata.py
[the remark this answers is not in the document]
TallJimbo: It turns out this one's tricky because conversion from `TaskMetadata` to `dict` here is recursive, and hence the type annotation would need to be recursive as well in order to be complete.  So I'll just stick to `Any` hre.

python/lsst/pipe/base/_task_metadata.py
```diff
@@ -224,7 +227,7 @@ def getAsDouble(self, key):
         """
         return float(self.__getitem__(key))
 
-    def getScalar(self, key):
+    def getScalar(self, key: str) -> Any:
```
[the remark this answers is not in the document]
TallJimbo: Done.

natelust (approved):

python/lsst/pipe/base/graph/_loadHelpers.py
```diff
@@ -20,37 +20,50 @@
...
+    from .graph import QuantumGraph
+
+
+_T = TypeVar("_T")
 
 
 # Create a custom dict that will return the desired default if a key is missing
-class RegistryDict(UserDict):
-    def __missing__(self, key):
+class RegistryDict(dict):
```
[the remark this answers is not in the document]
TallJimbo: I don't think there are any guarantees on which methods `UserDict` methods will delegate to either, though (there may be certain behavior devs have to learned to rely on from experience, but they are not documented).  Only `Mapping` and `MutableMapping` are safe in the way you describe.

python/lsst/pipe/base/graph/_versionDeserializers.py
```diff
@@ -189,8 +202,8 @@ def unpackHeader(self, rawHeader: bytes) -> Optional[str]:
         return None
 
     def constructGraph(
-        self, nodes: set[uuid.UUID], _readBytes: Callable[[int, int], bytes], universe: DimensionUniverse
-    ):
+        self, nodes: Set[uuid.UUID], _readBytes: Callable[[int, int], bytes], universe: DimensionUniverse
```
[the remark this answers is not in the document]
TallJimbo: I did not realize that `__future__ import annotations` gave us the 3.9 behavior early; I thought this was all premature.

I'll try reverting it and see if I get complaints.

python/lsst/pipe/base/graph/_loadHelpers.py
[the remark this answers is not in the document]
natelust: Oh nevermind, what I was thinking of is still in play, the other bit was just issues with the interpreter directly: See the last comment on https://stackoverflow.com/questions/1392396/advantages-of-userdict-class . You can get into a situation were using super() inheriting directly from dict does not work, but does with UserDict (though probably not important here)

PR #237, opened 2022-03-02
DM-33853: Add verbose log message if get() is taking a long time

timj:
On IDF some tasks take 7 hours in one call to `ButlerQuantumContext.get()`

Requires lsst/utils#116

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/pipe/base/butlerQuantumContext.py
+++ b/python/lsst/pipe/base/butlerQuantumContext.py
@@ -29,6 +29,7 @@ __all__ = ("ButlerQuantumContext",)
 from typing import Any, List, Sequence, Union
 
 from lsst.daf.butler import Butler, DatasetRef, Quantum
+from lsst.utils.introspection import get_full_type_name
 
 from .connections import DeferredDatasetRef, InputQuantizedConnection, OutputQuantizedConnection
 from .struct import Struct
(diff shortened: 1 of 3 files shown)


TallJimbo (approved):

arunkannawadi:
Coming here from https://lsstc.slack.com/archives/C01FBUGM2CV/p1646338188815929?thread_ts=1646337602.472119&cid=C01FBUGM2CV

I think it makes a lot of sense to define this function, and a default value for _LOGGING_INTERVAL as a module-level variable in the `lsst.utils.logging`. I don't think defining `loggingInterval` as a config field (the way I have done so far) makes much sense, since it's not innate to the task but refers to issues/restrictions in the computing environment.

python/lsst/pipe/base/butlerQuantumContext.py
```diff
@@ -97,6 +102,28 @@ def _put(self, value: Any, ref: DatasetRef):
...
+        *args : Any
+            Parameters to be passed to the log system.
+
+        Returns
+        -------
+        next_log_time : `float`
+            Time to issue next log message.
+        """
+        if (current_time := time.time()) > next_log_time:
+            _LOG.verbose(msg, *args)
```
arunkannawadi, line 123: _If_ we are moving it to the logging module, I think it'd also make more sense to make it `_LOG.log`, with level=logging.VERBOSE being passed in as a keyword argument

arunkannawadi: While we are generalizing, I wonder if we can make it a coroutine so that the `next_log_time` value is preserved without having to pass it during calling this. Only the relevant `msg` has to be passed in; `next_log_time` will be preserved and `_LOG_INTERVAL` is just a constant defined at the module-level.

python/lsst/pipe/base/butlerQuantumContext.py
```diff
@@ -24,13 +24,17 @@
 
 __all__ = ("ButlerQuantumContext",)
 
+import time
```
arunkannawadi, line 27: You shouldn't need to import time any more.

PR #238, opened 2022-03-14
DM-33963: add TaskMetadata.get() and improve Pipeline YAML output

TallJimbo:
All these minor changes have in common is that they were useful for adding a particular new PipelineTask and pipelines for it in downstream packages.

## Checklist

- [x] ran Jenkins


tests/test_taskmetadata.py
```diff
@@ -43,6 +43,9 @@ def testTaskMetadata(self):
         self.assertEqual(meta.getArray("test"), [42, 55, 1, 2])
         meta["new.int"] = 30
         self.assertEqual(meta["new.int"], 30)
+        self.assertEqual(meta.get("new.int", 20), 30)
+        self.assertEqual(meta.get("not.present.at.all", 20), 20)
+        self.assertEqual(meta.get("test"), 2)
```
timj, line 48: For completeness we should probably also test:

```python
md = meta.get("new")
self.assertEqual(md["int"], 30)
```

natelust (approved):

natelust (approved):

PR #239, opened 2022-03-21
DM-34105: Move some generic code from obs_base

timj:
obs_base is going to be hard to disentangle from afw, ip_isr, and pipe_tasks. Move a subset of Instrument class to pipe_base along with support material.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/tests/test_config_formatter.py
+++ b/tests/test_config_formatter.py
@@ -1,0 +1,71 @@
+# This file is part of pipe_base.
+#
+# Developed for the LSST Data Management System.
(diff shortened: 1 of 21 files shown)


tests/test_config_formatter.py
```diff
@@ -0,0 +1,71 @@
+# This file is part of daf_butler.
```
andy-slac, line 1: daf_butler -> pipe_base

andy-slac (approved):
Looks great, few comments about docstrings/comments.

PR #240, opened 2022-03-28
DM-34155: miscellaneous fixes and improvements to Pipeline and helpers

TallJimbo:
## Checklist

- [x] ran Jenkins


tests/test_pipeline.py
```diff
@@ -25,57 +25,9 @@
...
-    ConfigClass = MultConfig
-    _DefaultName = "mult"
-
-    def run(self, val):
-        self.metadata.add("mult", self.config.multiplicand)
-        return Struct(
-            val=val * self.config.multiplicand,
-        )
+from lsst.pipe.base import Pipeline, PipelineDatasetTypes, TaskDef
+from lsst.pipe.base.tests.simpleQGraph import AddTask, makeSimplePipeline
```
natelust, line 31: I think most of this duplication extends from before this code lived in pipe_base, and was in ctrl_mpexec, and so the dependency order meant that it could not be imported.

python/lsst/pipe/base/configOverrides.py
```diff
@@ -239,7 +239,7 @@ def applyTo(self, config):
         for otype, override in self._overrides:
             if otype is OverrideTypes.File:
                 with override.open("r") as buffer:
-                    config.loadFromStream(buffer, filename=str(override))
+                    config.loadFromStream(buffer, filename=override.ospath)
```
natelust, line 242: Outside the scope of this, but we should consider implementing `__fspath__` to make ResourcePaths adhere to the `PathLike` protocol. Likely we would default that method to `.ospath`, but it would move the exception to `__fspath__` being called. This would give us the option to implement other behavior at some point in the future (like automatically caching or something) and not have to hunt for all the locations we did `.ospath`

natelust (approved):

python/lsst/pipe/base/configOverrides.py
TallJimbo: Ticketed as [DM-34419](https://jira.lsstcorp.org/browse/DM-34419).

PR #241, opened 2022-04-05
DM-31701: Enable deterministic dataset loading order in Gen3 middleware

PaulPrice:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/pipe/base/graphBuilder.py
+++ b/python/lsst/pipe/base/graphBuilder.py
@@ -397,6 +397,8 @@ class _QuantumScaffolding:
         """
         allInputs = self.inputs.unpackMultiRefs()
         allInputs.update(self.prerequisites.unpackMultiRefs())
+        for name, refs in allInputs.items():
+            allInputs[name] = sorted(refs, key=lambda rr: rr.dataId)
         # Give the task's Connections class an opportunity to remove some
         # inputs, or complain if they are unacceptable.
         # This will raise if one of the check conditions is not met, which is
@@ -406,7 +408,10 @@ class _QuantumScaffolding:
         # input behave like a regular input; adjustQuantum should only raise
         # NoWorkFound if a regular input is missing, and it shouldn't be
         # possible for us to have generated ``self`` if that's true.
-        helper = AdjustQuantumHelper(inputs=allInputs, outputs=self.outputs.unpackMultiRefs())
+        outputs = self.outputs.unpackMultiRefs()
+        for name, refs in outputs.items():
+            outputs[name] = sorted(refs, key=lambda rr: rr.dataId)
+        helper = AdjustQuantumHelper(inputs=allInputs, outputs=outputs)
         helper.adjust_in_place(self.task.taskDef.connections, self.task.taskDef.label, self.dataId)
         initInputs = self.task.initInputs.unpackSingleRefs()
         quantum_records: Optional[Mapping[str, DatastoreRecordData]] = None
--- a/doc/changes/DM-31701.feature.rst
+++ b/doc/changes/DM-31701.feature.rst
@@ -1,0 +1,1 @@
+Made the ordering of input ``Quantum``s deterministic in graph building.


python/lsst/pipe/base/graphBuilder.py
```diff
@@ -311,6 +311,8 @@ def makeQuantum(self) -> Quantum:
         """
         allInputs = self.inputs.unpackMultiRefs()
         allInputs.update(self.prerequisites.unpackMultiRefs())
+        for name, refs in allInputs.items():
+            allInputs[name] = sorted(refs, key=lambda rr: rr.dataId)
```
natelust, line 315: Its not quite this simple, because you need to make sure all the inputs are sorted in the same order (grouped by how they correspond to each other) And that for a given set of inputs passed to run, the output references are sorted in the same way, such that something like `butlerQc.put(allOutputs[bla], self.run(**allInputs)` is valid (that is not the exact thing that needs to happen, just demoing the concern.

TallJimbo: Sorting outputs too is a good point, and zip-iteration is indeed not going to be enabled by this because input data IDs can still be missing from one connection and not another.

But in the absence of missing data IDs, this seems like it will ensure that the inputs with the same dimensions will be sorted consistently to me.  Am I missing something?

natelust: How does dataId sort thing as a comparable key? If I have one input that has (tract, patch, band), and other that is say (patch, band) or even just (band) how do I make sense of how any input has any relation to any other input, which as far as I can tell something that is useful sometimes. I see little value to having an individual connection sorted in isolation of everything else, unless you only care about 2 connections that have exactly the same dimensions.

TallJimbo: > unless you only care about 2 connections that have exactly the same dimensions

Yeah, I think that's exactly what this is for.  Or, well, it's what it would be for if it actually enabled zip-iteration reliably.  I actually think it's just about making the iteration order deterministic so we don't get e.g. random differences because a bunch of floats were added in different runs in unspecified order.

TallJimbo: > No, what I am saying is if you have 3 input connections, with slightly different dimensions but some have a common subset of dimensions, can you be sure that those are going to be sorted in the same order (I may be wrong, but isn't that an impossible sort?)

I think they'll be in the same order if and only if there are the same number of them (e.g. because one of the dimensions takes on only one value).  And if there aren't the same number of them, we don't care about consistent sorting.

But the important thing is still that you can _never_ count on just ordering anyway for zip iteration, because in any pair of connections, they could be missing different data IDs.  It is still true that if you want to be able to zip-iterate over any combination of connections, you have to override adjustQuantum or runQuantum to check for data ID consistency yourself; the only difference is that now this would be a check for set equality, not a sort _and_ a check for set equality.

natelust: You can have a `tract, patch, band` connection and and `tract, band` connection. In this case they will be ordered the same in tract, but the order in which bands appears might not be the same at all (unless you are doing something cleaver with dataId that I don't realize)

Or you could have `tract, patch, band` and say `visit, band` and I expect these have no common ordering, unless you can choose the primary dimension to sort over.

This leaves a situation where you still need to collate all the individual datarefs anyway according to whatever order is important, in which case, there really is little reason to pre-sort.

TallJimbo: > sometimes things are ordered in a way in which you can use them

I am categorically opposed to allowing zip-iteration or other assumptions of consistency without an `adjustQuantum` or `runQuantum` override.  I do concede that there's a new [Hyrum's Law](https://www.hyrumslaw.com/) hazard here, but I think determinism is worth it.

natelust: I think forcing people to confront that it is not ordered is better than having someone think it is generically applicable.

TallJimbo: We aren't doing that either, though; it is entirely possible (and I think the most likely outcome, at least in some contexts) for QG generation to _happen_ to make data ID lists that are ordered consistently.  If we want to poison the order so people don't depend on it, we need to do it explicitly.

As an alternative, we could start raising (perhaps start with a warning, for temporary backwards compatibility) when a set of connections with the same dimensions do not have the same set of data IDs, _and_ the task hasn't declared to the framework that it can handle this; perhaps in `ButlerQuantumContext`?  To be extra careful, we could even warn or raise whenever there are multiple connections with the same dimensions and the task hasn't declared itself capable of handling it _period_.

PaulPrice: Another suggestion from a Gen3 n00b (so maybe misguided): turn these lists of `DatasetRef` (inside the `inputRefs` and `outputRefs` that eventually get passed to `runQuantum`) into dicts indexed by `DatasetRef.dataId`. It seems to me that that's what every user is going to want when checking for consistency in `adjustQuantum` or `runQuantum`.

TallJimbo: That's a good idea but a tricky one for backwards compatibility.  I have a long-stalled ticket to do something similar by introducing a custom container for `DatasetRef`, with set-of-`DatasetRef` behavior overall but also easy access to the associated set of data IDs; I think that could be done in a backwards-compatible way.  Doesn't rule out unsafe zip-iteration entirely, but it does make it a lot easier to do the right thing.

TallJimbo:
Yes, it did fix the underlying issue as well - in fact, I think it was already fixed on `main` before that merge, because my first attempt at the QG gen rewrite did not make the dataset order deterministic and that broke some stuff downstream.

PR #242, opened 2022-04-12
DM-33493: Add datastore records to Quanta in a graph

andy-slac:
Graph builder adds an option to store datastore records in quanta in a
generated quantum graph.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/pipe/base/graphBuilder.py
+++ b/python/lsst/pipe/base/graphBuilder.py
@@ -41,6 +41,8 @@ from lsst.daf.butler import (
     DataCoordinate,
     DatasetRef,
     DatasetType,
+    Datastore,
+    DatastoreRecordData,
     DimensionGraph,
     DimensionUniverse,
     NamedKeyDict,
@@ -301,8 +303,14 @@ class _QuantumScaffolding:
     quantum.
     """
 
-    def makeQuantum(self) -> Quantum:
+    def makeQuantum(self, datastore_records: Optional[Mapping[str, DatastoreRecordData]] = None) -> Quantum:
         """Transform the scaffolding object into a true `Quantum` instance.
+
+        Parameters
+        ----------
+        datastore_records : `dict` [ `str`, `DatastoreRecordData` ], optional
+            If not `None` then fill datastore records in each generated Quantum
+            using the records from this structure.
 
         Returns
         -------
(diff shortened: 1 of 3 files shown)


python/lsst/pipe/base/graphBuilder.py
```diff
@@ -322,13 +330,25 @@ def makeQuantum(self) -> Quantum:
...
         helper.adjust_in_place(self.task.taskDef.connections, self.task.taskDef.label, self.dataId)
+        initInputs = self.task.initInputs.unpackSingleRefs()
+        quantum_records: Optional[Mapping[str, DatastoreRecordData]] = None
+        if datastore_records is not None:
+            quantum_records = {}
+            input_refs = list(itertools.chain.from_iterable(helper.inputs.values()))
+            input_refs += list(initInputs.values())
+            input_ids = set(ref.id for ref in input_refs if ref.id is not None)
+            for datastore_name, records in datastore_records.items():
+                matching_records = records.subset(input_ids)
```
TallJimbo, line 341: This `subset` is implemented as an operation that's linear in the total number of records, since `DatastoreRecordData` holds its records via a `list`.  And since we run this once on each quantum, we end up with this being O(N^2) in the size of the QuantumGraph.    That makes me a bit nervous, even if it's not a problem with the size of graph we have now.  Should we change `DatastoreRecordData` to hold its records via a `dict`, keyed by ID so lookups and set operations on IDs are faster?  Unfortunately the values of such a `dict` would still have to be containers like `list`, since composite disassembly makes it possible to have more than one row per ID (and possibly some future datastores will do the same, so I won't hold _this_ against composite disassembly, @timj :slightly_smiling_face:).

FWIW, I don't think that necessarily implies that the serialized form needs to use a `dict` instead of a `list`; that can be whatever is easiest and/or makes for the best serialized layout, since we'll always read and write full containers.

TallJimbo (approved):

python/lsst/pipe/base/graphBuilder.py
timj: There is no problem with this being represented as dict of dict. You can use the dataset_id in the toplevel dict and then have a dict indexed by component name (or `None` if composite). The primary key for datastore records is dataset_id+component. Datastore uses dict of list because in all the usages I'm iterating over the components and not trying to efficiently pull out a specific component (because a specific component has usually already been used in the table query so only one matching row would be returned).

PR #245, opened 2022-04-28
DM-34590: --longlog requires an unnecessary argument

PaulPrice:
--longlog is supposed to be a switch, and not require an argument,
but it was requiring one due to the absence of nargs=0.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/pipe/base/argumentParser.py
+++ b/python/lsst/pipe/base/argumentParser.py
@@ -509,7 +509,12 @@ class ArgumentParser(argparse.ArgumentParser):
             help="logging level; supported levels are [trace|debug|info|warn|error|fatal]",
             metavar="LEVEL|COMPONENT=LEVEL",
         )
-        self.add_argument("--longlog", action=LongLogAction, help="use a more verbose format for the logging")
+        self.add_argument(
+            "--longlog",
+            nargs=0,
+            action=LongLogAction,
+            help="use a more verbose format for the logging",
+        )
         self.add_argument("--debug", action="store_true", help="enable debugging output?")
         self.add_argument(
             "--doraise",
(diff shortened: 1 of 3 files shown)


timj:
I am pondering whether we really want to be doing gen2 fixes at this point since we are going to be deleting it soon. I think I'd rather there wasn't a change log entry for it since advertising gen2 bug fixes when we are saying everywhere else that gen2 is going away imminently might be more confusing than helpful.

PaulPrice:
@andy-slac The mypy failure appears to be related to DM-33493.

timj:
Which was #242 and did pass mypy at the time. I wonder if the mypy version changed.

timj:
mypy was 0.942 and is now 0.950. (conda does not seem to have 0.950 yet)

python/lsst/pipe/base/argumentParser.py
```diff
@@ -509,7 +509,12 @@ def __init__(self, name, usage="%(prog)s input [options]", **kwargs):
             help="logging level; supported levels are [trace|debug|info|warn|error|fatal]",
             metavar="LEVEL|COMPONENT=LEVEL",
         )
-        self.add_argument("--longlog", action=LongLogAction, help="use a more verbose format for the logging")
+        self.add_argument(
+            "--longlog",
+            nargs=0,
```
timj, line 514: I'm guessing this broke in DM-19634 (#178).

timj (approved):
Looks fine. We aren't currently planning to back port this to v23 and I can't promise it will be in v24 (although it might be).

python/lsst/pipe/base/argumentParser.py
PaulPrice: We have no expectations that this will survive past the next weekly.

PR #246, opened 2022-04-28
DM-33942: Use a negative integer for test universe

timj:
We currently cache universes based on version integer so we
need to use a version number that does not clash with any
possible real universe.

## Checklist

- [ ] ran Jenkins


timj:
Closed by #248

PR #251, opened 2022-05-24
DM-34924: Check for dataset type compatibility when creating execution butler

timj:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/pipe/base/executionButlerBuilder.py
+++ b/python/lsst/pipe/base/executionButlerBuilder.py
@@ -29,6 +29,7 @@ from typing import Callable, DefaultDict, Iterable, List, Mapping, Optional, Set
 
 from lsst.daf.butler import Butler, Config, DataCoordinate, DatasetRef, DatasetType
 from lsst.daf.butler.core.repoRelocation import BUTLER_ROOT_TAG
+from lsst.daf.butler.registry import ConflictingDefinitionError
 from lsst.daf.butler.transfers import RepoExportContext
 from lsst.resources import ResourcePath, ResourcePathExpression
 from lsst.utils.introspection import get_class_of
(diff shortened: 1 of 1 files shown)


python/lsst/pipe/base/executionButlerBuilder.py
```diff
@@ -39,6 +40,52 @@
...
+    if prevDsType := previous.get(name):
+        # Check compatibility. For now assume both directions have to
+        # be acceptable.
+        if prevDsType.is_compatible_with(candidate) and candidate.is_compatible_with(prevDsType):
+            # Store this mapping and declare that we will use it in the future.
+            previous[candidate] = prevDsType
+            # Use the first type encountered.
+            return prevDsType
+
+        raise ValueError(
```
mrawls, line 79: Suggest putting this in an explicit `else` clause for clarity

python/lsst/pipe/base/executionButlerBuilder.py
```diff
@@ -39,6 +40,52 @@
...
+            f"Dataset type incompatibility in graph: {prevDsType} not compatible with {candidate}"
+        )
+
+    # New dataset type.
+    previous[name] = candidate
+    previous[candidate] = candidate
```
mrawls, line 85: Is `previous` persisted somehow? I don't immediately see the benefit to updating it here.

python/lsst/pipe/base/executionButlerBuilder.py
```diff
@@ -39,6 +40,52 @@
...
+    candidate : `lsst.daf.butler.DatasetType`
+        The candidate dataset type.
+    previous : `dict` [Union[`str`, `DatasetType`], `DatasetType`]
+        Previous dataset types found, indexed by name and also by
+        dataset type. The latter provides a quick way of returning a
+        previously checked dataset type.
+
+    Returns
+    -------
+    datasetType : `lsst.daf.butler.DatasetType`
```
mrawls, line 60: Please clarify that it returns `candidate` (which is a DatasetType), or raises if `candidate` is not compatible

mrawls (approved):
I'm assuming I shouldn't worry too much about Codecov having lots of comments to share? I have minor comments only, overall this looks good and fixes my problem, thank you!

python/lsst/pipe/base/executionButlerBuilder.py
timj: It doesn't always return the supplied candidate. It will mostly return it but the purpose of this function is to sometimes have to return something different. The problem being solved is that if your pipeline declares a metadata dataset as input with PropertySet storage class and the graph has been built with the metadata dataset (that is automatically being injected as an output of every task) being a TaskMetadata one, you sometimes are given a TaskMetadata dataset type and sometimes a PropertySet. This function guarantees to return a consistent dataset type.

python/lsst/pipe/base/executionButlerBuilder.py
timj: It's used in the initial look up next time it gets called. There is a short circuit at the start of this function that sees if the dataset type has previously been handled and if it does exist it returns the value immediately. This saves having to do the is_compatible_with and == checks thousands of times as the graph is processed. I'm being a bit cute in storing keys of both DatasetType and the dataset type name in the same dict.  This is all done in line 64 above. It maps the candidate dataset type to a previously checked one and returns it if it already knows what to do. I will try to expand the function doc string.

python/lsst/pipe/base/executionButlerBuilder.py
```diff
@@ -218,7 +273,23 @@ def _import(
...
+            # an intermediate (and so must be able to support conversion
+            # from the registry storage class to an input) or solely an output
+            # dataset type. Test both compatibilities.
+            registryDsType = newButler.registry.getDatasetType(dsType.name)
+            if registryDsType.is_compatible_with(dsType) and dsType.is_compatible_with(registryDsType):
+                # Ensure that we use the registry type when inserting.
+                dsType = registryDsType
+            else:
+                # Not compatible to re-raise the original exception.
+                raise
```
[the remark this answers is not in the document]
timj: I am dreaming of python 3.11 exception notes. They would be perfect for this scenario. I think in this case though the exception that you are going to get is going to tell you "ConflictingDefinitionError(DatasetType A differs from Registry version B)" or something and there's not a lot I need to add.

PR #252, opened 2022-05-25
DM-34954: Fix Instrument.from_string see also section

jonathansick:
Numpydoc 1.2 was showing this See Also section with an extra `()` as an error instead of a warning.

## Checklist

- [ ] ran Jenkins


timj (approved):
Thanks. Turns out I had not quite got around to doing the sphinx build in this package (I only just made the gen2 stuff optional)

PR #256, opened 2022-06-10
DM-35047: Add upload to pypi

mwittgen:
## Checklist

- [ ] ran Jenkins


setup.cfg
```diff
@@ -1,11 +1,11 @@
 [metadata]
-name = pipe_base
+name = lsst-pipe-base
 description = Pipeline infrastructure for the Rubin Science Pipelines
```
timj, line 3: I think this file is missing the contact email.

timj (approved):
Thanks for adding the license. Do we need to list it in setup.cfg so it's included in the dist?

PR #259, opened 2022-07-04
DM-35082 Save DimensionUniverse in QuantumGraph

natelust:
Save the `DimensionUniverse` used to create the graph in the graphs header data. This removes the need to pass the Universe as an argument to load, and ensures that things are properly reconstructed, if the `DimensionUniverse` changes with a schema migration.

Requires lsst/daf_butler#711

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/pipe/base/config.py
+++ b/python/lsst/pipe/base/config.py
@@ -78,7 +78,7 @@ class TemplateField(pexConfig.Field):
 
     def __set__(
         self,
-        instance: pexConfig.Field,
+        instance: pexConfig.Config,
         value: Any,
         at: Optional[StackFrame] = None,
         label: str = "assignment",
@@ -107,7 +107,11 @@ class PipelineTaskConfigMeta(pexConfig.ConfigMeta):
     """
 
     def __new__(
-        cls: Type[_S], name: str, bases: Tuple[PipelineTaskConfig, ...], dct: Dict[str, Any], **kwargs: Any
+        cls: Type[_S],
+        name: str,
+        bases: Tuple[type[PipelineTaskConfig], ...],
+        dct: Dict[str, Any],
+        **kwargs: Any,
     ) -> _S:
         if name != "PipelineTaskConfig":
             # Verify that a connection class was specified and the argument is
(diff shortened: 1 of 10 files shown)


python/lsst/pipe/base/graph/graph.py
```diff
@@ -751,9 +756,11 @@ def loadUri(
         ----------
         uri : convertible to `ResourcePath`
             URI from where to load the graph.
-        universe: `~lsst.daf.butler.DimensionUniverse`
+        universe: `~lsst.daf.butler.DimensionUniverse` optional
             DimensionUniverse instance, not used by the method itself but
             needed to ensure that registry data structures are initialized.
+            If None it is loaded from the QuantumGraph saved structure. Only
+            pass this argument to overload explicitly.
```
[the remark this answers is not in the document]
andy-slac: If we want to override the universe it makes sense to verify that it is compatible (or identical). So maybe rephrase this to say that it validates graph universe against the argument?

python/lsst/pipe/base/config.py
```diff
@@ -179,14 +183,17 @@ class to allow configuration of the connections class. This dynamically
     `~lsst.pex.config.ConfigField` with the attribute name `connections`.
     """
 
-    saveMetadata = pexConfig.Field(
-        dtype=bool,
+    connections: pexConfig.ConfigField
+    """Field which refers to a dynamically added configuration class which is
+    based on a PipelineTaskConnections class
```
andy-slac, line 188: Sentences should end with a period.

python/lsst/pipe/base/pipeline.py
```diff
@@ -177,7 +181,7 @@ def metadataDatasetName(self) -> Optional[str]:
         """Name of a dataset type for metadata of this task, `None` if
         metadata is not to be saved (`str`)
         """
-        if self.config.saveMetadata:
+        if cast(PipelineTaskConfig, self.config).saveMetadata:
```
andy-slac, line 184: I think `self.config` needs to be a `PipelineTaskConfig`, can we just annotate it to be that to avoid `cast`?

andy-slac (approved):
Looks good, couple of minor comments.

python/lsst/pipe/base/graph/_versionDeserializers.py
```diff
@@ -474,14 +489,25 @@ def readHeaderInfo(self, rawHeader: bytes) -> SimpleNamespace:
         infoMappings.dimensionRecords = {}
         for k, v in infoMap["DimensionRecords"].items():
             infoMappings.dimensionRecords[int(k)] = SerializedDimensionRecord(**v)
+        # This is important to be a get call here, so that it supports versions
+        # of saved quantum graph that might not have a saved universe without
+        # changing save format
+        if (universeConfig := infoMap.get("universe")) is not None:
+            universe = DimensionUniverse(config=DimensionConfig(universeConfig))
+        else:
+            universe = DimensionUniverse()
```
timj, line 498: I worry a bit that we should be using a universe that was passed into the loader if such a universe exists, although I imagine this code path is never going to trigger because historically people haven't been persisting their graph so that they can load it again 6 months later.

PR #261, opened 2022-07-08
DM-35494 Allow execution butler creation to transfer datasets

PaulPrice:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/pipe/base/executionButlerBuilder.py
+++ b/python/lsst/pipe/base/executionButlerBuilder.py
@@ -227,7 +227,31 @@ def _export(
     return yamlBuffer
 
 
-def _setupNewButler(butler: Butler, outputLocation: ResourcePath, dirExists: bool) -> Butler:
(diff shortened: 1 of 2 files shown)


python/lsst/pipe/base/executionButlerBuilder.py
```diff
@@ -227,7 +227,31 @@ def _export(
...
+
+    Parameters
+    ----------
+    butler : `Butler`
+        The original butler, upon which the execution butler is based.
+    outputLocation : `ResourcePath`
+        Location of the execution butler.
+    dirExists : `bool`
+        Does the ``outputLocation`` exist, and if so, should it be clobbered?
+    datastoreRoot : `str`, optional
```
timj, line 246: I guess we should probably support `ResourcePath` here as well and then do `str(datastoreRoot)` below.

timj (approved):
Thanks. This is what I had in mind.

PR #262, opened 2022-07-11
DM-33492: Enable store of resolved DatasetRefs in QuantumGraph

andy-slac:
GraphBuilder adds an option to generate UUIDs for all unresolved
references. For consistency we need to save resolved InitOutputs
in the graph, so the graph is also extended to keep per-task refs for
InitInputs and InitOutputs. InitInputs refs are also duplicated in each
Quantum, we may consider dropping them later.

## Checklist

- [X] ran Jenkins

### DIFF
--- a/python/lsst/pipe/base/graphBuilder.py
+++ b/python/lsst/pipe/base/graphBuilder.py
@@ -33,12 +33,13 @@ import logging
 from collections import ChainMap
 from contextlib import contextmanager
 from dataclasses import dataclass
-from typing import Any, Collection, Dict, Iterable, Iterator, List, Mapping, Optional, Set, Union
+from typing import Any, Collection, Dict, Iterable, Iterator, List, Mapping, Optional, Set, Tuple, Union
 
 from lsst.daf.butler import (
     CollectionSearch,
     CollectionType,
     DataCoordinate,
+    DatasetIdGenEnum,
     DatasetRef,
     DatasetType,
     Datastore,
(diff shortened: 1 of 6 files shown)


python/lsst/pipe/base/graphBuilder.py
```diff
@@ -1017,6 +1063,20 @@ def resolveDatasetRefs(
...
+
+                # Resolve all task inputs and outputs.
+                if idMaker:
+                    for datasetType in task.inputs:
+                        resolved = idMaker.resolveDict(quantum.inputs[datasetType])
+                        quantum.inputs[datasetType].update(resolved)
+                    for datasetType in task.outputs:
+                        resolved = idMaker.resolveDict(quantum.outputs[datasetType])
+                        quantum.outputs[datasetType].update(resolved)
+                    for inits in (task.initInputs, task.initOutputs):
```
TallJimbo, line 1075: I haven't spotted why these two can be handled together but the inputs and outputs can't be.

TallJimbo (approved):
I took a quick look at this too, but while I know my way around `GraphBuilder` I'm not really a substitute for @natelust looking at the `QuantumGraph` changes.

python/lsst/pipe/base/graphBuilder.py
andy-slac: I thought that was the correct logic when I was writing it, but now I do not remember why 🙂 I'll try to unify that and see if that works.

andy-slac: I replaced that with more unified code.

python/lsst/pipe/base/graphBuilder.py
```diff
@@ -845,6 +881,11 @@ def resolveDatasetRefs(
                         collectionTypes=CollectionType.RUN,
                     )
 
+        idMaker: Optional[_DatasetIdMaker] = None
+        if resolveRefs:
+            assert run is not None, "run cannot be None when resolveRefs is True"
```
natelust, line 886: Is assert really the right kind of exception to raise here?

natelust (approved):
One minor comment

python/lsst/pipe/base/graphBuilder.py
andy-slac: I think it's fine here, it is internal code only called from `makeGraph()`. `makeGraph()` has an explicit check that raises `ValueError`, here assert is mostly to keep `mypy` happy.

PR #265, opened 2022-07-21
DM-35681: ensure QuantumGraph is passed a DimensionUniverse at construction

TallJimbo:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -13,7 +13,7 @@ repos:
         # supported by your project here, or alternatively use
         # pre-commit's default_language_version, see
         # https://pre-commit.com/#top_level-default_language_version
-        language_version: python3.8
+        language_version: python3.10
   - repo: https://github.com/pycqa/isort
     rev: 5.10.1
     hooks:
--- a/python/lsst/pipe/base/graph/graph.py
+++ b/python/lsst/pipe/base/graph/graph.py
@@ -142,9 +142,6 @@ class QuantumGraph:
         """Builds the graph that is used to store the relation between tasks,
         and the graph that holds the relations between quanta
         """
-        if universe is None:
-            universe = DimensionUniverse()
-        self._universe = universe
         self._metadata = metadata
         self._buildId = _buildId if _buildId is not None else BuildId(f"{time.time()}-{os.getpid()}")
         # Data structures used to identify relations between components;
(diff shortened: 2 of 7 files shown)


python/lsst/pipe/base/graph/graph.py
```diff
@@ -173,6 +170,15 @@ def _buildGraphs(
             # a newly created QuantumNode to the appropriate input/output
             # field.
             for quantum in quantumSet:
+                if quantum.dataId is not None:
```
TallJimbo, line 173: I'm not really bothered by codecov's complaints here: I'm just being super defensive about trying to avoid backwards-compatibilty breakage by guessing the `DimensionUniverse`, but really everybody should be using the code path that is tested, in which the universe is passed explicitly (since that's what `GraphBuilder`).

python/lsst/pipe/base/graph/graph.py
```diff
@@ -613,12 +628,13 @@ def subset(self: _T, nodes: Union[QuantumNode, Iterable[QuantumNode]]) -> _T:
...
         quantumDict: Dict[TaskDef, Set[Quantum]] = dict(quantumMap.items())
         # Create an empty graph, and then populate it with custom mapping
-        newInst = type(self)({})
+        newInst = type(self)({}, universe=self._universe)
         newInst._buildGraphs(
             quantumDict,
             _quantumToNodeId={n.quantum: n.nodeId for n in nodes},
             _buildId=self._buildId,
             metadata=self._metadata,
+            universe=self._universe,
```
MichelleGower, line 637: May be perfectly correct, but seems strange to create the object with the universe value on line 631 and then pass in the same value to that object's method on line 637.

.pre-commit-config.yaml
```diff
@@ -13,7 +13,7 @@ repos:
         # supported by your project here, or alternatively use
         # pre-commit's default_language_version, see
         # https://pre-commit.com/#top_level-default_language_version
-        language_version: python3.8
+        language_version: python3.10
```
MichelleGower, line 16: The changes in this ticket required jumping to python 3.10?

MichelleGower (approved):
Just a couple questions probably more for my info than anything about the code itself.  I don't know all of this code so there may be some location that was missed.  But I ran the pipe_base and bps unit tests which were all fine.  Also I submitted a test run on the idf that had the warning before and now it doesn't.  Merge approved.

python/lsst/pipe/base/graph/graph.py
TallJimbo: I agree it's strange; it's an artifact of some hoops the code jumps through to avoid having the "usual" `__init__` logic do a lot of unnecessary work.  Basically, passing the universe to `__init__` just saves `__init__` from complaining that it doesn't have a universe, and then the next call actually sets up the graph (including setting the universe).

.pre-commit-config.yaml
TallJimbo: No, it's just something that should have been done earlier that had gone unnoticed until now (I think it only matters when someone does `pre-commit install`).

PR #267, opened 2022-07-26
DM-35675: Remove Gen2

timj:
## Checklist

- [x] ran Jenkins


python/lsst/pipe/base/argumentParser.py
```diff
@@ -29,57 +29,16 @@
...
-        if path is None:
-            return None
-        return os.path.abspath(path)
-    return os.path.abspath(os.path.join(defRoot, path or ""))
+from deprecated.sphinx import deprecated
 
 
+@deprecated(
+    reason="Gen2 DataIdContainer is no longer supported. This functionality has been disabled.",
+    version="v23.0",
```
erykoff, line 39: In all of these, should the version listed be `v23.0` (current) or `v24.0` (next), thus promising that this routine will be in v24 but once that is released can be deleted? Reading the docs is ... unclear ... but that's how I read it: https://developer.lsst.io/stack/deprecating-interfaces.html#python-deprecation

erykoff (approved):
I'm not sure if the version strings for the deprecations are correct.  Otherwise, an impressive score of +53 -2847!

python/lsst/pipe/base/argumentParser.py
timj: We already promised that Gen2 was deprecated and could be removed any time as part of the v23 release. My statement is that CmdLineTask has been deprecated since v23 but now actually tells you it has been deprecated since then. I don't really see any point in keeping this code around for another two release cycles where it doesn't actually do anything and will break if something uses it.  The "version" here is the version the code was formally deprecated. I have not included in the statements when we will actually delete the stubs but I was assuming I could do it as soon as nothing in lsst_distrib uses CmdLineTask.

timj: I talked it over with @ktlim and he thinks that saying "since v23" is fine here.

PR #268, opened 2022-07-28
DM-35741: Add InMemoryDatasetHandle

timj:
This is like a DeferredDatasetHandle but allows you to create
it without a butler so that Task.run() methods can be more
easily tested.

Requires lsst/daf_butler#719

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/pipe/base/_dataset_handle.py
+++ b/python/lsst/pipe/base/_dataset_handle.py
@@ -1,0 +1,160 @@
+# This file is part of pipe_base.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (http://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+from __future__ import annotations
+
+__all__ = ["InMemoryDatasetHandle"]
+
+import dataclasses
+from typing import Any, Optional
+
+from lsst.daf.butler import DataCoordinate, DimensionUniverse, StorageClass, StorageClassFactory
+from lsst.utils.introspection import get_full_type_name
+
+
+# Use an empty dataID as a default.
+def _default_dataId() -> DataCoordinate:
+    return DataCoordinate.makeEmpty(DimensionUniverse())
+
+
+@dataclasses.dataclass(frozen=True)
+class InMemoryDatasetHandle:
+    """An in-memory version of a `~lsst.daf.butler.DeferredDatasetHandle`."""
+
+    def get(
+        self, *, component: Optional[str] = None, parameters: Optional[dict] = None, **kwargs: dict
+    ) -> Any:
+        """Retrieves the dataset pointed to by this handle
+
+        This handle may be used multiple times, possibly with different
+        parameters.
+
+        Parameters
+        ----------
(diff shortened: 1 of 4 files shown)


python/lsst/pipe/base/_dataset_handle.py
```diff
@@ -0,0 +1,153 @@
...
+__all__ = ["InMemoryDatasetHandle"]
+
+import dataclasses
+from typing import Any, Optional
+
+from lsst.daf.butler import DataCoordinate, DimensionUniverse, StorageClass, StorageClassFactory
+from lsst.utils.introspection import get_full_type_name
+
+# Use an empty dataID as a default.
+_DEFAULT_DATAID = DataCoordinate.standardize({}, universe=DimensionUniverse())
```
TallJimbo, line 32: I'm not a fan of instantiating the default dimension universe at import time unconditionally.  Mind turning this into a function, and using the `default_factory` argument to `dataclasses.field` instead?

There's also a `DataCoordinate.makeEmpty(universe)` method you can use.

And if we're not absolutely certain we need a default for the data ID, it might be best to remove it to avoid surprising behavior when some task tries to look at the data ID.

python/lsst/pipe/base/_dataset_handle.py
```diff
@@ -0,0 +1,153 @@
...
+        parameters.
+
+        Parameters
+        ----------
+        component : `str` or None
+            If the deferred object is a component dataset type, this parameter
+            may specify the name of the component to use in the get operation.
+        parameters : `dict` or None
+            The parameters argument will be passed to the butler get method.
+            It defaults to None. If the value is not None,  this dict will
```
TallJimbo, line 54: extra space

python/lsst/pipe/base/_dataset_handle.py
```diff
@@ -0,0 +1,153 @@
...
+        """Retrieves the dataset pointed to by this handle
+
+        This handle may be used multiple times, possibly with different
+        parameters.
+
+        Parameters
+        ----------
+        component : `str` or None
+            If the deferred object is a component dataset type, this parameter
+            may specify the name of the component to use in the get operation.
```
TallJimbo, line 51: Did you mean to write "if the deferred object is a composite dataset type"?

Does this all only supports the case where you can initialize the handle with the full composite, not just the component that you'll later need?

TallJimbo (approved):
I think this is fine as-is (just minor line comments on the PR), but I think it's worth thinking about where this should land on the spectrum of "easy to construct" (where the PR leans now) vs. "behaves more like a real `DeferredDatasetHandle`".  We don't have an ABC that defines what interface `PipelineTasks` can expect of these handles, just the concrete implementation, and that actually exposes a `LimitedButler` and `DatasetRef`, too.  Exposing the butler was probably a mistake (and probably mine, as I know @natelust has historically been much more careful about preventing task access to things than I have been), but it's quite possible it was an "intentional" mistake that was actually a pragmatic-at-the-time solution to a real problem, like trying to get at a data ID packer from deep inside some subtask.  And while I can't come up with a good reason for a task to get the `DatasetRef` if they've already got the data ID - they should only be referring to the dataset type via their own connection name - it also doesn't seem too onerous to demand that an `InMemoryDatasetHandle` user provide one, and in doing so sidestep all of your storage class inference concerns (though I'm not really bothered by the logic there now).

So, after typing all of that out, I think we probably do want to leave this as it is - but maybe we should RFC deprecating those public `butler` and `ref` attributes on `DeferredDatasetHandle`, so we can spot any existing usage and make sure no new usage develops.

python/lsst/pipe/base/_dataset_handle.py
timj: If someone gives this a component I think that's no different to it being given any other type of non-composite object. People don't do `.get(component="x")` if they had already passed in component `x` do they? I might be missing something.

TallJimbo: No, I think you're right - if someone starts with a component, then it just works.

timj:
I admit I had failed to notice that `.ref` and `.butler` are public interfaces for `DeferredDatasetHandle`. I would be interested in knowing if anyone is using them. Maybe I can run a test with them as internal parameters and see if anything breaks. Making people always pass in a butler seems to go against making this thing easy to use if you just want to call a `.run()` method with some objects you already have.

I was wondering if we had a need for an abstract base class and I'm still wondering if this code should be in daf_butler.

PR #270, opened 2022-08-03
DM-35803: Add test for InMemoryDatasetHandle(DataFrame).

erykoff:
## Checklist

- [ ] ran Jenkins

### DIFF
--- a/tests/test_dataset_handle.py
+++ b/tests/test_dataset_handle.py
@@ -25,6 +25,12 @@ from lsst.daf.butler import DataCoordinate, DimensionUniverse, StorageClassConfi
 from lsst.daf.butler.tests import MetricsExample
 from lsst.pipe.base import InMemoryDatasetHandle
 
+try:
+    import numpy as np
+    import pandas as pd
+except ImportError:
+    pd = None
+
 storageClasses = """
 Integer:
   pytype: int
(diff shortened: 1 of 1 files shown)


timj:
You are going to have to install pandas in the github action so that the test runs. Add it where moto is installed.

erykoff:
I figured as much. Now I don't know what to do since this needs the ticket branch of daf_butler...?

timj:
You make a temporary change to requirements.txt to point at the daf_butler ticket branch rather than main. Then remember not to merge that commit later.

timj:
As a general comment, I'm not convinced this pull request is needed at all. We know from daf_butler that the delegate works and we know that we have 100% test coverage of the InMemoryDatasetHandle using that delegate API.

timj:
I think this is telling us that the core of  `InMemoryDatasetHandle._getStorageClass()` method should in fact move to `daf_butler` and be `StorageClassFactory.findStorageClass(pytype: Type) -> StorageClass`... Do you want to do that? Then you can test that `findStorageClass` works in your test_parquet.py and we can drop a lot of this duplicated test code from pipe_base.

erykoff:
That looks like something that is beyond my abilities...

timj:
Closed because #271 moved the logic being tested down into daf_butler.

PR #271, opened 2022-08-03
DM-35815: Use StorageClassFactory.findStorageClass

timj:
Requires lsst/daf_butler#721

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/pipe/base/_dataset_handle.py
+++ b/python/lsst/pipe/base/_dataset_handle.py
@@ -26,7 +26,6 @@ import dataclasses
 from typing import Any, Optional
 
 from lsst.daf.butler import DataCoordinate, DimensionUniverse, StorageClass, StorageClassFactory
-from lsst.utils.introspection import get_full_type_name
 
 
 # Use an empty dataID as a default.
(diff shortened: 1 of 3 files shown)


python/lsst/pipe/base/_dataset_handle.py
```diff
@@ -127,14 +127,14 @@ def _getStorageClass(self) -> StorageClass:
...
-        for storageClass in factory.values():
-            # It is possible for a single python type to refer to multiple
-            # storage classes such that this could be quite fragile.
-            if storageClass.is_type(pytype):
-                return storageClass
+
+        try:
+            return factory.findStorageClass(pytype)
+        except KeyError:
+            pass
```
erykoff, line 134: Would this be clearer if it just raised the ValueError here?

erykoff (approved):

python/lsst/pipe/base/_dataset_handle.py
timj: I don't like raising inside an `except` block unless that's really necessary because of all the exception chaining that's not really relevant to the caller. A clean exception outside seemed better. The alternative is to say that this method actually raises `KeyError` and let the exception survive as is.

PR #272, opened 2022-08-05
DM-971: validate task configs during init

parejkoj:


natelust:
This happens in another location already, let me look up where

natelust:
Sorry I guess that is in the context of using tasks in terms of execution, not in just instantiation. Do we care about multiple validations?

parejkoj:
We can remove the `validate` call from pipetask once DM-972 is implemented and the config is frozen (but it shouldn't actually matter). We want to validate when instantiating the `Task`, to ensure behavior is the same in both notebooks and pipetask.

natelust:
I worry that if you are trying to enure the same behavior in pipetasks and notebooks this does not quite do that. If a config is validated when it is instantiated, but the config is not frozen, then it is possible it will pass validation when created, and a subsequent cell execution could alter the config in such a state that the config is no longer valid. Doing this in an execution framework ensures everything happens in the correct order.

I would say we could ensure the config is frozen prior to validating, but there likely could be valid reasons in certain circumstances where that might not be desirable, possibly in something like a unit tests.

parejkoj:
See DM-972 for freezing configs before validation. The reason it's going to be difficult to implement is our existing test behavior, where we change some configs repeatedly. Its fixable, just a pain.

erykoff:
I think this is fine, but I would like @natelust to chime in with the final "review" to ensure that his concerns have been addressed.

natelust:
I get that freezing configs is hard because of unit tests and such, and that there is a ticket to do it. Would you be open to throwing a warning if validation passes but the config is not frozen about potentially bad behavior if the config is modified, at least until that other ticket is done?

natelust:
Something like
```
config.validate()
if not config._frozen:  # why don't we have a property for this???
    warnings.warn(""Config used in task has not been frozen, any future modification may cause issues")
```

parejkoj:
I suspect there's no property for frozen configs, because we were expecting them to be frozen on task init (i.e. DM-972 is one of our earliest post-JIRA tickets!).

I could put in that warning, but it'll show up in a *lot* of tests, and it may be a while before anyone gets to DM-972. At least with this ticket done, problems like what Merlin has described (using a task in a notebook with a config that doesn't work in a pipetask run because it isn't valid) won't happen any more, and invalid configs are more likely to be caught.

natelust:
But I worry about the other problem with notebooks and such, where people create a config, instantiate a task, run cells that modify configs (either intentionally or by accident) and then their notebook fails because it was never validated after the changes.

I personally don't mind the warnings showing up, because it is more encouragement to NOT do things that way, and will help us find all the places that do, but that is just my opinion.

parejkoj:
I've added that warning: it results in 50 more warnings when running `pipe_base` tests, and an additional ~500 warnings in `pipe_tasks`, for the record. I don't know when anyone plans to schedule work on DM-972, so the additional >1000 warnings (rough guess) over the whole stack might be around a while.

natelust:
It's up to you, and or anyone else if you want to raise it in a room to get opinions. If you don't think it is worth it that's fine, do you have any other ideas how to help protect people (as we are adding this to protect people in a different way)

erykoff:
Then I would prefer not to put in the warning, because I _think_ that with this ticket the situation is better than before and not worse in any way.

What we really need is a way of flagging a config as "validated" but not "frozen" (replace the icebox with a fridge) and then only warn if you try to set anything on a "chilled" config.  (Maybe this, with less goofy terminology, could be a way of getting to DM-972 without blowing everything up?)

natelust:
What if we put an optional argument to init, `freezeConfig` or something like that, that defaults to True, but we could set to False for all the tests that cant handle it. (Alternatively on the Config class, but I would prefer not to mess with that part of pex_config). That way we call `validate` and then `freeze` unless told not to explicitly

erykoff:
@natelust maybe that's a mostly solution to DM-972?

parejkoj:
DM-972 will be "freeze then validate", to prevent race conditions.

An additional `freezeConfig` param might be almost as painful to add to the many, many tests that currently misbehave as would just fixing them to not misbehave. I definitely don't want to do something like that on this ticket, since I've been poking at it since ~December 2021, and want it off my plate.

natelust:
My objections have been lodged, if you feel its not worth addressing, then so be it, I'm not going to die on this hill or anything.

erykoff (approved):

PR #280, opened 2022-09-21
DM-36174: Require Python 3.10.

TallJimbo:
daf_butler will soon require 3.10, even if code directly in pipe_base does not yet.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yaml
```diff
@@ -84,7 +83,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v4
         with:
-          python-version: 3.8
+          python-version: 3.10
```
timj, line 86: I think this will need quotes otherwise it will by python 3.1.

timj (approved):

PR #285, opened 2022-10-10
DM-36145: Add additional quanta information for pipetask run

leeskelvin:
## Checklist

- [ ] ran Jenkins

### DIFF
--- a/python/lsst/pipe/base/graph/graph.py
+++ b/python/lsst/pipe/base/graph/graph.py
@@ -440,7 +440,7 @@ class QuantumGraph:
             The `set` of `Quantum` that is associated with the specified
             `TaskDef`.
         """
-        return frozenset(node.quantum for node in self._taskToQuantumNode[taskDef])
+        return frozenset(node.quantum for node in (self._taskToQuantumNode.get(taskDef) or tuple()))
 
     def getNodesForTask(self, taskDef: TaskDef) -> FrozenSet[QuantumNode]:
         """Return all the `QuantumNodes` associated with a `TaskDef`.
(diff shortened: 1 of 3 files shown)


python/lsst/pipe/base/graph/graph.py
```diff
@@ -440,7 +440,7 @@ def getQuantaForTask(self, taskDef: TaskDef) -> FrozenSet[Quantum]:
             The `set` of `Quantum` that is associated with the specified
             `TaskDef`.
         """
-        return frozenset(node.quantum for node in self._taskToQuantumNode[taskDef])
+        return frozenset(node.quantum for node in (self._taskToQuantumNode.get(taskDef) or tuple()))
```
taranu, line 443: If it's dict-like, `self._taskToQuantumNode.get(taskDef, ())` or similar should work.

leeskelvin: I've tried your suggestion, and it does work as intended. With that said, I feel like the syntax here is more explicit about what it's doing, which I prefer. My preference would be to keep it as-is (and similarly below in the new method that's been added), but if you feel strongly otherwise, I'd be happy to change it.

taranu: It shouldn't matter in this case because I wouldn't expect the get call to return anything False-but-not-None, but they're not identical. The current form would ignore any `False`y values (literal `False`, `0`, etc.), which I would find generally find to be a strange choice if deliberate. Consider `x = {'a': 0}` followed by `if y := x.get('a'): print(y)`; this prints nothing.

timj: I would tend to agree that this code is saying that the result should be a tuple if there is no element in the dict and so `.get()` is closer to that intent. It's not really saying that tuple should be returned if any of the elements evaluate to false.

timj: Yes, I was agreeing with you 100% but my phrasing was ambiguous.

leeskelvin: Thank you both. Both relevant line edited as per your recommendation.

PR #286, opened 2022-10-10
DM-36080: Remove an assert against component outputs.

kfindeisen:
The assert inadvertently catches component *inputs* because of an unidentified bug elsewhere. This breaks all pipelines that use the default config for `ip.diffim.GetTemplateTask`.


TallJimbo (approved):
Looks good, and thanks for taking care of it.  I'm adding a note to DM-33027 to track down the underlying bug there, so feel free to reference that ticket in the TODO.

kfindeisen:
Merging immediately, because this bug breaks GHA builds on lsst-dm/prompt_prototype#24.

PR #288, opened 2022-10-17
DM-33779: Implement drp_pipe pipeline unit tests

leeskelvin:
## Checklist

- [x] ran Jenkins


python/lsst/pipe/base/tests/pipelineStepTester.py
```diff
@@ -0,0 +1,118 @@
...
+    the named pipeline, either as intermediates or as final outputs.
+
+    These sets will be checked against user-supplied sets to ensure that the
+    named pipeline may still be run without raising a missing data error.
+
+    Parameters
+    ----------
+    filename : `str`
+        The full path to the pipeline YAML
+    step_suffixes : `list` of `str`
```
arunkannawadi, line 56: For documentation of containers, you should use this style:
https://developer.lsst.io/python/numpydoc.html#sequence-types

arunkannawadi: Why are these lists, as opposed to say sets? Does the order matter here?

arunkannawadi (approved):
Minor inline comments on documentation, otherwise all LGTM!

python/lsst/pipe/base/tests/pipelineStepTester.py
leeskelvin: Documentation updated as per the referenced style guide.

leeskelvin: Yes, the order is important in `step_suffixes`. Specifically, the pipeline needs to be progressively built in the order in which it's executed. This allows us to build up a list of overall expected input dataset types, and known expected outputs.

The `initial_dataset_types` parameter needs to be a list as it contains unhashable types.

PR #294, opened 2022-12-15
DM-33497: Updates for quantum-backed butler

andy-slac:
Multiple changes to support pipetask execution with quantum-backed butler.

## Checklist

- [X] ran Jenkins


python/lsst/pipe/base/cli/opt/options.py
```diff
@@ -19,8 +19,22 @@
...
 instrument_option = MWOptionDecorator(
     "--instrument", help="The name or fully-qualified class name of an instrument."
 )
+
+
+update_output_chain_option = MWOptionDecorator(
+    "--update-output-chain",
+    help=unwrap(
+        """If quantum graph metadata includes output run name and output collection
+        which is a chain, update the chain definition to include run name as a the
```
TallJimbo, line 35: "as a the" -> "as the"

python/lsst/pipe/base/script/transfer_from_graph.py
```diff
@@ -0,0 +1,138 @@
...
+def _update_chain(registry: Registry, output_chain: str, output_run: str, inputs: list[str] | None) -> None:
+    """Update chain definition if it exists to include run as the first item
+    in a chain. If it does not exist then create it to include all inputs and
+    output.
+    """
+    try:
+        chain_definition = list(registry.getCollectionChain(output_chain))
+    except CollectionTypeError:
+        # Not a chained collection, ignore it.
+        # TODO: Print a warning message?
```
TallJimbo, line 114: I think I'd just let this exception propagate up.  I can't think of a scenario where it should just be ignored - it seems to just be user error that they should fix, even if that means they need to run the transfer job again.

python/lsst/pipe/base/script/transfer_from_graph.py
```diff
@@ -0,0 +1,138 @@
...
+        return
+    except MissingCollectionError:
+        # This probably means that someone forgot to make output chain
+        # collection, we want to create it to include inputs and output run
+        # (this reproduces logic in CmdLineFwk).
```
TallJimbo, line 119: I don't think this is a case of "forgot" - it's just the common case where this is the first processing for that particular chain, as we're careful to avoid creating the chain until we have something to actually put in it.

python/lsst/pipe/base/graphBuilder.py
```diff
@@ -501,6 +501,9 @@ def __init__(self, registry: Registry, run: str):
     def resolveRef(self, ref: DatasetRef) -> DatasetRef:
         if ref.id is not None:
             return ref
+        # For components we need their parent dataset type.
+        if ref.isComponent():
+            ref = ref.makeCompositeRef()
```
TallJimbo, line 506: If we're calling this method in an order consistent with the graph, a component appearing here should mean the lookup in `self.resolved` should always succeed.  It might be worth using an else clause here to ensure that; we certainly don't want some other bug to lead to generating random UUIDs for components.

TallJimbo (approved):

python/lsst/pipe/base/script/transfer_from_graph.py
andy-slac: The exception here could happen, for example, when `--update-output-chain` is specified but QGraph metadata specifies non-chained collection as the `output`. Re-running a whole transfer job in this case does not make sense because all datasets should have been transferred already, and only the last step of attempting to update the chained collection definition fails. It may be more natural to tell users that `--update-output-chain` did not do anything instead of raising an exception here. And I do not know if this is going to be a common issue or if it's going to be an issue at all.

TallJimbo: I wouldn't have thought it was possible to put a non-chained collection as the `output` in the QG metadata; that seems like an error.

andy-slac: I think you can potentially put non-existing collection into metadata and later someone would make it with non-chained type. But that should deserve an exception too, I guess. I'll let propagate that exception and we'll see how that works.

python/lsst/pipe/base/graphBuilder.py
andy-slac: If a component is a pure input (not an intermediate) then its parent should not be in `self.resolved`? But then it should be already resolved, which I can certainly check here too.

TallJimbo (approved):

PR #295, opened 2023-01-04
DM-35962: Remove examples/ from all packages

parejkoj:


timj (commented):
Doesn't @natelust want these examples in the documentation? Maybe they should be moved to the doc directory?

natelust:
Yeah, we cant just drop all of those

parejkoj:
The existing documentation is broken anyway, since they don't appear: https://pipelines.lsst.io/v/daily/modules/lsst.pipe.base/creating-a-pipelinetask.html#appendix-a

natelust:
I know they did work at some point, so we should fix it, it is how the document is designed to progressively
 introduce concepts

natelust:
It is supposed to be a symbolic link here anyway https://github.com/lsst/pipe_base/blob/main/doc/lsst.pipe.base/PipelineTask_Examples per a suggestion from jsick during a review if I remember. So it can probably just be moved

TallJimbo:
:+1: to moving these to the doc directory; the example source code really benefits from the context of the docs, and that'll keep anyone else from thinking their dead code like our other examples in the future.

natelust (approved):

PR #304, opened 2023-02-14
DM-37786: allow PipelineTasks to control default dataset-query-constraint behavior.

TallJimbo:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.10"]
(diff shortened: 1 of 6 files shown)


timj:
@TallJimbo would you mind adding python 3.11 to the build matrix whilst you are doing this PR? It should work fine, ctrl_mpexec builds okay.

natelust (approved):
While I am slightly worried about the multiple ways and places dataset constraints are used that don't quite fully overlap being confusing to futures maintainers, I think this is a pragmatic approach without larger refactoring.

TallJimbo:
I had the same misgivings and sense that it was the least-bad option, and I'll add some code comments to that effect.

PR #306, opened 2023-02-17
DM-37786-v25: backport hooks for tasks to set dataset-query-constraint defaults

TallJimbo:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/doc/changes/DM-37786.bugfix.md
+++ b/doc/changes/DM-37786.bugfix.md
@@ -1,0 +1,1 @@
+Fix an error message that says that repository state has changed during QuantumGraph generation when init input datasets are just missing.
--- a/doc/changes/DM-37786.feature.md
+++ b/doc/changes/DM-37786.feature.md
@@ -1,0 +1,1 @@
+Allow `PipelineTasks` to provide defaults for the `--dataset-query-constraints` option for the `pipe_task` tool.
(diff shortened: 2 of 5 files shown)


TallJimbo:
Since this is a backport and Jenkins has succeeded, I'm going to ignore the CI failures here; those seem to be due to a newer Python stack or are otherwise unrelated to these changes, and `v25` is not trying to keep up with bleeding edge the same way `main` is.

timj:
I'm actually confused how it managed much at all since it installed sqlalchemy2.

timj:
@TallJimbo I don't think we can leave this branch in a failing state -- the build tests must pass before the PyPI uploads are completed so if we want the 25.0.0 release to appear on PyPI (I would suggest that we do) we have to make sure the tests pass. This would likely involve at least pinning sqlalchemy to <2.  

I'm not sure why the open-files failures are triggering now but the fix is to add an `open_files_ignore` section to the `pyproject.toml` as we do in `daf_butler`.

timj:
We probably also on this branch want to restrict the versions of dependencies we define in pyproject.toml to be `>24, <25.2000` and change the requirements.txt to point at 25.0.x branches of dependencies for the builds.

TallJimbo:
Sounds like we need to consider this part of including middleware in point releases (as well as v25, when it actually goes out for the first time, of course).

timj:
I'll have to add all the release notes to the 25 branch so I'll fix things up then.

PR #313, opened 2023-03-03
DM-38234: Improve DuplicateOutputError log message

leeskelvin:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/pipe/base/pipeTools.py
+++ b/python/lsst/pipe/base/pipeTools.py
@@ -153,14 +153,17 @@ def orderPipeline(pipeline: List[TaskDef]) -> List[TaskDef]:
     outputs = {}  # maps task index to its output DatasetType names
     allInputs = set()  # all inputs of all tasks
     allOutputs = set()  # all outputs of all tasks
+    dsTypeTaskLabels: dict[str, str] = {}  # maps DatasetType name to the label of its parent task
     for idx, taskDef in enumerate(pipeline):
         # task outputs
         dsMap = {name: getattr(taskDef.connections, name) for name in taskDef.connections.outputs}
         for dsTypeDescr in dsMap.values():
             if dsTypeDescr.name in allOutputs:
                 raise DuplicateOutputError(
-                    "DatasetType `{}' appears more than once as output".format(dsTypeDescr.name)
+                    f"DatasetType `{dsTypeDescr.name}' in task `{taskDef.label}' already appears as an "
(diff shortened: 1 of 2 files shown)


doc/changes/DM-38234.feature.md
```diff
@@ -0,0 +1 @@
+`DuplicateOutputError` logger now produces a more helpful error message.
```
timj, line 1: I'd be inclined to call this category "misc" rather than "feature".

python/lsst/pipe/base/pipeTools.py
```diff
@@ -153,14 +153,20 @@ def orderPipeline(pipeline: List[TaskDef]) -> List[TaskDef]:
...
         # task outputs
         dsMap = {name: getattr(taskDef.connections, name) for name in taskDef.connections.outputs}
         for dsTypeDescr in dsMap.values():
             if dsTypeDescr.name in allOutputs:
                 raise DuplicateOutputError(
-                    "DatasetType `{}' appears more than once as output".format(dsTypeDescr.name)
+                    "DatasetType `{}' in task `{}' already appears as an output in task `{}'.".format(
+                        dsTypeDescr.name,
+                        taskDef.label,
+                        dsTypeTaskLabels[dsTypeDescr.name],
```
kfindeisen, line 166: I think an f-string might be more readable here; it's a bit disorienting that the string is laid out horizontally while the substitution arguments are laid out vertically.

python/lsst/pipe/base/pipeTools.py
```diff
@@ -153,14 +153,20 @@ def orderPipeline(pipeline: List[TaskDef]) -> List[TaskDef]:
...
+                    )
                 )
+            dsTypeTaskLabels[dsTypeDescr.name] = taskDef.label
```
kfindeisen, line 169: The fact that the test is against `allOutputs` while the insertion is against `dsTypeTaskLabels` is a bit awkward, logically. Is it already guaranteed that a task can't have two connections with the same name?

kfindeisen (approved):
Looks good, just some clarity comments.

python/lsst/pipe/base/pipeTools.py
leeskelvin: A single task cannot output the same dataset type twice. The important insertion is `dsTypeDescr.name` into `outputs`, which then gets subsumed into `allOutputs` to facilitate future checks. The insertion you've highlighted here is merely a book-keeping exercise that keeps a record of the prior tasks which output a given dataset type.

PR #316, opened 2023-04-10
DM-38614: more fixes for storage class conversion support

TallJimbo:
## Checklist

- [ ] ran Jenkins

### DIFF
--- a/python/lsst/pipe/base/graphBuilder.py
+++ b/python/lsst/pipe/base/graphBuilder.py
@@ -160,11 +160,27 @@ class _DatasetDict(NamedKeyDict[DatasetType, Dict[DataCoordinate, DatasetRef]]):
                     # stream code will be more explicit about input
                     # vs output incompatibilities.
                     existing = combined_by_name[datasetType.name]
-                    if existing.is_compatible_with(datasetType) or datasetType.is_compatible_with(existing):
-                        _LOG.warning(
-                            "Dataset type mismatch (%s != %s) but continuing since they are compatible",
-                            datasetType,
-                            existing,
+                    convertible_to_existing = existing.is_compatible_with(datasetType)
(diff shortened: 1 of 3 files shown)


python/lsst/pipe/base/graphBuilder.py
```diff
@@ -160,11 +160,27 @@ def fromSubset(
...
+                    convertible_from_existing = datasetType.is_compatible_with(existing)
+                    if convertible_to_existing and convertible_from_existing:
+                        _LOG.debug(
+                            "Dataset type %s has multiple fully-compatible storage classes %s and %s",
+                            datasetType.name,
+                            datasetType.storageClass_name,
+                            existing.storageClass_name,
+                        )
+                        _dict[datasetType] = combined[existing]
+                    elif convertible_to_existing or convertible_from_existing:
```
timj, line 173: The code in the graph builder tries to work out whether this is solely an output or solely an input or is an intermediate and so doesn't always require that both directions work.

timj (approved):
Looks good.

python/lsst/pipe/base/graphBuilder.py
timj: I think I was mainly commenting on the comment saying that we can't tell whether this is an error or not. I'm not sure where in the graph builder this is occurring so it might be that the test that compare outputs to intermediates to inputs happen later. I do remember trying to be careful about only complaining (and failing to build the graph) if you try to use something as an intermediate that won't convert in both directions (it's possible that is buggy of course).

TallJimbo: Ah, ok, all I meant with the comment was that this code _right here_ doesn't have the information it needs to make that determination.  Other parts of the graph builder definitely can.

PR #317, opened 2023-04-10
DM-36162: Minor documentation fixes

kfindeisen:
This PR fixes some minor errors in the documentation of `QuantumGraph`.


python/lsst/pipe/base/graph/graph.py
```diff
@@ -356,27 +359,27 @@ def graph(self) -> nx.DiGraph:
...
-        to the graph, meaning those nodes to not depend on any other nodes in
+        """Return all `QuantumNode` objects that are 'input' nodes
+        to the graph, meaning those nodes do not depend on any other nodes in
         the graph.
 
         Returns
         -------
         inputNodes : iterable of `QuantumNode`
-            A list of nodes that are inputs to the graph
+            An iterable of nodes that are inputs to the graph
```
TallJimbo, line 369: Thanks for the fix, but no good deed goes unpunished - given that this a property, I think it should be documented like an attribute, i.e.
```
"""An iterable of nodes that are input to the graph (iterable of `QuantumNode`).

These are the nodes that do not depend on any other nodes in the graph.
"""
```

As an aside, having a `property` that's a generator expression bothers me a fair bit, because it means the iterator gets reset every time you access it, and that feels very un-attribute-like, but that's not something we can fix without an RFC.

TallJimbo (approved):

doc/changes/DM-36162.doc.md
```diff
@@ -0,0 +1,4 @@
+Loosened documentation of QuantumGraph.inputQuanta and outputQuanta.
```
timj, line 1:
```suggestion
Loosened documentation of `QuantumGraph.inputQuanta` and `outputQuanta`.
```

doc/changes/DM-36162.doc.md
```diff
@@ -0,0 +1,4 @@
...
+They are not guaranteed to be (and are currently not) lists, so the new documentation describes them as iterables.
+
+Documented `universe` constructor parameter to QuantumGraph.
```
timj, line 4:
```suggestion
Documented `universe` constructor parameter to `QuantumGraph`.
```

python/lsst/pipe/base/graph/graph.py
kfindeisen: Unfortunately, it looks like _all_ the class properties have this problem... 😞

TallJimbo: Feel free to leave it, then.  I'll probably be doing some work in QG in the not-to-distant future myself and can deal with it then.

kfindeisen: I'm willing to clean up these docstrings myself. However, I'm not sure what to do about the types. For example, [allDatasetTypes](https://pipelines.lsst.io/v/daily/py-api/lsst.pipe.base.QuantumGraph.html#lsst.pipe.base.QuantumGraph.allDatasetTypes) returns a tuple of `lsst.pipe.base.graph._impleDetails.DatasetTypeName`, which, needless to say, is not a public type and does not appear in the Science Pipelines docs.

TallJimbo: @natelust, I'm guessing that's a NewType or something that should just appear as `str` in the docs?

kfindeisen: Oops, so it is. Never mind.

kfindeisen:
@TallJimbo If you'd like, please take a look at da64ee71a7a11bd4146f72326dc5ea93ccb0e6c3 to make sure I haven't garbled anything.

TallJimbo (approved):
Additional doc fixes look good, thanks!

PR #321, opened 2023-04-21
DM-38814: Drop obscore manager from config when making execution butler

andy-slac:
The fix is one additional line `executionButlerBuilder.py`, the rest is
a new unit test to reproduce the problem with execution butler and obscore.

Also needs lsst/daf_butler#826

## Checklist

- [X] ran Jenkins


timj (approved):
Thanks

PR #323, opened 2023-04-27
DM-38780: Update graph builder to always produce resolved refs

andy-slac:
This commit also fixes a bug in graph builder which produced un-connected
graphs in case of existing intermediate datasets.

Add method `QuantumGraph.updateRun`. This new method updates run collection name and dataset IDs for all output
and intermediate datasets in a graph.

## Checklist

- [X] ran Jenkins


tests/test_graphBuilder.py
```diff
@@ -62,13 +62,16 @@ def testDefault(self):
             _, qgraph2 = simpleQGraph.makeSimpleQGraph(
                 butler=butler, datasetQueryConstraint=constraint, callPopulateButler=False
             )
+            # When all outputs are refs resolved and random direct comparison
```
TallJimbo, line 65:
```suggestion
            # When all outputs are random resolved refs, direct comparison
```

tests/test_graphBuilder.py
```diff
@@ -62,13 +62,16 @@ def testDefault(self):
...
+            # of graphs does not work because IDs are different. Can only
+            # verify the number of qunata in the graph without doing something
```
TallJimbo, line 67:
```suggestion
            # verify the number of quanta in the graph without doing something
```

TallJimbo (approved):

python/lsst/pipe/base/graph/graph.py
```diff
@@ -1262,6 +1262,70 @@ def iterTaskGraph(self) -> Generator[TaskDef, None, None]:
         """
         yield from nx.topological_sort(self.taskGraph)
 
+    def updateRun(self, run: str, *, metadata_key: str | None = None, update_graph_id: bool = False) -> None:
```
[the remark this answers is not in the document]
andy-slac: Higher level code that fills metadata is in [ctrl_mpexec.CmdLienFwk](https://github.com/lsst/ctrl_mpexec/blob/main/python/lsst/ctrl/mpexec/cmdLineFwk.py#L573-L584). I'm more comfortable with having a default here that does not depend on high-level code, and have ctrl_mpexec (which I think will be the only client of this method) to pass the correct value here.

We could potentially define a set of common metadata keys in pipe_base so that they can be usable by other clients, or even make metadata an actual class with a fixed list of common attributes (and optional dynamic mapping).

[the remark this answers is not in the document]
TallJimbo: Yeah, if it's CmdLineFwk setting the value in both cases, that's fine, too.

python/lsst/pipe/base/graphBuilder.py
```diff
@@ -1123,18 +1149,19 @@ def resolveDatasetRefs(
...
-                    unresolvedRefs = []
+                    unresolvedDataIds = []
                     haveMetadata = False
                     for datasetType, originalRefs in quantum.outputs.items():
-                        for ref in task.outputs.extract(datasetType, originalRefs.keys()):
-                            if ref.id is not None:
+                        for dataId, ref in task.outputs.extract(datasetType, originalRefs.keys()):
+                            if ref is not None:
                                 resolvedRefs.append(ref)
+                                originalRefs[dataId].ref = ref
```
andy-slac, line 1158: @TallJimbo, this is the line that fixes disconnected graphs (would read `originalRefs[dataId] = ref` in old code). With _RefHolder thing in this new implementation only the pure outputs need this line, intermediates are updated because _refHolders are actually shared between quanta. In the old code refs were not shared, so this line would also update intermediates.

TallJimbo: Ah, thanks.

And now that you've pointed me at this, I see a comment a file lines up from this point in the diff that mentions refs that are "never resolved"; that should get updated here.

andy-slac: Hmm, I never read the comments 🙁 I probably need to update that comment too, it's confusing now.

andy-slac: I removed that sentence.

PR #324, opened 2023-04-28
DM-38948: Fix handling of dataset query constraint options in QG generation.

TallJimbo:
...Probably.  It was definitely broken by DM-38780, but there are still no tests to show that it works now.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/pipe/base/graphBuilder.py
+++ b/python/lsst/pipe/base/graphBuilder.py
@@ -301,6 +301,28 @@ class _DatasetDict(NamedKeyDict[DatasetType, dict[DataCoordinate, _RefHolder]]):
         refs = self[datasetType]
         return ((dataId, refs[dataId].ref) for dataId in dataIds)
 
+    def isdisjoint(self, other: _DatasetDict) -> bool:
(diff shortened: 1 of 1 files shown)


TallJimbo:
I have not added a changelog entry because I figure we'll get this in pretty soon, and hence aside from a daily or two it'll be as if nothing happened.  Happy to entertain other perspectives on that.

python/lsst/pipe/base/graphBuilder.py
```diff
@@ -1072,9 +1103,8 @@ def resolveDatasetRefs(
                         )
 
         # Look up input and initInput datasets in the input collection(s).
-        # container to accumulate unfound refs, if the common dataIs were not
-        # constrained on dataset type existence.
-        self.unfoundRefs = set()
+        # container to accumulate datasets in self.missing, if the common
+        # dataIds were not constrained on dataset type existence.
```
andy-slac, line 1107: "container to" is probably not needed here?

andy-slac (approved):
Looks good.

PR #326, opened 2023-04-29
DM-38779: Use graph DatasetRef in execution butler

timj:
This attempts to ensure we have provenance agreement between the graph and the execution butler.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/pipe/base/executionButlerBuilder.py
+++ b/python/lsst/pipe/base/executionButlerBuilder.py
@@ -390,7 +390,7 @@ def buildExecutionButler(
 
     Parameters
     ----------
-    butler : `lsst.daf.butler.Bulter`
+    butler : `lsst.daf.butler.Butler`
         This is the existing `~lsst.daf.butler.Butler` instance from which
         existing datasets will be exported. This should be the
         `~lsst.daf.butler.Butler` which was used to create any `QuantumGraphs`
@@ -445,6 +445,10 @@ def buildExecutionButler(
     NotADirectoryError
         Raised if specified output URI does not correspond to a directory
     """
+    # Now require that if run is given it must match the graph run.
+    if run and graph.metadata and run != (graph_run := graph.metadata.get("output_run")):
+        raise ValueError(f"The given run, {run!r}, does not match that specified in the graph, {graph_run!r}")
+
     # We know this must refer to a directory.
     outputLocation = ResourcePath(outputLocation, forceDirectory=True)
     if datastoreRoot is not None:
--- a/python/lsst/pipe/base/executionButlerBuilder.py
+++ b/python/lsst/pipe/base/executionButlerBuilder.py
@@ -306,7 +306,6 @@ def _setupNewButler(
     # file data stores continue to look at the old location.
     config = Config(butler._config)
     config["root"] = outputLocation.geturl()
-    config["allow_put_of_predefined_dataset"] = True
     config["registry", "db"] = "sqlite:///<butlerRoot>/gen3.sqlite3"
 
     # Remove any namespace that may be set in main registry.
--- a/python/lsst/pipe/base/executionButlerBuilder.py
+++ b/python/lsst/pipe/base/executionButlerBuilder.py
@@ -23,11 +23,10 @@ from __future__ import annotations
 __all__ = ("buildExecutionButler",)
 
 import io
-import itertools
 from collections import defaultdict
-from typing import Callable, DefaultDict, Iterable, List, Mapping, Optional, Set, Tuple, Union
+from typing import Callable, Iterable, List, Mapping, Optional, Set, Tuple, Union
 
-from lsst.daf.butler import Butler, Config, DataCoordinate, DatasetRef, DatasetType, Registry
+from lsst.daf.butler import Butler, Config, DatasetRef, DatasetType, Registry
 from lsst.daf.butler.core.repoRelocation import BUTLER_ROOT_TAG
 from lsst.daf.butler.registry import ConflictingDefinitionError, MissingDatasetTypeError
 from lsst.daf.butler.transfers import RepoExportContext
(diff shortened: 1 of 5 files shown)


python/lsst/pipe/base/executionButlerBuilder.py
```diff
@@ -150,7 +152,7 @@ def _accumulate(
         if dataset_type.component() is not None:
             dataset_type = dataset_type.makeCompositeDatasetType()
         dataset_type = _validate_dataset_type(dataset_type, datasetTypes, butler.registry)
-        inserts[dataset_type].add(DataCoordinate.makeEmpty(dataset_type.dimensions.universe))
+        init_inserts[dataset_type].add(DataCoordinate.makeEmpty(dataset_type.dimensions.universe))
```
timj, line 155: I'm a bit confused about the initOutputs. Do they have resolved refs in the graph? When I ran tests for this in `pipelines_check` it seemed that `icSrc_schema` was in the outputs as well and so was picked up below explicitly but `packages` and `isr_config` was only here.

TallJimbo: Yes, `icSrc_schema` is an `initOutput` of `characterizeImage`, and an `initInput` of `calibrate`>

timj: Okay. So I'm assuming `packages` and `*_config` must be in the graph. I think this part is only working because we are still unresolving on put.

timj:
@andy-slac / @TallJimbo I think we need to work out what to do with this PR and the related one in lsst/ctrl_mpexec#234.

I think the change in execution butler to use the graph DatasetRef is fairly uncontroversial so maybe we do this for now and leave the second half of the PR.

The controversial part is that I now force `ButlerQuantumContext.put` to use the graph dataset ref (effectively becoming a `putDirect`). (I also make `SingleQuantumExecutor` do this as well). This allows us to remove the flag in execution butler that requires we check registry for a predefined dataset. What concerns me is that we might run into problems with people calling `runQuantum` repeatedly and tweaking the butler `run`. I don't know if anyone does that except maybe in test code. `pipelines_check` and `ctrl_mpexec` tests pass.

python/lsst/pipe/base/executionButlerBuilder.py
```diff
@@ -339,7 +340,7 @@ def _setupNewButler(
 def _import(
     yamlBuffer: io.StringIO,
     newButler: Butler,
-    inserts: DataSetTypeMap,
+    inserts: DataSetTypeRefMap,
     run: Optional[str],
```
timj, line 343: We no longer use this parameter and `buildExecutionButler` also needs to be documented to ignore the `run` parameter. We now require that it's all matching the graph.

andy-slac:
Would it be reasonable option for now to execute `graph.updateRun()` if people provide an explicit output run that is different from graph run (and give plenty of warnings maybe)?

timj:
> Would it be reasonable option for now to execute graph.updateRun()

Where? We can't do it inside execution butler creation because then we would get a ID mismatch if you try to run the pipeline with that graph and that execution butler. I think we have to assume that `buildExecutionButler` is effectively an internal API that is only called by `makeGraph` rather than try to deal with people using the API directly and wanting the run to be inconsistent. I think the best thing is for me to check that `run` matched the graph run and raise if they don't match.

andy-slac:
I was not thinking about execution butler, but rather about individual users trying to re-run their favorite graph with different output runs.

TallJimbo:
That "controversial" change sounds great to me; I think it's a big step towards having SingleQuantumExecutor and ButlerQuantumContext only use the LimitedButler interface without any special-casing for full Butler.

But I bet we have to fix datasetExists to get all the way there, at least for SQE.

python/lsst/pipe/base/_dataset_handle.py
```diff
@@ -49,7 +50,7 @@ def __init__(
         *,
         storageClass: StorageClass | None = None,
         parameters: dict[str, Any] | None = None,
-        dataId: dict[str, Any] | DataCoordinate | None = None,
+        dataId: dict[str, Any] | DataCoordinate | Mapping | None = None,
```
andy-slac, line 53: I guess this is not to allow arbitrary `Mapping[Any, Any]` but just to remove `type: ignore`? Could it be made `Mapping[str, Any]`? You could probably drop `dict` then.

python/lsst/pipe/base/executionButlerBuilder.py
```diff
@@ -141,16 +138,19 @@ def _accumulate(
...
     datasetTypes: dict[Union[str, DatasetType], DatasetType] = {}
 
-    # Add inserts for initOutputs (including initIntermediates); these are
-    # defined fully by their DatasetType, because they have no dimensions.
-    # initInputs are part of Quantum and that's the only place the graph stores
-    # the dataset IDs, so we process them there even though each Quantum for a
-    # task has the same ones.
-    for dataset_type in itertools.chain(dataset_types.initIntermediates, dataset_types.initOutputs):
+    # Find the initOutput refs.
+    initOutputRefs = [ref for ref in graph.globalInitOutputRefs()]
```
andy-slac, line 142: `list(graph.globalInitOutputRefs())` is slightly shorter 🙂

andy-slac (approved):
Looks good

python/lsst/pipe/base/_dataset_handle.py
timj: I added `Mapping` here because `frozendict` is a `Mapping`. I hadn't spotted that I already had `dict` in there so I can replace that with Mapping.

timj: I changed it to `Mapping[Any, Any]` and dropped the separate `dict`.

timj: I'm so slow catching up. Sorry. I see that that is exactly what `DataId` is defined as (without the `None`).

PR #327, opened 2023-05-01
DM-38957: Ignore new data IDs in QG gen follow-up queries.

TallJimbo:
These data IDs that were not present in the initial data ID query can come about because we have to drop the post-query region-filtering when doing the follow-up queries for datasets.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/pipe/base/graphBuilder.py
+++ b/python/lsst/pipe/base/graphBuilder.py
@@ -1097,7 +1097,8 @@ class _PipelineScaffolding:
                     except MissingDatasetTypeError:
                         resolvedRefQueryResults = []
                     for resolvedRef in resolvedRefQueryResults:
-                        assert resolvedRef.dataId in refs
+                        if resolvedRef.dataId not in refs:
+                            continue
                         refs[resolvedRef.dataId].ref = (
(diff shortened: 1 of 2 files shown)


python/lsst/pipe/base/graphBuilder.py
```diff
@@ -1097,7 +1097,8 @@ def resolveDatasetRefs(
                     except MissingDatasetTypeError:
                         resolvedRefQueryResults = []
                     for resolvedRef in resolvedRefQueryResults:
-                        assert resolvedRef.dataId in refs
+                        if resolvedRef.dataId not in refs:
```
andy-slac, line 1100: Maybe catching `KeyError` in this case would be more efficient, avoiding searching the map twice, but it probably does not matter.

andy-slac (approved):
Looks good.

python/lsst/pipe/base/graphBuilder.py
TallJimbo: I've seen some examples of Python dict performance on these sorts of questions being completely counterintuitive, so while I do try to avoid double-lookups (usually with `if (x := d.get(k)) is not None`, these days), that's mostly for consistency rather than any actual expectation of performance.

PR #328, opened 2023-05-03
DM-38694: Add support for copying on return with InMemoryDatasetHandle

timj:
Requires lsst/daf_butler#819

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/pipe/base/_dataset_handle.py
+++ b/python/lsst/pipe/base/_dataset_handle.py
@@ -26,7 +26,7 @@ import dataclasses
 from typing import Any, Optional
 
 from frozendict import frozendict
-from lsst.daf.butler import DataCoordinate, DimensionUniverse, StorageClass, StorageClassFactory
+from lsst.daf.butler import DataCoordinate, DataId, DimensionUniverse, StorageClass, StorageClassFactory
 
 
 # Use an empty dataID as a default.
@@ -49,7 +49,7 @@ class InMemoryDatasetHandle:
         *,
         storageClass: StorageClass | None = None,
         parameters: dict[str, Any] | None = None,
-        dataId: dict[str, Any] | DataCoordinate | None = None,
+        dataId: DataId | None = None,
         **kwargs: Any,
     ):
         object.__setattr__(self, "inMemoryDataset", inMemoryDataset)
@@ -59,7 +59,7 @@ class InMemoryDatasetHandle:
         # This will not be a full DataCoordinate.
         if dataId is None:
             if kwargs:
-                dataId = frozendict(kwargs)  # type: ignore
+                dataId = frozendict(kwargs)
             else:
                 dataId = DataCoordinate.makeEmpty(DimensionUniverse())
         elif kwargs:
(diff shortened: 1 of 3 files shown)


python/lsst/pipe/base/_dataset_handle.py
```diff
@@ -43,32 +50,36 @@ class InMemoryDatasetHandle:
     parameters will be converted into a dataId-like entity.
     """
 
+    _empty = DataCoordinate.makeEmpty(DimensionUniverse())
```
erykoff, line 53: You said on the ticket that this is a slow operation.  How often is this run now that it's in the class?

python/lsst/pipe/base/_dataset_handle.py
```diff
@@ -135,11 +146,31 @@ class can be found.
...
 
+        inMemoryDataset = self.inMemoryDataset
+
+        if self.copy:
+            # An optimization might be to defer copying until any components
+            # and parameters have been applied. This can be a problem since
+            # most component storage classes do not bother to define a
+            # storage class delegate and the default delegate uses deepcopy()
+            # which can fail if explicit support for deepcopy() is missing
+            # or pickle does not work (for example VisitInfo fails).
```
erykoff, line 157: Is this saying that if you put a VisitInfo in an inMemoryDatasetHandle and use copy=True it will raise an exception on get?

erykoff (approved):

python/lsst/pipe/base/_dataset_handle.py
timj: Once. It's a class property so defined at class definition. @TallJimbo I do wonder why `makeEmpty` is so slow -- `DimensionUniverse()` is meant to cache so I wasn't expecting there to be overhead in that call but maybe there is overhead I wasn't expecting and caching the universe would have been enough.

python/lsst/pipe/base/_dataset_handle.py
timj: Yes. Because you can't pickle a VisitInfo.

```python
>>> copy.deepcopy(VisitInfo())
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/timj/work/lsstsw/miniconda/envs/lsst-scipipe-6.0.0/lib/python3.10/copy.py", line 161, in deepcopy
    rv = reductor(4)
TypeError: cannot pickle 'lsst.afw.image._imageLib.VisitInfo' object
```

It looks like `VisitInfo` is immutable so I can add a `__deepcopy__` for it that returns itself.

timj: My feeling is that you are using `InMemoryDatasetHandle` in new code and will (1) know if you need to copy it or not (2) write a new delegate copy subclass if you really do need to copy it some special way, or (3) add a `__deepcopy__` method

timj: I have added `VisitInfo.__deepcopy__` in lsst/afw#686

python/lsst/pipe/base/_dataset_handle.py
TallJimbo: Looks like the caching remembers all of the `DimensionUniverse` instances that exist, but it doesn't record which one is the default, and I think that means it has to go read the default config file to find at least the version of the default one before it can look for it in the cache.  We could clearly make that more efficient.

PR #329, opened 2023-05-09
DM-39100 Move PipelineTask Config override handling

natelust:
Move the handling of PipelineTaskConfig override handling into the Config class and out of the Pipeline. This allows Config classes more control over how overrides are applied, and a place to add a hook to react to config values.

## Checklist

- [ ] ran Jenkins


python/lsst/pipe/base/config.py
```diff
@@ -29,15 +29,19 @@
 # -------------------------------
 #  Imports of standard modules --
 # -------------------------------
+import os
 from numbers import Number
-from typing import TYPE_CHECKING, Any, Dict, Optional, Tuple, Type, TypeVar
+from typing import TYPE_CHECKING, Any, Dict, Iterable, Optional, Tuple, Type, TypeVar
```
TallJimbo, line 34:
```suggestion
from collections.abc import Iterable
from typing import TYPE_CHECKING, Any, Dict, Optional, Tuple, Type, TypeVar
```
You don't need to all the way in terms of modernizing type annotations here, but this much should already be consistent with the rest of the module and only involves the new symbol you've imported.

python/lsst/pipe/base/config.py
```diff
@@ -199,6 +203,54 @@ class to allow configuration of the connections class. This dynamically
...
+        Parameters
+        ---------
+        instrument : `Instrument` or `None`
+            If this is not None, this is an `Instrument` instance which
+            will be used to apply instrumental overrides.
+        taskDefaultName : `str`
+            The default name associated with the `Task` class. This is
+            used with instrumental overrides.
+        pipelineConfigs : `Iterable` of `ConfigIR`
+            This is an iterable of `ConfigIR` objects that contain
```
TallJimbo, line 225:
```suggestion
            An iterable of `ConfigIR` objects that contain
```

python/lsst/pipe/base/config.py
```diff
@@ -199,6 +203,54 @@ class to allow configuration of the connections class. This dynamically
...
+                    )
+                # only apply override if it applies to everything
+                if subConfig.dataId is None:
+                    if subConfig.file:
+                        for configFile in subConfig.file:
+                            overrides.addFileOverride(os.path.expandvars(configFile))
+                    if subConfig.python is not None:
+                        overrides.addPythonOverride(subConfig.python)
+                    for key, value in subConfig.rest.items():
+                        overrides.addValueOverride(key, value)
```
TallJimbo, line 251: I'm guessing this is unchanged, but I would not have expected these overrides to be applied after the `python:` block when both exist, though I don't have a solid reason for that expectation.

TallJimbo (approved):
Please do add a changelog entry, otherwise looks good.

python/lsst/pipe/base/config.py
natelust: Value overrides are always last, as python block have the most side effects, importing things, retargeting, etc

PR #335, opened 2023-05-15
DM-39198: Ensure task connection storage classes are used in Quanta.

TallJimbo:
QuantumGraph should have the connection storage classes in its Quantum objects, and the registry storage classes for those dataset types in its "registryDatasetTypes" attribute.  Apparently we were making no effort to ensure the former before.

## Checklist

- [x] ran [Jenkins](https://ci.lsst.codes/blue/organizations/jenkins/stack-os-matrix/detail/stack-os-matrix/38768/pipeline)

### DIFF
--- a/python/lsst/pipe/base/graphBuilder.py
+++ b/python/lsst/pipe/base/graphBuilder.py
@@ -269,7 +269,7 @@ class _DatasetDict(NamedKeyDict[DatasetType, dict[DataCoordinate, _RefHolder]]):
 
     def unpackMultiRefs(self, storage_classes: dict[str, str]) -> NamedKeyDict[DatasetType, list[DatasetRef]]:
         """Unpack nested multi-element `DatasetRef` dicts into a new
-        mapping with `DatasetType` keys and `set` of `DatasetRef` values.
+        mapping with `DatasetType` keys and `list` of `DatasetRef` values.
 
         Parameters
         ----------
--- a/python/lsst/pipe/base/graphBuilder.py
+++ b/python/lsst/pipe/base/graphBuilder.py
@@ -55,12 +55,12 @@ from lsst.daf.butler.registry.queries import DataCoordinateQueryResults
 from lsst.daf.butler.registry.wildcards import CollectionWildcard
 from lsst.utils import doImportType
 
-from ._datasetQueryConstraints import DatasetQueryConstraintVariant
-from ._status import NoWorkFound
-
 # -----------------------------
 #  Imports for other modules --
 # -----------------------------
+from . import automatic_connection_constants as acc
+from ._datasetQueryConstraints import DatasetQueryConstraintVariant
+from ._status import NoWorkFound
 from .connections import AdjustQuantumHelper, iterConnections
 from .graph import QuantumGraph
 from .pipeline import Pipeline, PipelineDatasetTypes, TaskDatasetTypes, TaskDef
@@ -242,12 +242,20 @@ class _DatasetDict(NamedKeyDict[DatasetType, dict[DataCoordinate, _RefHolder]]):
             return base
         return base.union(*[datasetType.dimensions for datasetType in self.keys()])
 
-    def unpackSingleRefs(self) -> NamedKeyDict[DatasetType, DatasetRef]:
+    def unpackSingleRefs(self, storage_classes: dict[str, str]) -> NamedKeyDict[DatasetType, DatasetRef]:
         """Unpack nested single-element `DatasetRef` dicts into a new
         mapping with `DatasetType` keys and `DatasetRef` values.
 
         This method assumes that each nest contains exactly one item, as is the
         case for all "init" datasets.
+
+        Parameters
+        ----------
+        storage_classes : `dict` [ `str`, `str` ]
+            Mapping from dataset type name to the storage class to use for that
+            dataset type.  These are typically the storage classes declared
+            for a particular task, which may differ rom the data repository
+            definitions.
 
         Returns
         -------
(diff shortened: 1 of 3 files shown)


python/lsst/pipe/base/graphBuilder.py
```diff
@@ -242,42 +242,60 @@ def dimensions(self) -> DimensionGraph:
...
         mapping with `DatasetType` keys and `DatasetRef` values.
 
         This method assumes that each nest contains exactly one item, as is the
         case for all "init" datasets.
 
+        Parameters
+        ----------
+        storage_classes : `dict` [ `str`, `str` ]
+            Mapping from dataset type name to the storage class to use for that
+            dataset type.  These are is typically the storage classes declared
```
timj, line 256:
```suggestion
            dataset type.  These are typically the storage classes declared
```

python/lsst/pipe/base/graphBuilder.py
```diff
@@ -242,42 +242,60 @@ def dimensions(self) -> DimensionGraph:
...
-    def unpackMultiRefs(self) -> NamedKeyDict[DatasetType, list[DatasetRef]]:
+    def unpackMultiRefs(self, storage_classes: dict[str, str]) -> NamedKeyDict[DatasetType, list[DatasetRef]]:
         """Unpack nested multi-element `DatasetRef` dicts into a new
         mapping with `DatasetType` keys and `set` of `DatasetRef` values.
 
+        Parameters
+        ----------
+        storage_classes : `dict` [ `str`, `str` ]
+            Mapping from dataset type name to the storage class to use for that
+            dataset type.  These are is typically the storage classes declared
```
timj, line 278:
```suggestion
            dataset type.  These are typically the storage classes declared
```

python/lsst/pipe/base/graphBuilder.py
```diff
@@ -242,42 +242,60 @@ def dimensions(self) -> DimensionGraph:
...
-        def getOne(refs: dict[DataCoordinate, _RefHolder]) -> DatasetRef:
-            (holder,) = refs.values()
-            return holder.resolved_ref
-
-        return NamedKeyDict({datasetType: getOne(refs) for datasetType, refs in self.items()})
-
-    def unpackMultiRefs(self) -> NamedKeyDict[DatasetType, list[DatasetRef]]:
+    def unpackMultiRefs(self, storage_classes: dict[str, str]) -> NamedKeyDict[DatasetType, list[DatasetRef]]:
         """Unpack nested multi-element `DatasetRef` dicts into a new
         mapping with `DatasetType` keys and `set` of `DatasetRef` values.
```
timj, line 272:
```suggestion
        mapping with `DatasetType` keys and `list` of `DatasetRef` values.
```
?

python/lsst/pipe/base/graphBuilder.py
```diff
@@ -381,7 +399,10 @@ def __repr__(self) -> str:
     quantum.
     """
 
-    def makeQuantum(self, datastore_records: Optional[Mapping[str, DatastoreRecordData]] = None) -> Quantum:
+    def makeQuantum(
```
timj, line 402: Why did this get reformatted? I can't see what changed. Was it reformatted because of an early API modification that was backed out?

timj (approved):

python/lsst/pipe/base/graphBuilder.py
TallJimbo: Yes, that's exactly it.  Black vs. the trailing commas.

python/lsst/pipe/base/script/transfer_from_graph.py
```diff
@@ -62,13 +62,26 @@ def transfer_from_graph(
...
-            output_refs.update(refs)
+            original_output_refs.update(refs)
     for qnode in qgraph:
         for refs in qnode.quantum.outputs.values():
-            output_refs.update(refs)
+            original_output_refs.update(refs)
+
+    # Get data repository definitions from the QuantumGraph; these can have
+    # different storage classes than those in the quanta.
+    dataset_types = {dstype.name: dstype for dstype in qgraph.registryDatasetTypes()}
```
timj, line 75: I see this line enough I'm starting to wonder if QuantumGraph could have a method that returned the dict.

python/lsst/pipe/base/script/transfer_from_graph.py
```diff
@@ -62,13 +62,26 @@ def transfer_from_graph(
...
+
+    # Convert output_refs to the data repository storage classes, too.
+    output_refs = set()
+    for ref in original_output_refs:
+        internal_dataset_type = dataset_types.get(ref.datasetType.name, ref.datasetType)
+        if internal_dataset_type.storageClass_name != ref.datasetType.storageClass_name:
+            output_refs.add(ref.overrideStorageClass(internal_dataset_type.storageClass_name))
```
timj, line 82:
```suggestion
            output_refs.add(ref.overrideStorageClass(internal_dataset_type.storageClass))
```
since the storage class should be attached and can be used directly without having to proxy through the name string.

timj (approved):

python/lsst/pipe/base/script/transfer_from_graph.py
TallJimbo: :+1: .  I'm informally saving up a number of changes I'd like to make to QG.

python/lsst/pipe/base/script/transfer_from_graph.py
TallJimbo: If you think that's safe, sure.  I tend to default to using the name when I'm not sure because I don't want to accidentally trigger imports of things that might not be available, but that may just be a habit relevant for working down in the guts of registry, not here.

timj: I think at this point the storage class is going to be resolved into a real storage class almost immediately so it won't matter. I guess there's no harm in going via the name here and then deferring sotrage class creation as much as possible.

PR #341, opened 2023-06-05
DM-39377: Deprecate `PipelineTaskConfig.saveMetadata` field

andy-slac:
Also removes unused `ResourceConfig` class.

## Checklist

- [X] ran Jenkins


python/lsst/pipe/base/config.py
```diff
@@ -200,6 +200,7 @@ class to allow configuration of the connections class. This dynamically
         default=True,
         optional=False,
         doc="Flag to enable/disable metadata saving for a task, enabled by default.",
+        deprecated="This field is deprecated and will be removed after v25.",
```
timj, line 203: I think it has to b v26 because v25 is already being worked on from earlier in the year.

timj (commented):
Don't we also have to stop using the config in pipeline.py?

andy-slac:
> Don't we also have to stop using the config in pipeline.py?

Hmm, I was thinking that we'll be doing some cleanup when we actually drop it completely. But I can certainly update things now to assume that saveMetadata is always true (ctrl_mpexec is the other place with that logic).

timj:
I think if we deprecate a config we are meant to stop using it.

timj (approved):

timj:
I guess the change note should mention explicitly that the saveMetadata config is now ignored.

PR #344, opened 2023-06-15
DM-39661: Rename ButlerQuantumContext to QuantumContext

timj:
## Checklist

- [x] ran Jenkins


python/lsst/pipe/base/_quantumContext.py
```diff
@@ -39,8 +43,79 @@
...
+        kwargs : `dict`
+            Keyword arguments to be used when pickling.
+        """
+        kwargs: dict[str, Any] = {"num_cores": self.num_cores}
+        if self.max_mem is not None:
+            # .value is a numpy float. Cast it to a python int since we
+            # do not want fractional bytes.
+            # __post_init__ ensures that this is a Quantity but mypy cannot
+            # work that out.
+            kwargs["max_mem"] = int(self.max_mem.value)  # type: ignore[union-attr]
```
andy-slac, line 92: Would it be better to declare `self.max_mem` as `u.Quantity | None` and provide a factory method from `max_mem: int`? Then mypy will be happy and you don't need `__post_init__`.

python/lsst/pipe/base/_quantumContext.py
```diff
@@ -49,10 +124,12 @@ class ButlerQuantumContext:
     quantum : `lsst.daf.butler.core.Quantum`
         Quantum object that describes the datasets which will be get/put by a
         single execution of this node in the pipeline graph.
+    resources : `ExecutionResources`
```
andy-slac, line 127:
```suggestion
    resources : `ExecutionResources`, optional
```

andy-slac (approved):
Looks good.

python/lsst/pipe/base/_quantumContext.py
timj: Oh, you mean have `ExecutionResources.from_dict(kwargs: dict[str, Any])` or something? (having trouble with the name). I guess I can move the logic from cmdLineFwk here so long as there is a default unit specified for the plain float/int in this constructor. I still want the pickling to store the plain int but I can change the pickle factory to use the new method.

andy-slac: MAybe something like `ExecutionResources.from_bytes(num_proc: int, mem_bytes: int)`?

timj: If I'm having a factory method I may as well accept strings as well and then it seems odd to have one per type of `max_mem` value. If this factory takes `str|int|Quantity` and converts to Quantity then the ctrl_mpexec code only ever has to call that one API and doesn't care where the max_mem value came from.

timj: @andy-slac I've implemented the `__init__` approach as discussed on Slack. If you want to look you can since I'm going to have to run jenkins again.

python/lsst/pipe/base/_quantumContext.py
```diff
@@ -39,8 +43,120 @@
...
+
+        Returns
+        -------
+        kwargs : `dict`
+            Keyword arguments to be used when pickling.
+        """
+        kwargs: dict[str, Any] = {"num_cores": self.num_cores}
+        if self.max_mem is not None:
+            # .value is a numpy float. Cast it to a python int since we
+            # do not want fractional bytes. __post_init__ ensures that this
```
andy-slac, line 131: __post_init__ -> __init__?

andy-slac (approved):
Looks good.

python/lsst/pipe/base/_quantumContext.py
timj: Yeah, I copied the code from the factory version.

PR #348, opened 2023-06-27
DM-39582: Decrease memory usage and load times when reading graphs

natelust:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/doc/changes/DM-39582.api.md
+++ b/doc/changes/DM-39582.api.md
@@ -1,1 +1,0 @@
-Deprecated reconstituteDimensions argument from QuantumNode.from_simple
--- a/doc/changes/DM-39582.api.md
+++ b/doc/changes/DM-39582.api.md
@@ -1,0 +1,1 @@
+Deprecated reconstituteDimensions argument from QuantumNode.from_simple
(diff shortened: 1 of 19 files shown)


doc/changes/DM-39582.api.md
```diff
@@ -0,0 +1 @@
+Deprecated reconstituteDimensions argument from QuantumNode.from_simple
```
andy-slac, line 1: Probably `removal.md` instead of `api.md`?

python/lsst/pipe/base/graph/_loadHelpers.py
```diff
@@ -219,7 +220,11 @@ def load(
         _readBytes = self._readBytes
         if universe is None:
             universe = headerInfo.universe
-        return self.deserializer.constructGraph(nodeSet, _readBytes, universe)
+        # use the daf butler context vars to aid in ensuring deduplication in
+        # object instantiation.
+        runner = PersistenceContextVars()
+        graph = runner.run(self.deserializer.constructGraph, nodeSet, _readBytes, universe)
+        return graph  # type: ignore
```
andy-slac, line 227: I do not entirely understand PersistenceContextVars/context vars, would it be nice if it could work like a regular context manager, e.g.:

    with PersistenceContextVars():
       return self.deserializer.constructGraph(nodeSet, _readBytes, universe)

but may be it's not possible?

andy-slac (approved):
Looks good.

python/lsst/pipe/base/graph/_loadHelpers.py
natelust: I worked on that for a while, and the outcome kept coming out more awkward, and not at all nice like that. I think it is mostly because context vars are more geared toward asyncio, callback programming style.

PR #349, opened 2023-06-30
DM-39227: Add support for deprecating connections.

TallJimbo:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/pipe/base/connectionTypes.py
+++ b/python/lsst/pipe/base/connectionTypes.py
@@ -50,12 +50,19 @@ class BaseConnection:
         consistent (i.e. zip-iterable) in `PipelineTask.runQuantum()` and
         notify the execution system as early as possible of outputs that will
         not be produced because the corresponding input is missing.
+    deprecated : `str`, optional
+        A description of why this connection is deprecated, including the
+        version after which it may be removed.
+
+        If not `None`, the string is appended to the docstring for this
+        connection and the corresponding config Field.
     """
 
     name: str
     storageClass: str
     doc: str = ""
     multiple: bool = False
+    deprecated: str | None = dataclasses.field(default=None, kw_only=True)
 
     _connection_type_set: ClassVar[str]
 
--- a/doc/changes/DM-39227.feature.md
+++ b/doc/changes/DM-39227.feature.md
@@ -1,0 +1,1 @@
+Make it possible to deprecate `PipelineTask` connections.
--- a/python/lsst/pipe/base/config.py
+++ b/python/lsst/pipe/base/config.py
@@ -146,7 +146,10 @@ class PipelineTaskConfigMeta(pexConfig.ConfigMeta):
                 docString = "Template parameter used to format corresponding field template parameter"
                 for templateName, default in connectionsClass.defaultTemplates.items():
                     configConnectionsNamespace[templateName] = TemplateField(
-                        dtype=str, doc=docString, default=default
+                        dtype=str,
+                        doc=docString,
+                        default=default,
+                        deprecated=connectionsClass.deprecatedTemplates.get(templateName),
                     )
             # add a reference to the connection class used to create this sub
             # config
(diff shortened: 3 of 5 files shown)


python/lsst/pipe/base/connectionTypes.py
```diff
@@ -57,6 +57,14 @@ class BaseConnection:
     doc: str = ""
     multiple: bool = False
 
+    deprecated: str | None = dataclasses.field(default=None, kw_only=True)
+    """A description of why this connectoin is deprecated, including the
```
timj, line 61:
```suggestion
    """A description of why this connection is deprecated, including the
```

natelust: It seems we are not even documenting `doc` above, but it seems weird to have this comment here rather than in the class level documentation like the others. Can you either move the doc to the docstring, or move the other docs to the objects? Which way does sphinx prefer? I do know that python does nothing with docstrings under the object, so interactively they just disappear.

python/lsst/pipe/base/connectionTypes.py
```diff
@@ -57,6 +57,14 @@ class BaseConnection:
     doc: str = ""
     multiple: bool = False
 
+    deprecated: str | None = dataclasses.field(default=None, kw_only=True)
+    """A description of why this connection is deprecated, including the
```
natelust, line 61: It seems a bug that we are not documenting `doc` above. It seems weird that the docs for this are here and not up with the docstring like the others. What does sphinx perfer? I know python does nothing with strings defined below an attribute and just drops them on the floor, so interactively the docs are unavailable. Can you make these docs consistent one way or another?

natelust (approved):
Is it enough to announce the deprecation in the Config class? Does it make sense to also warn in the construction of the Connections class?

python/lsst/pipe/base/connectionTypes.py
TallJimbo: Ah, I hadn't even noticed that the others had been documented earlier.  I usually think of the class doc's "Parameters" section as being for `__init__` arguments, not instance attributes, but I suppose for dataclasses there's (usually) not much distinction.  I'll just move this docstring up with the rest.

natelust: In a dataclass they are all init args unless you play with that though right?

natelust: Like I said, I am fine if you want to move them all there, but it is a bit sad because it wont show up with `help`

TallJimbo: Right.  My tendency is to add class `Parameters` to a dataclass only if I actually define an `__init__` but to always document the attributes.  I don't think that's obviously better or worse than only documenting them in class `Parameters` as you've done here, though maybe we should always be documenting both?  (Of course, duplication is then its own problem.)

TallJimbo: I have actually moved the new one to `Parameters`; I was just explaining my usual practice elsewhere.  I'd rather resolve the general question about which (if either) of us should adopt the other's practice outside this ticket, and just keep things closer to the way they are already are for now.

TallJimbo:
> Does it make sense to also warn in the construction of the Connections class?

I was originally thinking that'd have to be an unconditional warning, which would be bad, but I think I could make it warn if any deprecated connections have not been removed by the end of the Connections construction. I'll give that a try.

TallJimbo:
@natelust, two more commits here for when you have a moment:
 - 9a97e06f959d1039f583eea6e52a44e5b52d1af8 implements your suggestion to warn at connections class construction;
 - 464d4abae8128745b7cc279cb336d228fc0c9c95 adds support for deprecating the config fields that go with connection templates.

TallJimbo:
@andy-slac , I just realized that @natelust is out for a few days and hence won't see my previous comment about the last two commits here.  Would you mind filling in as reviewer for those?

TallJimbo:
Since the rest of this ticket has a longer test-and-review future ahead of it and it's logically distinct, I've moved the changes on this branch to DM-39902 and #351.

PR #355, opened 2023-07-12
DM-39944: Update Execution butler builder to use Butler._registry

andy-slac:
Depends on lsst/daf_butler#864

## Checklist

- [X] ran Jenkins

### DIFF
--- a/python/lsst/pipe/base/executionButlerBuilder.py
+++ b/python/lsst/pipe/base/executionButlerBuilder.py
@@ -240,7 +240,7 @@ def _export(butler: Butler, collections: Iterable[str] | None, inserts: DataSetT
     # export/import
     BackendClass = get_class_of(butler._config["repo_transfer_formats", "yaml", "export"])
     backend = BackendClass(yamlBuffer, universe=butler.dimensions)
-    exporter = RepoExportContext(butler.registry, butler._datastore, backend, directory=None, transfer=None)
+    exporter = RepoExportContext(butler._registry, butler._datastore, backend, directory=None, transfer=None)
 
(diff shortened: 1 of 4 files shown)


python/lsst/pipe/base/executionButlerBuilder.py
```diff
@@ -240,7 +240,7 @@ def _export(butler: Butler, collections: Iterable[str] | None, inserts: DataSetT
     # export/import
     BackendClass = get_class_of(butler._config["repo_transfer_formats", "yaml", "export"])
     backend = BackendClass(yamlBuffer, universe=butler.dimensions)
-    exporter = RepoExportContext(butler.registry, butler._datastore, backend, directory=None, transfer=None)
+    exporter = RepoExportContext(butler._registry, butler._datastore, backend, directory=None, transfer=None)
```
timj, line 243: It is looking suspiciously like `RepoExportContext` is not a public API...

timj:
Sorry about the ruff problem. Not sure how I committed such a broken noqa statement last time.

python/lsst/pipe/base/executionButlerBuilder.py
andy-slac: I think its instantiation is intended to happen inside `Butler.export()`, but execution butler does not like to use export() for some reason. I prefer to wait until we remove execution butler rater than fixing it.

TallJimbo: I consider everything but construction of `RepoExportContext` to be public API, but construction definitely is not.

TallJimbo (approved):

PR #364, opened 2023-08-04
DM-39779: text-based graph visualization of pipelines

TallJimbo:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/pipe/base/pipeline_graph/__main__.py
+++ b/python/lsst/pipe/base/pipeline_graph/__main__.py
@@ -1,0 +1,589 @@
+# This file is part of pipe_base.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (http://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This software is dual licensed under the GNU General Public License and also
+# under a 3-clause BSD license. Recipients may choose which of these licenses
+# to use; please see the files gpl-3.0.txt and/or bsd_license.txt,
+# respectively.  If you choose the GPL option then the following text applies
+# (but note that there is still no warranty even if you opt for BSD instead):
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+from __future__ import annotations
+
+__all__ = ()
+
+import argparse
+import dataclasses
+import sys
+import textwrap
+from collections.abc import Sequence
+from contextlib import ExitStack
+
+from lsst.daf.butler import Butler
+from lsst.resources import ResourcePath
+
+from ..pipeline import Pipeline
+from ._pipeline_graph import PipelineGraph, TaskImportMode
+from .visualization._options import NodeAttributeOptions
+from .visualization._show import show
+
+
+def main(argv: Sequence[str]) -> int:
+    """Run the development command-line interface.
+
+    Parameters
+    ----------
+    argv : `~collections.abc.Sequence` [ `str` ]
+        Commmand-line arguments, not including the program (typically
+        ``sys.argv[1:]``).
+
+    Notes
+    -----
+    This CLI is much more capable than pipetask's for very specific things, but
+    not as polished or user friendly.  It is intended primarily for use during
+    development of pipeline graph (especially visualization, where tests can't
+    really be used to judge the quality of the outputs, and a fast turnaround
+    is desirable).
+    """
+    parser = argparse.ArgumentParser(
+        description="Expand, resolve, and display pipelines as graphs.",
+        epilog=textwrap.dedent(
+            """
+            WARNING: This is an experimental/development command-line interface
+            that is subject to change or removal without warning. The `--show`
+            option to 'pipetask build' (with 'pipeline-graph', 'task-graph', or
+            'dataset-type-graph' as the argument) is the preferred way to
+            display text-based pipeline graphs.
+            """
+        ),
+    )
+    Arguments.add_args_to_parser(parser)
+    args = Arguments.from_parsed_args(parser.parse_args(argv))
+    pipeline_graph = read_input_pipeline(args.input_pipeline)
+    if args.resolve:
+        butler = Butler(args.resolve, writeable=False)
+        pipeline_graph.resolve(butler.registry)
+    if args.save:
+        pipeline_graph._write_uri(ResourcePath(args.save))
+    if args.show:
+        with ExitStack() as stack:
+            if args.show == "-":
+                stream = sys.stdout
+            else:
+                path = ResourcePath(args.show)
+                # ResourcePath.open doesn't have the kind of mode overloading
+                # necessary to make the return type clear enough for MyPy.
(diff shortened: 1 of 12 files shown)


python/lsst/pipe/base/pipeline_graph/visualization/_options.py
```diff
@@ -0,0 +1,103 @@
...
+    """Options for displaying dimensions.
+
+    Possible values include:
+
+    - ``"full"``: report fully-expanded dimensions.
+    - ``"concise"``: report only dimensions that are not required or implied
+      dependencies of any reported dimension.
+    - `False`: do not report dimensions at all.
+    - `None`: context-dependent default behavior.
+    """
```
TallJimbo, line 43: @leeskelvin, here's part of the answer to your question about which dimensions are shown: the default is "concise".

python/lsst/pipe/base/pipeline_graph/visualization/_formatting.py
```diff
@@ -0,0 +1,175 @@
...
+            self.deferred.append((index, style, terms))
+            return f"{description[:self.width - len(index) - 6]}...{style[0]}{index}{style[1]} "
+        return description
+
+    def format_dimensions(self, dimensions: DimensionGraph) -> str:
+        """Format the dimensions of a task or dataset type node."""
+        match self.options.dimensions:
+            case "full":
+                return str(dimensions)
+            case "concise":
```
TallJimbo, line 126: Here's the second half of the answer to the question about which dimensions are shown: the algorithm that removes unnecessary dimensions in the "concise" version.

python/lsst/pipe/base/pipeline_graph/__main__.py
```diff
@@ -0,0 +1,583 @@
...
+
+    Notes
+    -----
+    This CLI is much more capable than pipetask's for very specific things, but
+    not as polished or user friendly.  It is intended primarily for use during
+    development of pipeline graph (especially visualization, where tests can't
+    really be used to judge the quality of the outputs, and a fast turnaround
+    is desirable).
+    """
+    parser = argparse.ArgumentParser(
```
timj, line 58: Please follow the approach for non-click command lines  https://developer.lsst.io/stack/argparse-script-topic-type.html and separate arparse creation from the `main` function.

python/lsst/pipe/base/pipeline_graph/__main__.py
```diff
@@ -0,0 +1,583 @@
+# This file is part of pipe_base.
```
timj, line 1: We normally separate out the argument parser, the minimal main implementation, and the functionality into different places. This would allow the core functionality to at least be loaded in a test. There is no precedent in our code base for naming something like this `__main__.py`. This file is not even loaded by the tests to check it's okay (use ImportTestCase from lsst.utils to do that).

timj (commented):
There needs to be an entry point for this command added to pyproject.toml (I haven't looked to see how people run it from scons so I'm probably missing something very important).

I also need to understand why click can't be used.

python/lsst/pipe/base/pipeline_graph/__main__.py
TallJimbo: Will do (if we keep this; see main-PR-thread comment).

python/lsst/pipe/base/pipeline_graph/__main__.py
TallJimbo: The functionality is already clearly separated; that's the `show` function, and if there's nothing tricky about importing a `__main__` module (or an easy alternative) I'm happy to do that.

And I'm fine with making the `ArgumentParser` instance available for Sphinx purposes (per the other comment), but I think it's very important to keep bits of argument parser definition code close to the corresponding bits of minimal-main-implementation code that translates those CLI arguments to the form expected by the Python interfaces.  If you factor the other way, and keep all argument parser definition code together and keep all non-argument-parser-definition minimal-main-implementation code together instead, you cannot add structure to keep complexity under control.

TallJimbo (commented):
I'm of two minds about the future of the CLI here.  It was useful when developing the visualization code, and it'd be useful for making substantive changes to it.  I might use it now and then to actually inspect pipelines while doing other work, but if all I had was the more limited `pipetask` interface to this code I probably wouldn't really miss it.  I don't think it's worth writing exhaustive unit tests for it (in many respects, the CLI should be considered test utility code, but like the code it helps test, testing that code itself just doesn't have a good cost/benefit ratio, aside from basic things like import tests).

I don't want to try to integrate all of the CLI's functionality into `pipetask`, because adding another 20 or so function arguments [in the script](https://github.com/lsst/ctrl_mpexec/blob/8a7d0b7f272c39045684c744edef706fbbe3c794/python/lsst/ctrl/mpexec/cli/script/build.py#L29), corresponding decorators [in the commands module](https://github.com/lsst/ctrl_mpexec/blob/8a7d0b7f272c39045684c744edef706fbbe3c794/python/lsst/ctrl/mpexec/cli/cmd/commands.py#L112), and definitions for those decorators somewhere [in the options module](https://github.com/lsst/ctrl_mpexec/blob/8a7d0b7f272c39045684c744edef706fbbe3c794/python/lsst/ctrl/mpexec/cli/opt/options.py#L307), and then trying (in vain) to provide a glimmer of structure to all that by repeating all of those option names yet again [in the option groups module](https://github.com/lsst/ctrl_mpexec/blob/8a7d0b7f272c39045684c744edef706fbbe3c794/python/lsst/ctrl/mpexec/cli/opt/optionGroups.py#L85) sounds horrifying.  And in a nutshell that's why I don't want to do this or anything else in Click: unless we're just fundamentally using it wrong (which is I suspect part but not all of the problem), using decorators to define CLI options does not scale well to big, rich CLIs.

So I think the options are:
 - just delete the CLI (and I can do that in a separate commit, so it's `git revert`-able if I want it back);
 - make some slight modifications (e.g. splitting out the `ArgumentParser` generation so Sphinx can see it), and maybe move it (if it's test utility code, should it live under `tests`?), but otherwise leave it here.

Back before we ditched the `examples` directory in all of our packages, that would have been a great place for this.  That thought first made me thing, "oh, then just remove it, 'examples' was a bad idea because it wasn't maintainable".  But then I started thinking that maybe it's okay to have a _place_ to put things of dubious maintainability, as long as there's an understanding that it can just be deleted as soon as it stops working if nobody wants to fix it, because those examples do _start out_ quite useful and it feels like a shame to throw them away preemptively.

python/lsst/pipe/base/pipeline_graph/__main__.py
timj: At the moment the `lsst.utils.tests.ImportTestCase` skips trying to import `__*.py` files so maybe I should explicitly only skip the `__init__.py` files. 

https://github.com/lsst/utils/blob/main/python/lsst/utils/tests.py#L388-L389

python/lsst/pipe/base/pipeline_graph/visualization/_formatting.py
```diff
@@ -0,0 +1,175 @@
...
+        raise ValueError(f"Invalid display option for dimensions: {self.options.dimensions!r}.")
+
+    def format_task_class(self, task_class_name: str) -> str:
+        """Format the type object for a task or task init node."""
+        match self.options.task_classes:
+            case "full":
+                return task_class_name
+            case "concise":
+                return task_class_name.split(".")[-1]
+            case False:
```
arunkannawadi, line 150: I see that this is covered by the tests, so it must be correct. But I am curious why this is `False` instead of `None`?

TallJimbo: I reserved `None`  to mean something like "automatic", i.e. "turn this on or off based on other context-dependent information", so I went with `False` as the "definitely don't display this information" option.  But by the time the code gets here any `None` value will have been replaced by one of the other options.

python/lsst/pipe/base/pipeline_graph/__main__.py
TallJimbo: I've updated the test code to run `__main__.main` (passing command-line arguments as a sequence of strings), so there's no need for import-only tests here anymore.

PR #366, opened 2023-08-08
DM-40322: Add connections hooks for spatial/temporal bounds.

TallJimbo:
These will be used in prerequisite input lookups in DM-38498; they don't do anything yet.  Adding them early allows us to implement them on existing PipelineTasks in advance.

## Checklist

- [x] ran Jenkins
- [ ] added a release note for user-visible changes to `doc/changes` (will defer this to DM-38498, where these changes will start to do something)

### DIFF
--- a/python/lsst/pipe/base/connections.py
+++ b/python/lsst/pipe/base/connections.py
@@ -919,6 +919,67 @@ class PipelineTaskConnections(metaclass=PipelineTaskConnectionsMetaclass):
                 )
         return {}, {}
 
+    def getSpatialBoundsConnections(self) -> Iterable[str]:
+        """Return the names of regular input and output connections whose data
+        IDs should be used to compute the spatial bounds of this task's quanta.
+
+        The spatial bound for a quantum is defined as the union of the regions
+        of all data IDs of all connections returned here, along with the region
+        of the quantum data ID (if the task has spatial dimensions).
+
+        Returns
+        -------
+        connection_names : `collections.abc.Iterable` [ `str` ]
+            Names of collections with spatial dimensions.  These are the
+            task-internal connection names, not butler dataset type names.
+
+        Notes
+        -----
+        The spatial bound is used to search for prerequisite inputs that have
+        skypix dimensions. The default implementation returns an empty
+        iterable, which is usually sufficient for tasks with spatial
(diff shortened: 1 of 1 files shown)


python/lsst/pipe/base/connections.py
```diff
@@ -919,6 +919,56 @@ def adjustQuantum(self, inputs, outputs, label, data_id):
...
+        quanta.
+
+        The temporal bound for a quantum is defined as the union of the
+        timespans of all data IDs of all connections returned here, along with
+        the timespan of the quantum data ID (if the task has temporal
+        dimensions).
+
+        Notes
+        -----
+        The temproal bound is used to search for prerequisite inputs that are
```
erykoff, line 958: Typo in temporal here.

python/lsst/pipe/base/connections.py
```diff
@@ -919,6 +919,56 @@ def adjustQuantum(self, inputs, outputs, label, data_id):
...
+        calibration datasets. The default implementation returns an empty
+        iterable, which is usually sufficient for tasks with temporal
+        dimensions, but if a task's inputs or outputs are associated with
+        timespans that extend beyond the quantum data ID's timespan, this
+        method may need to be overridden to expand the set of prerequisite
+        inputs found.
+
+        Tasks that do not have temporal dimensions that have calibration
+        prerequisite inputs should always override this method, as the default
+        is that quanta are not temporally bounded.
```
erykoff, line 968: I find the phrase "the default is that quanta are not temporally bounded" to be confusing.  I'm not quite sure what this means?

erykoff (approved):

python/lsst/pipe/base/connections.py
```diff
@@ -919,6 +919,56 @@ def adjustQuantum(self, inputs, outputs, label, data_id):
                 )
         return {}, {}
 
+    def getSpatialBoundsConnections(self) -> Iterable[str]:
+        """Return the names of regular input and output connections whose data
```
erykoff, line 923: I just realized here and below `name` is (unfortunately) overloaded, in that we have the name of the class variable (which I think is intended) and also the `name` property of the connection.  I'm not sure how we distinguish this in our documentation now?

PR #367, opened 2023-08-14
DM-40367: Add connection name to deprecation warning

timj:
This provides some context. Also remove the find_outside_stacklevel because there is nothing in the stack trace that is relevant to the deprecation warning.

Depends on lsst/utils#168

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/pipe/base/connections.py
+++ b/python/lsst/pipe/base/connections.py
@@ -369,10 +369,11 @@ class PipelineTaskConnectionsMetaclass(type):
         instance._allConnections.clear()
         instance._allConnections.update(updated_all_connections)
 
-        for obj in instance._allConnections.values():
+        for connection_name, obj in instance._allConnections.items():
             if obj.deprecated is not None:
                 warnings.warn(
-                    f"{obj.name} (from {obj._deprecation_context}): {obj.deprecated}",
+                    f"Connection {connection_name} with datasetType {obj.name} "
+                    f"(from {obj._deprecation_context}): {obj.deprecated}",
                     FutureWarning,
                     stacklevel=1,  # Report from this location.
                 )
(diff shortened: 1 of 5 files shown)


python/lsst/pipe/base/connections.py
```diff
@@ -372,7 +372,9 @@ def __call__(cls, *, config: PipelineTaskConfig | None = None) -> PipelineTaskCo
         for obj in instance._allConnections.values():
             if obj.deprecated is not None:
                 warnings.warn(
-                    obj.deprecated, FutureWarning, stacklevel=find_outside_stacklevel("lsst.pipe.base")
+                    f"{obj.name} (from {obj._deprecation_context}): {obj.deprecated}",
```
TallJimbo, line 375: Instead of `obj.name` here, I think we want to iterate over `_allConnections.items()` and use the key from that mapping.  Currently this is the dataset type name, which is not unique, and the mapping key is the task-internal connection name, which is.

timj: Maybe if we had that info I wouldn't need to find the filename and line number where it was originally defined because the content would be unique enough that it could be found from searching the source.

natelust: I think the combination of the two would be good in addition to the work you did here for maximum info. something like "connection {dict key} with dataset type {obj.name} is deprecated....."

natelust (approved):

python/lsst/pipe/base/connections.py
timj: That was easy. Now it looks like:

```
FutureWarning: Connection finalizedPsfApCorrCatalog with datasetType finalVisitSummary (from /Users/timj/work/lsstsw/stack/lsst-scipipe-7.0.0/Darwin/pipe_tasks/g85f8d04a89+f020c27613/python/lsst/pipe/tasks/makeWarp.py:121): Deprecated in favor of 'visitSummary'.  Will be removed after v26.
```

natelust (approved):

PR #368, opened 2023-08-16
DM-40381: Improve mock support and fix bugs in init-input/output handling in QBB

TallJimbo:
Depends on lsst/daf_butler#879

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/pipe/base/graphBuilder.py
+++ b/python/lsst/pipe/base/graphBuilder.py
@@ -1609,7 +1609,8 @@ class _PipelineScaffolding:
         datastore_records: Mapping[str, DatastoreRecordData] | None = None
         if datastore is not None:
             datastore_records = datastore.export_records(
-                itertools.chain(
+                ref.makeCompositeRef() if ref.isComponent() else ref
+                for ref in itertools.chain(
                     self.inputs.iter_resolved_refs(),
                     self.initInputs.iter_resolved_refs(),
                     self.prerequisites.iter_resolved_refs(),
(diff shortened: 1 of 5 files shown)


python/lsst/pipe/base/graphBuilder.py
```diff
@@ -1553,6 +1547,31 @@ def resolveDatasetRefs(
...
+                # use them as inputs, or not at all.
+                for dataset_type in task.initOutputs:
+                    init_datasets = self.initIntermediates.pop(dataset_type, None)
+                    if init_datasets is not None:
+                        self.initInputs[dataset_type] = init_datasets
+                    self.initOutputs.pop(dataset_type, None)
+                # Removing the initInputs of this task from the scaffolding
+                # data structures is trickier, because the same initInput may
+                # be used by multiple tasks. This will be better addressed on
+                # DM-38498; for now we just leave them and let the QG have some
```
timj, line 1572: Maybe try to make a line match `# TODO:.*DM-38498` to follow standard form?

timj (commented):
The barren wasteland that is test coverage on this PR really makes me think that at some point we need to add coverage recording for middleware to ci_middleware so we can get some idea of what our coverage really is.

TallJimbo:
> The barren wasteland that is test coverage on this PR really makes me think that at some point we need to add coverage recording for middleware to ci_middleware so we can get some idea of what our coverage really is.

ci_middleware is already recording all of the coverage it's providing when it's run, but it's a pain to use, because you have to run it locally (and be careful about cleaning up coverage outputs from previous runs) and then do `coverage combine` and `coverage report` (or set up editor integration) yourself.  Since it doesn't run in GitHub Actions, I think we're stuck with Jenkins as our only option for coverage reporting that in a friendlier, more automated way.

PR #388, opened 2023-11-29
DM-41853: Handle MetaClass behavioral changes

natelust:
Some meta programming logic used in DatasetQueryConstraints was affected by a change in behavior in newer versions of python. This commit introduces a workaround, as this code may be refactored in the medium to long term.

## Checklist

- [ ] ran Jenkins

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.11"]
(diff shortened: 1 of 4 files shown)


timj (approved):
I can confirm that this works on python 3.12 and it's kind of strange to have a metaclass needing a metaclass...

natelust:
If you care on the why, its basically because this works kind of like a specialized version of Enum that allows state, so instances of variants are also considered instances of the enum. This supports both match operations, and also python type checking. I think there are different ways to do this now, but I shot for the minimal change.

PR #389, opened 2023-11-30
Dm-41203: Add pipeline subsetting control option.

natelust:
Add an option to pipeline subsetting which controls how named subsets are handled when they contain a task label which is not contained in the new subsetted pipeline.

## Checklist

- [ ] ran Jenkins


tests/test_pipelineIR.py
```diff
@@ -529,6 +531,40 @@ def testReadNamedSubsets(self):
         with self.assertRaises(ValueError):
             PipelineIR.from_string(pipeline_str)
 
+    def testSubsetttingPipeline(self):
```
leeskelvin, line 534: `ttt` typo

python/lsst/pipe/base/pipeline.py
```diff
@@ -584,6 +597,57 @@ def findSubsetsWithLabel(self, label: str) -> set[str]:
...
+    def addLabeledSubset(self, label: str, description: str, taskLabels: set[str]) -> None:
+        """Add a new labeled subset to the `Pipeline`.
+
+        Parameters
+        ----------
+        label : `str`
+            The label to assign to the subset.
+        description : `str`
+            A description of what the subset is for.
+        taskLabels : `set` of `str`
```
leeskelvin, line 619: Can/should this be:
```
`set` [`str`]
```
?

python/lsst/pipe/base/pipelineIR.py
```diff
@@ -52,6 +52,18 @@
 from lsst.utils.introspection import find_outside_stacklevel
 
 
+class PipelineSubsetCtrl(enum.Enum):
```
leeskelvin, line 55: It would be nice to add a class docstring here, to override the default Enum docstring that gets inherited instead.

python/lsst/pipe/base/pipeline.py
```diff
@@ -330,7 +332,11 @@ def from_uri(cls, uri: ResourcePathExpression) -> Pipeline:
...
 
-    def subsetFromLabels(self, labelSpecifier: LabelSpecifier) -> Pipeline:
+    def subsetFromLabels(
+        self,
+        labelSpecifier: LabelSpecifier,
+        subsetCtrl: pipelineIR.PipelineSubsetCtrl = PipelineSubsetCtrl.DROP,
+    ) -> Pipeline:
         """Subset a pipeline to contain only labels specified in labelSpecifier
 
         Parameters
```
leeskelvin, line 342: You need to add the `subsetCtrl` parameter to the docstring here too.

python/lsst/pipe/base/pipeline.py
```diff
@@ -342,6 +348,13 @@ def subsetFromLabels(self, labelSpecifier: LabelSpecifier) -> Pipeline:
         -------
         pipeline : `Pipeline`
             A new pipeline object that is a subset of the old pipeline
+        subsetCtrl : `PipelineSubsetCtrl`
```
leeskelvin, line 351: Should this be here? I think it needs to move up.

python/lsst/pipe/base/pipelineIR.py
```diff
@@ -651,6 +670,14 @@ def process_args(argument: str | dict) -> dict:
...
                 if "instrument" in argument and argument["instrument"] == "None":
                     argument["instrument"] = None
+                if "labeledSubsetModifyMode" in argument:
+                    match argument["labeledSubsetModifyMode"]:
+                        case "DROP":
+                            argument["labeledSubsetModifyMode"] = PipelineSubsetCtrl.DROP
+                        case "EDIT":
+                            argument["labeledSubsetModifyMode"] = PipelineSubsetCtrl.EDIT
+                        case unknown:
+                            raise ValueError(f"{unknown} is not a valid mode for labeledSubsetModifyMode")
```
leeskelvin, line 680: If you're in a unit-testing kind of mood, you could also add one to cover this raise? I won't object if you don't want to however.

PR #392, opened 2023-12-05
DM-41650 Add step definitions to pipelines

natelust:
## Checklist

- [ ] ran Jenkins


doc/changes/DM-41650.feature.md
```diff
@@ -0,0 +1 @@
+Added a section to pipelines which allows the explicit declaration of which susbsets correspond to steps and what the dimensions that steps are expected to operate over.
```
TallJimbo, line 1:
```suggestion
Added a section to pipelines which allows the explicit declaration of which susbsets correspond to steps and the dimensions the step's quanta can be sharded with.
```

doc/lsst.pipe.base/creating-a-pipeline.rst
```diff
@@ -341,6 +341,36 @@ Once a ``subset`` is created the label associated with it can be used in
...
+
+-----
+Steps
+-----
+Subsets are designed to be an encapsulation of a collection of tasks that are
+useful to be run together. These runs could be for reasons as small as a
+producing quick set of QA debugging plots or as large as divisions of the
+complete pipeline for survey level production.
+
+Because divisions of the pipeline have a special place in survey production
```
TallJimbo, line 354:
```suggestion
Because divisions of the pipeline have a special place in survey production,
```

doc/lsst.pipe.base/creating-a-pipeline.rst
```diff
@@ -341,6 +341,36 @@ Once a ``subset`` is created the label associated with it can be used in
...
+pipelines have a special place where they are highlighted, the ``steps``. key.
```
TallJimbo, line 355:
```suggestion
pipelines have a special place where they are highlighted, the ``steps`` key.
```

doc/lsst.pipe.base/creating-a-pipeline.rst
```diff
@@ -341,6 +341,36 @@ Once a ``subset`` is created the label associated with it can be used in
...
+This key is a list of the labels for all subsets that are considered steps in
+end to end processing. Alongside each label, a step must be declared with the
+set of dimensions the step is expected to run over. An example of what the step
+syntax looks like can be seen in the example below. These bits of information
+allow campaign management / batch production software better reason about how
+to handle the processing workflow found withing a pipeline.
```
TallJimbo, line 361: Please reformat the new text to have newlines after sentences and only after sentences.

python/lsst/pipe/base/pipelineIR.py
```diff
@@ -531,6 +538,24 @@ def __eq__(self, other: object) -> bool:
...
+    label : `str`
+        The label associated with this step
+    sharding_dimensions : `list` of `str`
+    """
+
+    label: str
+    sharding_dimensions: list[str]
+
+    def __hash__(self) -> int:
+        return hash((self.label, tuple(self.sharding_dimensions)))
```
TallJimbo, line 556: This seems dangerous; this is still a mutable object.  Is it actually needed?

python/lsst/pipe/base/pipelineIR.py
```diff
@@ -531,6 +538,24 @@ def __eq__(self, other: object) -> bool:
         )
 
 
+@dataclass(eq=True)
```
TallJimbo, line 541: Defining equality on a raw list of dimensions is not good; it makes `[instrument, visit]` different from `[visit]`, and even worse, different from `[visit, instrument]`.  The comparison at least needs to be on sets, but I'd avoid that, too if you can, even if that means deferring some tests until `PipelineGraph` can use a `DimensionUniverse` to do it.

python/lsst/pipe/base/pipelineIR.py
```diff
@@ -637,6 +668,28 @@ def _read_labeled_subsets(self, loaded_yaml: dict[str, Any]) -> None:
...
+    def _read_step_declaration(self, loaded_yaml: dict[str, Any]) -> None:
+        """Process the steps portion of the loaded yaml document
+
+        Steps are subsets that are declared to be normal parts of the overall
+        processing of the pipeline. Not all subsets need to be a step, as they
+        can exist for certain targeted processing, such as debugging.
+
+        Parameters
+        ----------
+        loaded_yaml: `MutableMapping`
```
TallJimbo, line 680:
```suggestion
        loaded_yaml: `dict`
```

If the type annotation says `dict`, use that here.  If you want `MutableMapping` in both places, you have to say `~collections.abc.MutableMapping` in the docs for it to link properly.

python/lsst/pipe/base/pipelineIR.py
```diff
@@ -637,6 +668,28 @@ def _read_labeled_subsets(self, loaded_yaml: dict[str, Any]) -> None:
...
+            A dictionary which matches the structure that would be produced
+            by a yaml reader which parses a pipeline definition document
+        """
+        loaded_steps = loaded_yaml.pop("steps", [])
+        self.steps: set[StepIR] = set()
```
TallJimbo, line 685: I think you can avoid all of the problems in `StepIR.__eq__` and `__hash__` by making this a `dict[str, StepIR]` or even a `dict[str, list[str]]` (and dropping `StepIR`).  It makes the IR slightly more different from the YAML form, but that seems like the lesser evil.

In fact, weren't we planning to make this a `list` in the YAML in order to validate the order, and then we decided not to validate the order because it was incompatible with importing steps?  So maybe this should be a `dict[str, list[str]]` in that YAML, too, after all?  That actually does preserve order anyway, which is nice for round-tripping even if we don't validate order.

python/lsst/pipe/base/pipelineIR.py
```diff
@@ -723,6 +786,7 @@ def merge_pipelines(self, pipelines: Iterable[PipelineIR]) -> None:
         accumulate_tasks: dict[str, TaskIR] = {}
         accumulate_labeled_subsets: dict[str, LabeledSubset] = {}
         accumulated_parameters = ParametersIR({})
+        accumulated_steps: set[StepIR] = set()
```
TallJimbo, line 789: This makes more sense as a `dict`, too.

TallJimbo (commented):
I think `StepIR.__hash__` and `StepIR.__eq__` as defined so far are going to be problematic, and I think it'll all go away if you hold them via `dict` instead of `set` (at least in Python, maybe in YAML).

doc/lsst.pipe.base/creating-a-pipeline.rst
```diff
@@ -341,6 +341,33 @@ Once a ``subset`` is created the label associated with it can be used in
...
+Subsets are designed to be an encapsulation of a collection of tasks that are
+useful to be run together. These runs could be for reasons as small as a
+producing quick set of QA debugging plots or as large as divisions of the
+complete pipeline for survey level production.
+
+Because divisions of the pipeline have a special place in survey production, pipelines have a special place where they are highlighted, the ``steps`` key.
+This key is a list of the labels for all subsets that are considered steps in end to end processing.
+Alongside each label, a step must be declared with the set of dimensions the step is expected to run over.
+An example of what the step syntax looks like can be seen in the example below.
+These bits of information allow campaign management / batch production software better reason about how to handle the processing workflow found withing a pipeline.
```
TallJimbo, line 358:
```suggestion
These bits of information allow campaign management / batch production software to better reason about how to handle the processing workflow found within a pipeline.
```

TallJimbo (approved):

PR #394, opened 2023-12-06
DM-41606: Option to output pipetask report info to command line

eigerx:
Changes in `pipe_base`: Add human-readable option to report summary dictionaries

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/pipe/base/execution_reports.py
+++ b/python/lsst/pipe/base/execution_reports.py
@@ -197,7 +197,9 @@ class TaskExecutionReport:
             else:
                 dataset_type_report.n_produced += 1
 
-    def to_summary_dict(self, butler: Butler, do_store_logs: bool = True) -> dict[str, Any]:
+    def to_summary_dict(
+        self, butler: Butler, do_store_logs: bool = True, human_readable: bool = False
+    ) -> dict[str, Any]:
         """Summarize the results of the TaskExecutionReport in a dictionary.
 
         Parameters
@@ -206,6 +208,9 @@ class TaskExecutionReport:
             The Butler used for this report.
         do_store_logs : `bool`
             Store the logs in the summary dictionary.
+        human_readable : `bool`
+            Store more human-readable information to be printed out to the
+            command-line.
 
         Returns
         -------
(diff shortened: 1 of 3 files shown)


python/lsst/pipe/base/execution_reports.py
```diff
@@ -218,11 +223,17 @@ def to_summary_dict(self, butler: Butler, do_store_logs: bool = True) -> dict[st
             - n_quanta_blocked: The number of quanta which failed due to
               upstream failures.
             - n_succeded: The number of quanta which succeeded.
+            And possibly, if human-readable is passed:
```
timj, line 226: I think the doc error is from this line. You might need to add a blank line above and below.

python/lsst/pipe/base/execution_reports.py
```diff
@@ -284,17 +309,26 @@ def to_summary_dict(self, butler: Butler, do_store_logs: bool = True) -> dict[st
...
+            command-line.
 
         Returns
         -------
         summary_dict : `dict`
             A dictionary containing a summary of a `TaskExecutionReport` for
             each task in the quantum graph.
         """
-        return {
-            task: report.to_summary_dict(butler, do_store_logs=do_store_logs)
```
kherner, line 295: Why not just change this line to be
`task: report.to_summary_dict(butler, do_store_logs=do_store_logs, human_readable=human_readable)` 

instead of adding the separate if/else statement?

python/lsst/pipe/base/execution_reports.py
```diff
@@ -235,7 +248,17 @@ def to_summary_dict(self, butler: Butler, do_store_logs: bool = True) -> dict[st
                     quantum_info["error"] = [
                         record.message for record in log if record.levelno >= logging.ERROR
                     ]
-            failed_quanta[str(node_id)] = quantum_info
+            if human_readable:
+                failed_quanta["data_id"] = data_ids
```
kherner, line 252: This ought to be easy enough to fix; you could just repeat whatever test calls this function and explicitly set human_readable to True in one case and not in the other.

kherner (approved):

python/lsst/pipe/base/execution_reports.py
eigerx: Thanks! I was thinking too hard about that one.

python/lsst/pipe/base/execution_reports.py
eigerx: Let's see if that works! I repeated the same tests (which should work, because the only difference between these reports is in the `failed_quanta` section which we don't introspect into much). I'll merge if my attempt at following all your suggestions pass Jenkins.

PR #396, opened 2023-12-21
DM-42302: Drop support for Pydantic v1.

TallJimbo:
Requires lsst/daf_butler#929

## Checklist

- [x] ran Jenkins

### DIFF
--- a/.github/workflows/build_docs.yaml
+++ b/.github/workflows/build_docs.yaml
@@ -31,7 +31,7 @@ jobs:
           pip install wheel
 
       - name: Install documenteer
-        run: pip install 'documenteer[pipelines]>=0.8'
+        run: pip install 'documenteer[pipelines]==0.8.2'
 
       - name: Install dependencies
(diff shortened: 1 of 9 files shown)


TallJimbo:
I have no idea why the doc build is trying to build docs for Pydantic itself (that seems to be what it's failing on), or why it's a problem here but not in daf_butler.

TallJimbo:
I've added some trivial forwarders to public classes that inherit from `pydantic.BaseModel`, to just override the docstrings so they aren't inherited.  It's not at all pretty, but it gets us through without having to weaken our guards against problems in our own docs.

It's hard for me to believe that `automodapi` is really designed to assume that it should parse the docstrings from any third-party classes that inherited from in its targeted package as rST, with no way to turn it off, but I can find no evidence that anyone else has run into this before.

timj (approved):
The sphinx fixes are pretty bad but seem pragmatic for now.

PR #398, opened 2024-01-08
DM-42301: Make calibration-lookup logic less restrictive.

TallJimbo:
Logic in the prerequisite finders that was expecting a regular calibration lookup was not correctly handling cases where the task did not have temporal dimensions even though the dataset was a calibration; it was raising exceptions even though falling through to findDataset or queryDatasets calls would have worked just fine.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/doc/changes/DM-42301.bugfix.md
+++ b/doc/changes/DM-42301.bugfix.md
@@ -1,0 +1,1 @@
+Fix a QG generation bug involving unusual combinations of dimensions and calibration datasets.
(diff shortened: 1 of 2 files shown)


timj (approved):
Looks okay. Do we know that ci_middleware exercises this code?

TallJimbo:
`ci_middleware` exercises most of the logic branches in `PrerequisiteFinders`, but not all of them, and I know for a fact that it doesn't exercise the bug that spawned this ticket, though `ci_cpp` might, but I just can't prioritize adding CPP-like data and tasks to `ci_middleware` right now.

Some of the other branches cannot actually be exercised since we only have one concrete `QuantumGraphBuilder` implementation class and that does optimized versions of some of those branches before they ever have a chance to fire.  I figure currently-unreachable code that's probably correct is better than raising or asserting, but :man_shrugging:?

PR #402, opened 2024-02-01
DM-42636: Use class methods for Instrument constructors

timj:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/pipe/base/_instrument.py
+++ b/python/lsst/pipe/base/_instrument.py
@@ -34,12 +34,13 @@ import datetime
 import os.path
 from abc import ABCMeta, abstractmethod
 from collections.abc import Sequence
-from typing import TYPE_CHECKING, Any, cast, final
+from typing import TYPE_CHECKING, Any, Self, cast, final
 
 from lsst.daf.butler import DataCoordinate, DataId, DimensionPacker, DimensionRecord, Formatter
 from lsst.daf.butler.registry import DataIdError
 from lsst.pex.config import Config, RegistryField
 from lsst.utils import doImportType
+from lsst.utils.introspection import get_full_type_name
 
 from ._observation_dimension_packer import observation_packer_registry
 
@@ -136,8 +137,8 @@ class Instrument(metaclass=ABCMeta):
         """
         raise NotImplementedError()
 
-    @staticmethod
-    def fromName(name: str, registry: Registry, collection_prefix: str | None = None) -> Instrument:
+    @classmethod
+    def fromName(cls, name: str, registry: Registry, collection_prefix: str | None = None) -> Self:
         """Given an instrument name and a butler registry, retrieve a
         corresponding instantiated instrument object.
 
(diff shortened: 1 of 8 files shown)


python/lsst/pipe/base/_instrument.py
```diff
@@ -80,6 +81,10 @@ class Instrument(metaclass=ABCMeta):
     of the dataset type name, a tuple of dimension names, and the storage class
     name. If `None` the ingest system will use its default definition."""
 
+    day_obs_offset: int = 0
+    """Offset (in seconds) to apply when converting a ``day_obs`` value
+    to a timespan (`int`)."""
```
TallJimbo, line 86: This might be better as an abstract method or property so third-party `obs_` packages don't miss the need to set it and fill their databases with bad `day_obs` timespans before they find out.

TallJimbo (approved):
Changes look good, but I wonder if they'd be better in the `obs_base` `Instrument` subclass, to make `pipe_base` assume our dimension universe a bit less.

python/lsst/pipe/base/_instrument.py
```diff
@@ -735,3 +744,27 @@ def _make_default_dimension_packer(
             IDs into integers.
         """
         return config_dict.apply_with(default, data_id, is_exposure=is_exposure)
+
+    @classmethod
+    def group_name_to_group_id(cls, group_name: str) -> int:
```
timj, line 749: @TallJimbo you think this should move to `lsst.obs.base.Instrument` rather than being in the base class?

python/lsst/pipe/base/_instrument.py
```diff
@@ -75,6 +77,13 @@ class Instrument(metaclass=ABCMeta):
     """Instrument specific name to use when locating a policy or configuration
     file in the file system."""
 
+    translatorClass: MetadataTranslator | None = None
```
timj, line 80: @gpdf are SPHEREx using MetadataTranslator? In the review we were wondering if this should be moved to `obs_base` instead since it might be assuming too much about how people generate exposure records.

gpdf: Not at present.  What is the "this" that you want to move?

python/lsst/pipe/base/_instrument.py
timj: One downside of moving this to obs_base is that mypy is clever enough to notice that `Instrument.fromName` is not returning something that has that method. I've changed the return types to `Self` on `Instrument` for the constructors but that's not strictly true in that those methods are not going to magically return the obs_base variant if you call `lsst.obs.base.Instrument.fromName("name.that.inherits.from.pipe.base.Instrument")`.

timj: I've modified the pipe_base variant to use class methods and then check that the returned instrument instance is a subclass of the class you used to call `fromName` et al. That will at least let things crash early if people are using obs_base define visits without using an obs_base Instrument.

python/lsst/pipe/base/_instrument.py
timj: I have moved all the changes to lsst/obs_base#477

PR #403, opened 2024-02-20
DM-42928: Add methods to get and set nested dictionaries from TaskMetadata

TallJimbo:
The methods here are trivial, but the PropertyList implementation is not, and it's the consistency between all three types (TaskMetadata, PropertyList, and PropertySet) that's really the value added here.

See also lsst/daf_base#82

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/pipe/base/_task_metadata.py
+++ b/python/lsst/pipe/base/_task_metadata.py
@@ -69,7 +69,7 @@ class SetDictMetadata(Protocol):
     dictionary into these objects that avoids their historical idiosyncrasies.
 
     The form in which these entries appear in the object's native keys and
-    values is implementation-defined.  *Empty nested dictionaries *may* be
+    values is implementation-defined.  *Empty nested dictionaries may be
     dropped, and if the top-level dictionary is empty this method may do
     nothing.*
     """
--- a/python/lsst/pipe/base/_task_metadata.py
+++ b/python/lsst/pipe/base/_task_metadata.py
@@ -541,7 +541,8 @@ class TaskMetadata(BaseModel):
         Parameters
         ----------
         key : `str`
-            String key associated with the mapping.
+            String key associated with the mapping.  May not have a ``.``
+            character.
 
         Returns
         -------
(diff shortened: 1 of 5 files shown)


python/lsst/pipe/base/_task_metadata.py
```diff
@@ -56,6 +60,41 @@ def _isListLike(v: Any) -> bool:
...
+class SetDictMetadata(Protocol):
+    """Protocol for objects that can be assigned a possibly-nested `dict` of
+    primitives.
+
+    This protocol is satisfied by `TaskMetadata`, `lsst.daf.base.PropertySet`,
+    and `lsst.daf.base.PropertyList`, providing a consistent way to insert a
+    dictionary into these objects that avoids their historical idiosyncrasies.
+
+    The form in which these entries appear in the object's native keys and
+    values is implementation-defined.  *Empty nested dictionaries *may* be
```
timj, line 72: This nesting of markup doesn't work in sphinx.

python/lsst/pipe/base/_task_metadata.py
```diff
@@ -56,6 +60,41 @@ def _isListLike(v: Any) -> bool:
     return isinstance(v, Sequence) and not isinstance(v, str)
 
 
+class SetDictMetadata(Protocol):
```
timj, line 63: Could there be a single `DictMetadata` protocol that has both in?

python/lsst/pipe/base/_task_metadata.py
```diff
@@ -477,6 +516,46 @@ def __delitem__(self, key: str) -> None:
...
+        Parameters
+        ----------
+        key : `str`
+            String key associated with the mapping.
+        value : `~collections.abc.Mapping`
+            Possibly-nested mapping, with `str` keys and values that are `int`,
+            `float`, `str`, `bool`, or another `dict` with the same key and
+            value types.
+        """
+        self[key] = TaskMetadata.from_dict(value)
```
timj, line 557:
```python
self[key] = value
```
should work fine. The `TaskMetadata.from_dict` will be called later on.

python/lsst/pipe/base/_task_metadata.py
```diff
@@ -477,6 +516,46 @@ def __delitem__(self, key: str) -> None:
...
+        ----------
+        key : `str`
+            String key associated with the mapping.
+
+        Returns
+        -------
+        value : `~collections.abc.Mapping`
+            Possibly-nested mapping, with `str` keys and values that are `int`,
+            `float`, `str`, `bool`, or another `dict` with the same key and
+            value types.  Will be empty if ``key`` does not exist.
```
timj, line 535: Maybe mention this is not a view and so changes to the returned dict won't affect the original?

timj (approved):
This looks okay to me.

PR #405, opened 2024-03-15
DM-43060: Write CachingLimitedButler

mfisherlevine:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/pipe/base/caching_limited_butler.py
+++ b/python/lsst/pipe/base/caching_limited_butler.py
@@ -1,0 +1,90 @@
+# This file is part of pipe_base.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (http://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This software is dual licensed under the GNU General Public License and also
+# under a 3-clause BSD license. Recipients may choose which of these licenses
+# to use; please see the files gpl-3.0.txt and/or bsd_license.txt,
+# respectively.  If you choose the GPL option then the following text applies
+# (but note that there is still no warranty even if you opt for BSD instead):
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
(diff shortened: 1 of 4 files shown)


TallJimbo (approved):
Given that we pair-coded this, it shouldn't be a surprise that I'm happy with it.

The lack of test coverage is a problem.  I don't think it would be hard to add some here - especially if you add some debug logging as we've discussed to fire when the cache is used; you can then trap those warnings in the tests.  But I also know that it might be a big ask for you to do it _right now_ (as it would be for me to do them).  If you aren't up for adding some tests, could you at least create a ticket for that?

I would also appreciate a ticket for handling `getDeferred` vs. caching; I think that'd be best handled by refactoring `DeferredDatasetHandle` to have a reference back to its butler, so your override to `get` can be used, but I don't think your use case needs it done.

mfisherlevine:
Of course, and thank you!

[Ticket for tests](https://rubinobs.atlassian.net/browse/DM-43353)
[Ticket for `getDeferred`](https://rubinobs.atlassian.net/browse/DM-43354)

mfisherlevine:
I just added some debug logs to show the caching, ahead of the testing.

mfisherlevine:
I just added some debug logs to show the caching, ahead of the testing.

timj:
It looks like the caching butler is not even imported anywhere. Can we have a short test that imports it at least?

python/lsst/pipe/base/caching_limited_butler.py
```diff
@@ -0,0 +1,194 @@
...
+        obj = self._wrapped.get(ref, parameters=parameters, storageClass=storageClass)
+        if ref.datasetType.name in self._cache_on_get and not parameters:
+            # and not parameters is to make sure we don't cache sub-images etc
+            self._cache[ref.datasetType.name] = (
+                ref.id,
+                InMemoryDatasetHandle(
+                    obj,
+                    storageClass=storageClass,
+                    dataId=ref.dataId,
+                    copy=ref.datasetType.name not in self._no_copy_on_cache,
```
andy-slac, line 121: I understand that `copy` flag only matters on `handle.get`? So if you save this handle and return `obj`, the task could modify `obj` which will also modify contents of the `handle`? Probably need to copy `obj` before storing it in handle (unless `copy` is false)?

mfisherlevine:
I have pushed an _extremely_ minimal test that the class instantiates here, and linked the tests ticket in a `TODO` there. Let me know if that doesn't look OK.

python/lsst/pipe/base/caching_limited_butler.py
mfisherlevine: I'm actually not quite sure. @TallJimbo could you comment, do you think?

mfisherlevine: We resolved this OOB. Changes pushed and Jenkins running.

PR #406, opened 2024-03-21
DM-43225: Save stack versions in QuantumGraph.

MichelleGower:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/pipe/base/graph/graph.py
+++ b/python/lsst/pipe/base/graph/graph.py
@@ -53,6 +53,7 @@ from lsst.daf.butler import (
 from lsst.daf.butler.persistence_context import PersistenceContextVars
 from lsst.resources import ResourcePath, ResourcePathExpression
 from lsst.utils.introspection import get_full_type_name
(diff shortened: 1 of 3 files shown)


python/lsst/pipe/base/graph/graph.py
```diff
@@ -175,6 +176,14 @@ def _buildGraphs(
         and the graph that holds the relations between quanta
         """
         self._metadata = metadata
+
+        # Save packages to metadata
+        metadata2 = {}
+        if self._metadata is not None:
+            metadata2 = dict(self._metadata)
+        metadata2["packages"] = Packages.fromSystem()
+        self._metadata = metadata2
```
natelust, line 185: I think what you wrote is fine, but if you are now going to always be adding to the metadata now may I suggest:
```
self._metadata = dict(metadata) if metadata is not None else {}
Self._metadata[‘packages’] = Packages.fromSystem()
```
Fewer round trips and copies being made.

natelust (approved):

PR #407, opened 2024-03-25
Add Input connection with deferred binding

tgoldina:
These are minimal changes to make an input connection with deferred binding, which would allow us to have an input connection with the same dataset type as an output. (The input would have a different dataset reference than the output.)

This is the first step in creating a `Quantum Graph`, which would process images with persistence. In this use case, the images must be processed in order, each `QuantumNode` takes prior persistence state and produces its own.

## Checklist

- [ ] ran Jenkins

### DIFF
--- a/python/lsst/pipe/base/connectionTypes.py
+++ b/python/lsst/pipe/base/connectionTypes.py
@@ -298,6 +298,13 @@ class Input(BaseInput):
         spatial overlaps.  This option has no effect when the connection is not
         an overall input of the pipeline (or subset thereof) for which a graph
         is being created, and it never affects the ordering of quanta.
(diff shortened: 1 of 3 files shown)


TallJimbo (approved):
Changes all look good, thanks!  I don't recall whether you can just merge this directly or whether we should make a Jira ticket for it (maybe @timj does).

timj:
Jira please. Tatiana should have access.

timj:
Also can we have a news fragment please?

timj:
I've created [DM-43572](https://rubinobs.atlassian.net/browse/DM-43572) so this PR should be closed and reopened with the new ticket branch.

PR #410, opened 2024-03-28
DM-43572: Input connection with deferred binding

tgoldina:
These are minimal changes to make an input connection with deferred binding, which would allow us to have an input connection with the same dataset type as an output. (The input would have a different dataset reference than the output.)

This is the first step in creating a Quantum Graph, which would process images with persistence. In this use case, the images must be processed in order, each QuantumNode takes prior persistence state and produces its own.

This replaces https://github.com/lsst/pipe_base/pull/407

## Checklist

- [x] ran Jenkins

### DIFF
--- a/doc/changes/DM-43572.feature.rst
+++ b/doc/changes/DM-43572.feature.rst
@@ -1,0 +1,2 @@
+Added ``deferBinding`` attribute to ``Input`` connection, which allows us
+to have an input connection with the same dataset type as an output.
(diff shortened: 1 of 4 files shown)


TallJimbo (approved):

timj:
The mypy problem is caused by a change in daf_butler. I'm fixing it in lsst/daf_butler#986 and the mypy check can be rerun once that merges.

PR #413, opened 2024-04-12
DM-43869: Explicitly allow pydantic to use NaN and Inf

timj:
## Checklist

- [x] ran Jenkins


ktlim (approved):
Seems a bit magical, but I guess that's how Pydantic works.

PR #415, opened 2024-04-22
DM-43020: Implement region and time extraction for preload

kfindeisen:
This PR adds a `RegionTimeInfo` class for bundling regions and timespans into one Butler dataset. It also fixes some pre-existing warnings.

## Checklist

- [x] ran Jenkins


TallJimbo (approved):
Thanks for all the extra tidies, especially the comment about the stuff in dataset constraints, which was quite surprising to me, too.

PR #421, opened 2024-06-04
DM-44648: Allow conflicting module paths to be resolved late

natelust:
When instantiating a pipeline that contains imports, allow the system to resolve conflicting module paths of tasks late. This allows the same task imported from a different path to still resolve as the same.

## Checklist

- [ ] ran Jenkins


TallJimbo (approved):
Code looks fine, but codecov has spotted some logic branches the new tests didn't cover, and it doesn't look at first glance like it'd be hard to include them.

PR #430, opened 2024-07-29
u/parejkoj/pipeline: improve docs on config overrides

parejkoj:


doc/lsst.pipe.base/creating-a-pipeline.rst
```diff
@@ -451,15 +451,57 @@ class name of the python camera object.  For instance, for an ``obs_subaru``
 
   instrument: lsst.obs.subaru.HyperSuprimeCam
```
kfindeisen, line 452: I don't see any documentation of what order config settings are applied in (field defaults, `setDefaults`, general obs overrides, instrument obs overrides, top-level pipelines, included pipelines, and maybe other stuff I'm forgetting). It would be hard for the reader to make informed decisions without that kind of big picture.

kfindeisen (changes requested):
I don't think this documentation is either accurate or accessible to new users. Maybe it would be better to create a Jira issue and take the time to do it properly?

timj:
@parejkoj are you going to be able to get back to this PR or should we close it?

parejkoj:
Given the pushback in the review, you can just close it. If someone wants to revive it and get this convention into the docs, they've got some text here to crib from.

PR #432, opened 2024-08-01
DM-45536: address edge cases in task mocking system

TallJimbo:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/pipe/base/connections.py
+++ b/python/lsst/pipe/base/connections.py
@@ -493,15 +493,8 @@ class DeferredDatasetRef:
 
     datasetRef: DatasetRef
 
-    @property
-    def datasetType(self) -> DatasetType:
-        """The dataset type for this dataset."""
-        return self.datasetRef.datasetType
-
-    @property
-    def dataId(self) -> DataCoordinate:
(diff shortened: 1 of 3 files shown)


python/lsst/pipe/base/connections.py
```diff
@@ -493,15 +493,8 @@ class DeferredDatasetRef:
...
-    def datasetType(self) -> DatasetType:
-        """The dataset type for this dataset."""
-        return self.datasetRef.datasetType
-
-    @property
-    def dataId(self) -> DataCoordinate:
-        """The data ID for this dataset."""
-        return self.datasetRef.dataId
+    def __getattr__(self, name: str) -> Any:
+        return getattr(self.datasetRef, name)
```
timj, line 497: This change is because you needed some other method on datasetRef? Or because you wanted to simplify the code a bit?

timj (approved):

python/lsst/pipe/base/connections.py
TallJimbo: Yes, there was some previously-unused code in the mocking stuff that needed `id` that was uncovered by the other change.

PR #433, opened 2024-08-02
Mention the ability to disable output connections

gpdf:
## Checklist

- [ ] ran Jenkins

Reviewer: please advise on whether "necessary" or "recommended" is more correct in:

"It is now no longer ( necessary / recommended ) to delegate to `super` in either case,"

There was clearly a word missing in any event.

### DIFF
--- a/doc/lsst.pipe.base/creating-a-pipelinetask.rst
+++ b/doc/lsst.pipe.base/creating-a-pipelinetask.rst
@@ -415,7 +415,8 @@ The dimensions of the task itself can also be modified in ``__init__``, where ``
                 self.inputs.remove("background")
 
     This still works, but it is more verbose and less intuitive than the now-recommended attribute-manipulation approach.
-    It is now no longer to delegate to `super` in either case, since the base class `super` does nothing (the first steps of initialization are instead handled by a metaclass).
+    It is now no longer recommended to delegate to `super` in either case,
+    since the base class `super` does nothing (the first steps of initialization are instead handled by a metaclass).
 
 
 The last step in modifying this task will be to update the ``run`` method to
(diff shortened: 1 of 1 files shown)


TallJimbo:
Delegation to `super` is now guaranteed to be a no-op, so "recommended" is probably better.

doc/lsst.pipe.base/creating-a-pipelinetask.rst
```diff
@@ -472,7 +473,12 @@ passed on to the run method. The body of the run method checks if the background
 has been passed, and if so adds it back in and does a local background
 subtraction.
 
-To bring this all together, see :ref:`pipeline-appendix-c`
+To bring this all together, see :ref:`pipeline-appendix-c`.
+
+Note that the same mechanism works for output connections:
+an output from a `PipelineTask` can be made optional (under config control)
+or a config switch can even be used to choose between different ways of
+providing output.
```
TallJimbo, line 481: Might be worth clarifying that this is different from making an input optional at runtime by setting `minimum=0` in the `Input` connection declaration.  In the configuration case, the connection is treated as if it never existed by QG generation and execution code, while `minimum=0` allows a quantum to be generated and run even no dataset for that input are found.  (There is no `minimum` for `Output` connections because they are _always_ considered runtime-optional.)

TallJimbo (approved):

doc/lsst.pipe.base/creating-a-pipelinetask.rst
gpdf: Let me know if you think I've captured it now.

gpdf:
Squashed commits in preparation for merge

PR #440, opened 2024-08-27
DM-45722: CRITICAL logs on empty quantum graph

kfindeisen:
This PR demotes the explanation of empty quantum graph queries from critical to error (to avoid confusion with truly fatal errors that force termination), and also merges the explanation into a single multiline log (for compatibility with JSON/Loki output).

## Checklist

- [x] ran Jenkins

### DIFF
--- a/doc/changes/DM-45722.bugfix.md
+++ b/doc/changes/DM-45722.bugfix.md
@@ -1,0 +1,1 @@
+Explanatory logs for "initial data ID query returned no rows" now appear as a single log message instead of one entry per line. This improves display in log aggregators, but there is no change to console behavior.
--- a/python/lsst/pipe/base/all_dimensions_quantum_graph_builder.py
+++ b/python/lsst/pipe/base/all_dimensions_quantum_graph_builder.py
@@ -507,7 +507,7 @@ class _AllDimensionsQuery:
                 yield result
 
     def log_failure(self, log: LsstLogAdapter) -> None:
-        """Emit a series of CRITICAL-level log message that attempts to explain
+        """Emit a series of ERROR-level log message that attempts to explain
         why the initial data ID query returned no rows.
 
         Parameters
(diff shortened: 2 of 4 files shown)


doc/changes/DM-45722.bugfix.md
```diff
@@ -0,0 +1 @@
+Explanatory logs for "initial data ID query returned no rows" now appear as a single entry in Loki/Json instead of being split on line breaks. There is no change to console behavior.
```
timj, line 1: I don't think Loki needs to be mentioned explicitly. I think you are saying "now appears as a single log message rather than multiple log entries".

doc/changes/DM-45722.other.md
```diff
@@ -0,0 +1 @@
+Explanatory logs for "initial data ID query returned no rows" are now reported at ERROR, not CRITICAL, level.
```
timj, line 1: This should be "misc". We decided to merge misc and other but I see we forgot to edit the README about that. Would you mind deleting the "other" entry from `doc/changes/README.rst` please?

python/lsst/pipe/base/all_dimensions_quantum_graph_builder.py
```diff
@@ -507,31 +508,39 @@ def from_builder(
...
+            buffer.write(
+                "To reproduce this query for debugging purposes, run "
+                "Registry.queryDataIds with these arguments:\n"
+            )
+            # We could just repr() the queryArgs dict to get something
+            # the user could make sense of, but it's friendlier to
+            # put these args in an easier-to-reconstruct equivalent form
+            # so they can read it more easily and copy and paste into
+            # a Python terminal.
+            buffer.write("  dimensions=%s," % list(self.query_args["dimensions"].names))
```
timj, line 535: We generally prefer `.format()` or f-strings in middleware. `ruff check --select UP031 .` triggers for this line. I haven't enabled UP031 in pipe_base yet because I didn't need to (this reminded me I planned to do that). I can fix this later when I turn on the check.

timj (approved):
Thanks for the cleanup.

python/lsst/pipe/base/all_dimensions_quantum_graph_builder.py
kfindeisen: I thought it would be better to keep the arguments analogous to what the direct logging calls would have been, but ok.

doc/changes/DM-45722.bugfix.md
kfindeisen: On second thought, I think I need at least an indirect reference to Loki, because the behavior there is what makes this a bugfix and not a coding style change. Is the new wording adequate?
> Explanatory logs for "initial data ID query returned no rows" now appear as a single log message instead of one entry per line. This improves display in log aggregators, but there is no change to console behavior.

timj: Works for me.

PR #442, opened 2024-09-04
DM-46064: permit storage class conversions of components in PipelineGraph

TallJimbo:
This can leave the component DatasetType/Ref seen by the component-reading task with an unexpected parentStorageClass, but we hope the task won't care.

## Checklist

- [x] ran Jenkins


timj (approved):
Thanks for doing this. Hopefully it will all work out. I can't think of an easy way for anything to know what the relevant parent should be without adding that as an optional parameter to the connection object.

PR #447, opened 2024-09-16
DM-46351: Add debug log in _pipeline_graph to ID task label

leeskelvin:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/pipe/base/pipeline_graph/_pipeline_graph.py
+++ b/python/lsst/pipe/base/pipeline_graph/_pipeline_graph.py
@@ -712,6 +712,7 @@ class PipelineGraph:
             label = task_class._DefaultName
         if config is None:
             config = task_class.ConfigClass()
+        _LOG.debug("Adding task %s %s to the pipeline graph", label, task_class)
         task_node = TaskNode._from_imported_data(
             key=NodeKey(NodeType.TASK, label),
             init_key=NodeKey(NodeType.TASK_INIT, label),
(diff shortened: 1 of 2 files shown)


python/lsst/pipe/base/pipeline_graph/_pipeline_graph.py
```diff
@@ -712,6 +712,7 @@ def add_task(
             label = task_class._DefaultName
         if config is None:
             config = task_class.ConfigClass()
+        _LOG.debug("Task: %s %s", label, task_class)
```
natelust, line 715: I think something more verbose would be more helpful when disambiguating debug logs. Maybe something like `Adding task with label: %s and class: %s to the pipeline graph`

natelust (approved):

python/lsst/pipe/base/pipeline_graph/_pipeline_graph.py
leeskelvin: The current ticket produces output that looks like this:
```
lsst.pipe.base.pipeline_graph DEBUG: Task: photometricMatchVisit <class 'lsst.analysis.tools.tasks.photometricCatalogMatch.PhotometricCatalogMatchVisitTask'>
```

Your suggestion would produce output that looks like this:
```
lsst.pipe.base.pipeline_graph DEBUG: Adding task with label: photometricMatchVisit and class: <class 'lsst.analysis.tools.tasks.photometricCatalogMatch.PhotometricCatalogMatchVisitTask'> to the pipeline graph
```

Is the double "class" what you had in mind?

leeskelvin: Having discussed in-person, modifying to:
```
Adding task: %s %s to the pipeline graph
```

PR #449, opened 2024-10-12
DM-46599: Stop using deprecated property butler.collections

dhirving:
This has been superseded by 'butler.collections.defaults'.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/pipe/base/tests/simpleQGraph.py
+++ b/python/lsst/pipe/base/tests/simpleQGraph.py
@@ -500,7 +500,7 @@ def makeSimpleQGraph(
         pipeline_graph,
         butler,
         skip_existing_in=skipExistingIn if skipExistingIn is not None else [],
-        input_collections=butler.collections.defaults if butler.collections.defaults is not None else [run],
+        input_collections=butler.collections.defaults,
         output_run=run,
         where=userQuery,
         bind=bind,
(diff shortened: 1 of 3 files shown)


dhirving:
That mypy breakage is totally unrelated to my change and just popped up because I had to re-push after a rebase; I'm juggling 6 different repos and will deal with it later.

python/lsst/pipe/base/tests/simpleQGraph.py
```diff
@@ -500,7 +500,7 @@ def makeSimpleQGraph(
         pipeline_graph,
         butler,
         skip_existing_in=skipExistingIn if skipExistingIn is not None else [],
-        input_collections=butler.collections if butler.collections is not None else [run],
+        input_collections=butler.collections.defaults if butler.collections.defaults is not None else [run],
```
timj, line 503: I am wondering why we had the `is not None` there. Wouldn't `if butler.collections.defaults else` work to cover None and empty list? Does the new way of doing it ever return `None`?

timj (approved):

python/lsst/pipe/base/tests/simpleQGraph.py
dhirving: Must be deep historical reasons -- even prior to your replacement of `butler.collections` I don't think this could have been `None`.  Both versions just look at `RegistryDefaults.collections` which is an empty tuple if there are no collections.

timj: The test code presumably never ended up with `[run]` then.

PR #453, opened 2024-10-28
DM-47143: Enable entry points for comand-line plugin discovery

timj:
Depends on lsst/daf_butler#1107

## Checklist

- [x] ran Jenkins

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -63,10 +63,9 @@ jobs:
 
       - name: Run tests
         shell: bash -l {0}
-        env:
-          DAF_BUTLER_PLUGINS: ${{ github.workspace }}/python/lsst/pipe/base/cli/resources.yaml
         run: |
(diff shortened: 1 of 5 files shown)


.github/workflows/build.yaml
```diff
@@ -63,10 +63,9 @@ jobs:
 
       - name: Run tests
         shell: bash -l {0}
-        env:
-          DAF_BUTLER_PLUGINS: ${{ github.workspace }}/python/lsst/pipe/base/cli/resources.yaml
         run: |
           pytest -r a -v -n 3 --cov=lsst.pipe.base --cov=tests --cov-report=xml --cov-report=term --cov-branch
+          butler -h
```
andy-slac, line 68: Should this be `butler register-instrument -h`?

andy-slac (approved):

.github/workflows/build.yaml
timj: Ok. I was mostly using this to look at the output to make sure it is working but using one of the expected plugin names is better.

PR #455, opened 2024-11-11
DM-47505: Don't raise when --dataset-query-constraint args are irrelevant.

TallJimbo:
They might be relevant for a different subgraph, and this option needs to be usable on multi-subgraph pipelines where the subgraphs don't have overall inputs in common.

## Checklist

- [x] ran Jenkins


hlin730 (approved):
Verified with these changes that LSSTComCam DRP step3b indeed generates quantum graph rapidly when --dataset-query-constraints for all 3 step3b sub-graphs are specified together, without raising an error as it did before.

PR #460, opened 2024-12-18
DM-48206: Graph-based visualization of pipeline processing status (text representation)

enourbakhsh:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/pipe/base/pipeline_graph/visualization/_status_annotator.py
+++ b/python/lsst/pipe/base/pipeline_graph/visualization/_status_annotator.py
@@ -83,6 +83,14 @@ class StatusColors:
     or statuses.
     """
 
+    # Base task status colors.
+    expected: str = WHITE
+    succeeded: str = GREEN
+    failed: str = RED
+
+    # Base dataset type status colors.
+    produced: str = GREEN
+
     # Reset to default color.
     reset: str = RESET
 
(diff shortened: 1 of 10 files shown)


python/lsst/pipe/base/pipeline_graph/visualization/_status_annotator.py
```diff
@@ -0,0 +1,248 @@
...
+    @overload
+    def __call__(self, xgraph: networkx.DiGraph, dataset_types: Literal[False]) -> None: ...
+
+    @overload
+    def __call__(self, xgraph: networkx.MultiDiGraph, dataset_types: Literal[True]) -> None: ...
+
+    def __call__(self, xgraph: networkx.DiGraph | networkx.MultiDiGraph, dataset_types: bool) -> None:
+        for task_label, task_summary in self.qpg_summary.tasks.items():
+            fields = {
+                name.replace("n_", "").replace("successful", "succeeded"): getattr(task_summary, name)
```
eigerx, line 230: Oh no, successful and succeeded are back to get us again! Is this something that would be easier if we renamed it in the QPG itself?

eigerx (approved):
Really nice work. I was going to finish off this review by digging up a recent/interesting pipetask report and visualizing it myself, but I see this work is already going on with @erinleighh in #ops-cm-team ! In that case I'll approve pending a visualization of a recent graph based on the current weekly that looks like the fantastic example on the Jira ticket.

python/lsst/pipe/base/pipeline_graph/visualization/_status_annotator.py
enourbakhsh: It's probably a good idea!

enourbakhsh: Not totally sure what side effects that renaming might have in other parts of the codebase - it might be a more involved process. For now, I think sticking with `replace("successful", "succeeded")` is fine.

PR #463, opened 2025-01-15
DM-47730: add "success caveats" system to report on NoWorkFound and similar cases

TallJimbo:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/pipe/base/_status.py
+++ b/python/lsst/pipe/base/_status.py
@@ -28,14 +28,16 @@
 from __future__ import annotations
 
 import abc
+import enum
 import logging
-from typing import Protocol
+from typing import ClassVar, Protocol
 
 from lsst.utils import introspection
 
 from ._task_metadata import GetSetDictMetadata, NestedMetadataDict
 
 __all__ = (
+    "QuantumSuccessCaveats",
     "UnprocessableDataError",
     "AnnotatedPartialOutputsError",
     "NoWorkFound",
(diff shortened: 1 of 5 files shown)


tests/test_quantum_success_caveats.py
```diff
@@ -0,0 +1,63 @@
+# This file is part of ctrl_mpexec.
```
MichelleGower, line 1: Copy/Paste error.  Should state part of pipe_base.

MichelleGower (approved):
Looks good.  1 copy/paste mistake.  Merge approved.

python/lsst/pipe/base/_status.py
```diff
@@ -46,6 +48,130 @@
...
+    NO_CAVEATS = 0
+    """All outputs were produced and no exceptions were raised."""
+
+    ANY_OUTPUTS_MISSING = enum.auto()
+    """At least one predicted output was not produced."""
+
+    ALL_OUTPUTS_MISSING = enum.auto()
+    """No predicted outputs (except logs and metadata) were produced.
+
+    `ANY_OUTPUTS_MISSING` is also set whenever this flag it set.
```
eigerx, line 69: May want to change docstring to ``ANY_OUTPUTS_MISSING` is also set whenever this flag is set.` 

I think the use of "any" here is okay, but might suggest something in `ctrl_mpexec.cli.script.report.print_summary` to make these a little easier to parse. (Feel free to show an example if it is already straightforward and I'm missing something).

eigerx: Whoops! If this happens again I'll send a quick DM. I was just reading through this to understand the other two PR's and wanted to comment since I was here. It seems like the most important stuff made it in.

PR #464, opened 2025-01-24
DM-35396: Record provenance in QuantumContext

timj:
Requires lsst/daf_butler#1147

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/pipe/base/_quantumContext.py
+++ b/python/lsst/pipe/base/_quantumContext.py
@@ -242,7 +242,7 @@ class QuantumContext:
     def _put(self, value: Any, ref: DatasetRef) -> None:
         """Store data in butler."""
         self._checkMembership(ref, self.allOutputs)
-        self.__butler.put(value, ref)
+        self.__butler.put(value, ref, provenance=self.dataset_provenance)
         self.outputsPut.add((ref.datasetType, ref.dataId))
 
     def get(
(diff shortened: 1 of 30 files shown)


TallJimbo (approved):
There's a `CachingLimitedButler` in this package that needs to get the new `put` signature.

python/lsst/pipe/base/_quantumContext.py
```diff
@@ -214,25 +230,30 @@ def __init__(
             for ref in refs:
                 self.allOutputs.add((ref.datasetType, ref.dataId))
         self.outputsPut: set[tuple[DatasetType, DataCoordinate]] = set()
+        self.outputRefsPut: list[DatasetRef] = []
```
timj, line 233: @TallJimbo you added `ouputsPut` here as a tuple of DatasetType and DataCoordinate. I need the ref ID. I wasn't sure why you were storing a tuple for comparison rather than the ref itself so I added another item (which I should probably have as a set now I think about it but I wanted to retain order). Should I add the ref.id to your tuple? Or keep my own list?

TallJimbo: I needed to be able to do a set-difference with `allOutputs` (in `ctrl.mpexec.SingleQuantumExecutor`), so I just used the same type as that.  If you can keep that set-difference (or equivalent) working, feel free to change it however you'd like.

andy-slac (approved):
Looks great.

PR #471, opened 2025-02-20
DM-49045: run/runQuantum documentation updates from RFC-1052

parejkoj:
## Checklist

- [ ] ran Jenkins

### DIFF
--- a/doc/lsst.pipe.base/creating-a-pipelinetask.rst
+++ b/doc/lsst.pipe.base/creating-a-pipelinetask.rst
@@ -47,8 +47,6 @@ Next, create a |Task| class that performs the measurements:
                 "apFlux", type=np.float64, doc="Ap flux measured"
             )
 
-            self.outputCatalog = afwTable.SourceCatalog(self.outputSchema)
-
         def run(
             self, exposure: afwImage.Exposure, inputCatalog: afwTable.SourceCatalog
         ) -> pipeBase.Struct:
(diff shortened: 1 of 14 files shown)


doc/lsst.pipe.base/creating-a-pipelinetask.rst
```diff
@@ -1056,6 +1041,8 @@ method to have a convenient interface for user interaction within a notebook
 or shell, but still match the types of input `PipelineTask`\ s will get when
 run by the middleware system.
 
+TODO: the below explicitly violates the "stateless" convention for Tasks that we've tried to maintain since forever. Might be some work to redo it completely; is it worth keeping?
+
```
TallJimbo, line 1045: How about we turn this TODO into a `.. note:` that basically says, "this is for educational purposes only; don't really do this."  The last sentence in the paragraph below basically says that but could say it a long stronger.

If you want to cook up an even better example of overriding `runQuantum` (maybe passing down a data ID value?) that'd be even better, but I'm trying not to suggest too many scope increases in a review.

TallJimbo (approved):

doc/lsst.pipe.base/creating-a-pipelinetask.rst
parejkoj: I'll file a ticket about revising this further, but I'd rather not try to do that here.

PR #473, opened 2025-03-02
DM-48482: Build image rendering around `mermaid-py` for Mermaid diagrams

enourbakhsh:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/pipe/base/_datasetQueryConstraints.py
+++ b/python/lsst/pipe/base/_datasetQueryConstraints.py
@@ -117,7 +117,7 @@ if sys.version_info.minor < 12:
 else:
 
     class MetaMeta(type(Protocol)):
-        def __init__(cls, *args, **kwargs):
+        def __init__(cls, *args, **kwargs):  # noqa: N805
             # Note: this prepends an *extra* cls to type's argument list
             return super().__init__(cls, *args, **kwargs)
 
--- a/python/lsst/pipe/base/pipeline_graph/visualization/_mermaid.py
+++ b/python/lsst/pipe/base/pipeline_graph/visualization/_mermaid.py
@@ -29,9 +29,11 @@ from __future__ import annotations
 __all__ = ("show_mermaid",)
 
 import html
+import importlib.util
 import os
 import sys
 from collections.abc import Mapping
+from io import BufferedIOBase, BytesIO, StringIO, TextIOBase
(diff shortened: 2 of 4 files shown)


jrmullaney (approved):
Look fine - I've just got one question about whether to throw an error with the import.

python/lsst/pipe/base/pipeline_graph/visualization/_mermaid.py
```diff
@@ -40,6 +42,12 @@
 from ._options import NodeAttributeOptions
 from ._show import parse_display_args
 
+MERMAID_AVAILABLE = importlib.util.find_spec("mermaid") is not None
+
+if MERMAID_AVAILABLE:
+    from mermaid import Mermaid  # type: ignore
+    from mermaid.graph import Graph  # type: ignore
```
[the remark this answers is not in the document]
timj: We can't raise an exception on import if the file is being imported with all visualization imports unless we make mermaid a mandatory requirement.

python/lsst/pipe/base/_datasetQueryConstraints.py
```diff
@@ -117,7 +117,7 @@ def fromExpression(cls, expression: str) -> DatasetQueryConstraintVariant:
 else:
 
     class MetaMeta(type(Protocol)):
-        def __init__(cls, *args, **kwargs):
+        def __init__(cls, *args, **kwargs):  # noqa: N805
```
timj, line 120: Interesting that `ruff check` doesn't think this is an error. We don't use flake8 for this package so not sure how you found this but I confirm flake8 is upset.

enourbakhsh: I may have run `flake8` on the code locally out of habit.

timj:
@enourbakhsh what happened to the tests I asked you to write?

timj:
I just saw the comment on Jira. I thought I had responded to that but it was a draft message. I was hoping you could mock out the mermaid-py part and have it return the completed thing so that the code here could be tested rather than us having no idea at all whether it worked or not.

enourbakhsh:
> @enourbakhsh what happened to the tests I asked you to write?

Working on it on [DM-49310](https://rubinobs.atlassian.net/browse/DM-49310).

PR #474, opened 2025-03-03
DM-49266: Fix out-of-order logic in QG builder.

TallJimbo:
When trying to skip a quantum because it had previously succeeded, but some predicted outputs had not been produced, were were deleting the node in the 'skeleton' QG before trying to access some of its attributes.

## Checklist

- [ ] ran Jenkins


andy-slac (approved):
Looks good.

PR #477, opened 2025-03-11
DM-49300: Add hook for adjusting all of a task's quanta.

TallJimbo:
This is much more powerful than adjusting a single quantum, because it allows inputs to be moved from one quantum to another.  It's also very much locked down to the minimum we need for the motivating use case, because I'm worried about it being so powerful that it becomes hard to support in the future if we change the QG generation algorithm.  We can open it up further as needed.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/pipe/base/connections.py
+++ b/python/lsst/pipe/base/connections.py
@@ -35,6 +35,7 @@ __all__ = [
     "InputQuantizedConnection",
     "OutputQuantizedConnection",
     "PipelineTaskConnections",
+    "QuantaAdjuster",
     "QuantizedConnection",
     "ScalarError",
     "ScalarError",
@@ -45,8 +46,8 @@ import dataclasses
 import itertools
 import string
 import warnings
-from collections import UserDict
-from collections.abc import Collection, Generator, Iterable, Mapping, Sequence, Set
+from collections import UserDict, defaultdict
+from collections.abc import Collection, Generator, Iterable, Iterator, Mapping, Sequence, Set
 from dataclasses import dataclass
 from types import MappingProxyType, SimpleNamespace
 from typing import TYPE_CHECKING, Any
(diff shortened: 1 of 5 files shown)


tests/test_adjust_all_quanta.py
```diff
@@ -0,0 +1,142 @@
...
+class GroupTestConfig(PipelineTaskConfig, pipelineConnections=GroupTestConnections):
+    pass
+
+
+class GroupTestTask(PipelineTask):
+    ConfigClass = GroupTestConfig
+
+
+class AdjustAllQuantaTestCase(unittest.TestCase):
+    """Tests for the `PipelineTaskConnection.adjust_all_quanta` hook in
```
andy-slac, line 100:
```suggestion
    """Tests for the `PipelineTaskConnections.adjust_all_quanta` hook in
```

python/lsst/pipe/base/connections.py
```diff
@@ -1130,3 +1152,159 @@ def adjust_in_place(
...
+        If two connections have the same dataset type, the current
+        implementation assumes the set of datasets is the same for the two
+        connections.  This limitation may be removed in the future.
+        """
+        from .quantum_graph_skeleton import DatasetKey, QuantumKey
+
+        quantum_key = QuantumKey(self._task_node.label, quantum_data_id.required_values)
+        if (read_edge := self._task_node.inputs.get(connection_name)) is None:
+            read_edge = self._task_node.prerequisite_inputs[connection_name]
+        dataset_key = DatasetKey(read_edge.parent_dataset_type_name, dataset_data_id.required_values)
```
andy-slac, line 1281: Should this be `PrerequisiteDatasetKey` in case it is in `prerequisite_inputs`?

andy-slac (approved):
Looks great!

python/lsst/pipe/base/connections.py
TallJimbo: Actually, I need to make it an error for this to be be prerequisite input, because those can't be identified wholly by their data IDs (they might be calibrations with validity ranges).  I'd dropped support elsewhere (and even updated the docs) when I realized that halfway through, but I apparently missed this bit.

PR #480, opened 2025-03-17
DM-49440: Optimizations for QuanutmProvenanceGraph

TallJimbo:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/pipe/base/quantum_provenance_graph.py
+++ b/python/lsst/pipe/base/quantum_provenance_graph.py
@@ -1308,6 +1308,7 @@ class QuantumProvenanceGraph:
             as well as diagnostic information and error messages for failed
             quanta and strange edge cases, and a list of recovered quanta.
         """
+        status_log = PeriodicLogger(_LOG)
         if not self._finalized:
             raise RuntimeError(
                 """resolve_duplicates must be called to finalize the
@@ -1315,12 +1316,12 @@ class QuantumProvenanceGraph:
             )
         result = Summary()
         futures: list[concurrent.futures.Future[None]] = []
+        _LOG.verbose("Summarizing %s tasks.", len(self._quanta.keys()))
         with concurrent.futures.ThreadPoolExecutor(n_cores, initializer=self._thread_init) as executor:
-            for task_label, quanta in self._quanta.items():
-                _LOG.verbose("Summarizing %s quanta for task %r.", len(quanta), task_label)
+            for m, (task_label, quanta) in enumerate(self._quanta.items()):
                 task_summary = TaskSummary()
                 task_summary.n_expected = len(quanta)
-                for quantum_key in quanta:
+                for n, quantum_key in enumerate(quanta):
                     quantum_info = self.get_quantum_info(quantum_key)
                     future = task_summary._add_quantum_info(
                         quantum_info,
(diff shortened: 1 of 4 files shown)


python/lsst/pipe/base/quantum_provenance_graph.py
```diff
@@ -1195,245 +1286,562 @@ def get_dataset_info(self, key: DatasetKey) -> DatasetInfo:
...
+                    status_log.log(
+                        "Summarized %s of %s quanta of task %s of %s.",
+                        n + 1,
+                        len(quanta),
+                        m + 1,
+                        len(self._quanta.keys()),
+                    )
+                result.tasks[task_label] = task_summary
+            for n, future in enumerate(concurrent.futures.as_completed(futures)):
+                if (err := future.exception()) is not None:
```
timj, line 1341: In theory you can collect all the exceptions from the futures and raise them in an `ExceptionGroup` when they've all completed.

python/lsst/pipe/base/quantum_provenance_graph.py
```diff
@@ -1195,245 +1286,562 @@ def get_dataset_info(self, key: DatasetKey) -> DatasetInfo:
...
+                    len(datasets),
+                    m + 1,
+                    len(self._datasets.keys()),
+                )
+            result.datasets[dataset_type_name] = dataset_type_summary
+        return result
+
+    def iter_outputs_of(self, quantum_key: QuantumKey) -> Iterator[DatasetKey]:
+        """Iterate through the outputs of a quantum, yielding all the
+        `DatasetKey`s produced by the quantum.
```
timj, line 1368: Are you sure that rst syntax works? I think you might need
```
`DatasetKey`\ s
```

python/lsst/pipe/base/quantum_provenance_graph.py
```diff
@@ -1195,245 +1286,562 @@ def get_dataset_info(self, key: DatasetKey) -> DatasetInfo:
...
+            return
+        quantum_info = self.get_quantum_info(quantum_key)
+        quantum_run = quantum_info["runs"][output_run]
+
+        def read_metadata() -> None:
+            md = self._butler_get(quantum_run.metadata_ref, storageClass="TaskMetadata")
+            try:
+                # Int conversion guards against spurious conversion to
+                # float that can apparently sometimes happen in
+                # TaskMetadata.
```
timj, line 1833: `TaskMetadata` is in some sense `dict[str, Any]` but I think the order of StrictInt vs StrictFloat matters.

https://github.com/lsst/pipe_base/blob/main/python/lsst/pipe/base/_task_metadata.py#L130

timj (approved):
This is a bit of a tour de force and I can't say I've deeply analyzed every line. It does look good.

Have you ever tried with a `-j` greater than 10 on non-posix? boto has a connection pool of 10 and for small files it will likely start issuing warnings for more than 10 threads. Similarly, if the log files are large boto will split them up into chunks so you might run into connection exhaustion sooner. For signed URLs an env var can control the number of connections in `resources`.

python/lsst/pipe/base/quantum_provenance_graph.py
TallJimbo: Thanks for the tip. In this case I don't think there are any error conditions we expect, and so it's better to fail fast because it means something is seriously wrong.

python/lsst/pipe/base/quantum_provenance_graph.py
TallJimbo: I was wondering how this could happen.   Should I go reorder those on this ticket, and see if that lets me remove this?  I'm a little worried that could break something else (including things that might not be tested).

TallJimbo (commented):
I have not tested this on anything other than POSIX at all, actually, though I think my advice to users would be, "don't use -j if you're using implicit threading already".

I'll run some tests on `embargo` before I merge to make sure things don't go horribly there.

python/lsst/pipe/base/quantum_provenance_graph.py
timj: It looks like the code has been `StrictFloat, StrictInt` from the beginning. Are you saying that `5` always comes back as `5.` because of that order?  I wonder if `5.` would be read as `5` if we flip the order. I think we used Strict versions because there were issues with numpy floats/ints in the serialization.

timj:
PS Thanks for adding periodic logger support with `--verbose`.

python/lsst/pipe/base/quantum_provenance_graph.py
TallJimbo: I think that's right; I was thinking that an exception would make the context manager do `shutdown(wait=True, cancel_futures=True)` , but it seems it does `shutdown(wait=True, cancel_futures=False)` regardless of whether the context is exited with an exception or not.  So the result is that it runs to completion but shows the first error message.

That's okay in this context, as I think it's most likely they'd all have the same error here and 100k copies of the same error message is not a great experience.  But it's something we should think about if we add our own tooling for this; a context manager that does cancel on exceptions certainly seems more useful to me.

python/lsst/pipe/base/quantum_provenance_graph.py
TallJimbo: I can't say for sure that it always happens, but I did observe it happen at least sometimes, and my only theory for what's going on would suggest that it always happens.

And I do think you're right about what would happen if we flipped it, which makes me think we should not flip it, since it's just trading one problem for another + potential disruption.

TallJimbo:
I had completely forgotten that I already did test this with S3, because some of my benchmarking was on OR5 runs.

PR #488, opened 2025-04-22
DM-50310: use daf_butler dimension record attacher instead of local one

TallJimbo:
## Checklist

- [x] ran Jenkins


timj (approved):
Looks like it has reduced a lot of code here so that is good.

PR #489, opened 2025-04-25
DM-50490: Track time spent in get or put in QuantumContext

timj:
Needs lsst/utils#212

## Checklist

- [ ] ran Jenkins


timj:
We have decided that it is simpler to put the metrics directly into daf_butler in lsst/daf_butler#1187

PR #492, opened 2025-04-29
DM-50539: minor QG builder fixes

TallJimbo:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/doc/changes/DM-50539.feature.md
+++ b/doc/changes/DM-50539.feature.md
@@ -1,0 +1,1 @@
+Added support for moving outputs from one quantum to another in the `adjust_all_quanta` hook, as well as access to a `Butler` instance with implementations of that hook.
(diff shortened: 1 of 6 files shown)


TallJimbo:
Test coverage is artificially bad because the main test is skipped when `daf_butler` is provided by piff, because that doesn't include some test data.  That test is run in the Jenkins CI.

timj:
Re: test coverage

```python
DATA_DIR = os.path.join(getPackageDir("daf_butler"), "tests", "data", "registry")
```
This test is a bit flaky in general. Jenkins is going to be okay because eups is installing the test data but the test data doesn't get installed by pip or docker containers so in those scenarios it would never work (and `getPackageDir` is already saying that EUPS is required).

The best fix would be to move the registry test data into the tests directory in the `python` tree and then include it in the pip install and use package data to find it. Something to ponder for the future.

doc/changes/DM-50538.feature.md
```diff
@@ -0,0 +1 @@
+Added a `butler` attribute and support for moving outputs from one quantum to another in the `adjust_all_quanta` hook.
```
andy-slac, line 1: Not clear whose attribute it is, maybe say `QuantaAdjuster.butler`?

andy-slac (approved):
Looks great.

PR #494, opened 2025-04-30
DM-50491: Add metrics support to CachingLimitedButler

timj:
## Checklist

- [ ] ran Jenkins


mfisherlevine (commented):
I don't know enough about the code to review it on the merits, but I can confirm that it fixes the breakage in RA 👍

PR #495, opened 2025-05-08
DM-50550: Add exception diagnostics table to quantum provenance

enourbakhsh:
## Checklist

- [ ] ran Jenkins


python/lsst/pipe/base/quantum_provenance_graph.py
```diff
@@ -1075,6 +1126,110 @@ def make_exception_table(self) -> astropy.table.Table:
...
+            messages.
+        """
+        add_exposure_records = True
+        needed_exposure_records = ["day_obs", "physical_filter", "exposure_time", "target_name"]
+
+        # Preload all exposure dimension records up front for faster O(1)
+        # lookup later. Querying per data ID in the loop is painfully slow.
+        if butler:
+            exposure_record_lookup = {
+                d.dataId["exposure"]: d for d in butler.query_dimension_records("exposure", explain=False)
```
enourbakhsh, line 1175: I wish I could limit `butler.query_dimension_records` to just the data IDs with exceptions, instead of pulling everything. But it only takes one dataId, and I don't think it's possible to use `where` to filter it.

timj: This works fine for me if you already know the exposure IDs:
```python
>>> butler.query_dimension_records("exposure", where="exposure in (:exp)", bind={"exp": (903334, 903336)}, instrument="HSC")
```

TallJimbo: Use
```
with butler.query() as query:
    query = query.join_data_coordinates(<list of exposure data IDs>)
    ...
```

TallJimbo: Actually, change `.required` to `.mapping` in this line:

https://github.com/lsst/pipe_base/blob/924f49135fb97c9293592500239795adae769306/python/lsst/pipe/base/quantum_provenance_graph.py#L658

and then the data IDs you're starting with should have all of those implied values.  They're already in the data IDs in the QG, but we were dropping them on the floor.

enourbakhsh: Will give it a try!

enourbakhsh: Done!

python/lsst/pipe/base/quantum_provenance_graph.py
```diff
@@ -1075,6 +1164,141 @@ def make_exception_table(self) -> astropy.table.Table:
...
+            messages.
+        """
+        if add_dimension_records and butler is None:
+            raise ValueError("Butler is required to fetch dimension records.")
+
+        # The additional columns for visit and exposure records to add to the
+        # output table, if available. Note that 'band', 'day_obs', and
+        # 'physical_filter' already exist in `exception.data_id` below.
+        needed_visit_records = ["exposure_time", "target_name"]
+        needed_exposure_records = ["exposure_time", "target_name"]
```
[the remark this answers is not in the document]
timj: We modified the code so that here it calls an entry point that returns a dict of dimensions with list values of the metadata fields to be included.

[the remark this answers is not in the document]
enourbakhsh: @timj by the time I had it polished, @cmsaunders was already in the middle of the first review round. I didn’t want to interfere, so I’m planning to push those changes in the second round.

PR #497, opened 2025-05-21
DM-50969: Fix QG logging to reflect what's actually going on.

TallJimbo:
There can be a substantial delay between submitting the query and starting to iterate over the rows, which was making it hard to tell when the query plan was so bad the DB wasn't even returning anything vs. a large result set being returned.

This also adds verbose-level periodic logging to monitor progress when processing the query result rows.

## Checklist

- [ ] ran Jenkins

### DIFF
--- a/python/lsst/pipe/base/all_dimensions_quantum_graph_builder.py
+++ b/python/lsst/pipe/base/all_dimensions_quantum_graph_builder.py
@@ -48,7 +48,7 @@ from lsst.daf.butler import (
     DimensionRecordSet,
     MissingDatasetTypeError,
 )
-from lsst.utils.logging import LsstLogAdapter
+from lsst.utils.logging import LsstLogAdapter, PeriodicLogger
 from lsst.utils.timer import timeMethod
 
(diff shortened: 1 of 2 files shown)


python/lsst/pipe/base/all_dimensions_quantum_graph_builder.py
```diff
@@ -215,17 +215,26 @@ def _query_for_data_ids(self, tree: _DimensionGroupTree) -> None:
...
+            progress_logger: PeriodicLogger | None = None
             for common_data_id in query.data_ids(tree.all_dimensions):
+                if progress_logger is None:
+                    # There can be a long wait between submitting the query and
+                    # returning the first row, so we want to make sure we log
+                    # when we get it; note that PeriodicLogger is not going to
+                    # do that for us, as it waits for its interval _after_ the
+                    # first log is seen.
+                    self.log.info("Iterating over data ID query results.")
+                    periodic_logger = PeriodicLogger(self.log)
```
erykoff, line 232: Should this be `progress_logger = `?

andy-slac (approved):
Looks good.

PR #499, opened 2025-06-09
DM-51273: Improve transfer_from_graph efficiency.

MichelleGower:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/pipe/base/script/transfer_from_graph.py
+++ b/python/lsst/pipe/base/script/transfer_from_graph.py
@@ -27,11 +27,17 @@
 
 __all__ = ["transfer_from_graph"]
 
+import math
+
 from lsst.daf.butler import Butler, CollectionType, QuantumBackedButler, Registry
 from lsst.daf.butler.registry import MissingCollectionError
 from lsst.pipe.base import QuantumGraph
+from lsst.utils.iteration import chunk_iterable
+from lsst.utils.logging import getLogger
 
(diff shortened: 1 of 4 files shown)


python/lsst/pipe/base/script/transfer_from_graph.py
```diff
@@ -92,18 +97,25 @@ def transfer_from_graph(
...
-        register_dataset_types=register_dataset_types,
-        transfer_dimensions=transfer_dimensions,
-        dry_run=dry_run,
-    )
-    count = len(transferred)
+    # For faster restarts, filter out those the destination already knows.
+    filtered_refs = filter_by_existence(dest_butler, filtered_refs)
+
+    # Transfer in chunks
+    count = 0
```
timj, line 108: Can you do something like:
```python
n_chunks = math.ceil(len(refs) / chunk_size)
chunk_num = 0
for chunk in ...:
    chunk_num += 1
    if n_chunks > 1:
        _LOG.verbose("Transferring %d datasets in chunk %d/%d", len(chunk), chunk_num, n_chunks)
    ...
```
so that when people are doing a lot of this we can see the progress? The alternative is to use the progress bar stuff but hardly anyone turns on the progress bar flag and that's not really suitable for batch anyhow.

timj (approved):
Thanks. Looks good. Minor comments on logging.

PR #500, opened 2025-06-12
DM-51363: update how test data from daf_butler is loaded

TallJimbo:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/tests/test_adjust_all_quanta.py
+++ b/tests/test_adjust_all_quanta.py
@@ -30,7 +30,6 @@ from __future__ import annotations
 __all__ = ()
 
 import operator
-import os
 import unittest
 from collections import defaultdict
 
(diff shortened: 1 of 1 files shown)


tests/test_adjust_all_quanta.py
```diff
@@ -111,13 +110,9 @@ class AdjustAllQuantaTestCase(unittest.TestCase):
 
     @staticmethod
     def make_butler() -> Butler:
-        try:
-            DATA_DIR = os.path.join(getPackageDir("daf_butler"), "tests", "data", "registry")
```
timj, line 115: I had no idea we were doing this. The test data is also not available in a pipelines docker container either (we strip the tests directories to make the container smaller)

timj (approved):

PR #503, opened 2025-06-25
DM-51509: Extend QuantaAdjuster to handle PrerequisiteInputs

enourbakhsh:
## Checklist

- [x] ran Jenkins


python/lsst/pipe/base/prerequisite_helpers.py
```diff
@@ -353,6 +353,7 @@ def find(
                         where=" AND ".join(where_terms),
                         bind=bind,
                         with_dimension_records=True,
+                        limit=None,
```
enourbakhsh, line 356: Should we also add `explain = False`?

TallJimbo (approved):

PR #506, opened 2025-07-16
DM-51821: Fix transfer-from-graph's update chain

MichelleGower:
Fix transfer-from-graph to update chain when asked and output run collection exists even if didn't transfer any datasets.

## Checklist

- [ ] ran Jenkins

### DIFF
--- a/python/lsst/pipe/base/script/transfer_from_graph.py
+++ b/python/lsst/pipe/base/script/transfer_from_graph.py
@@ -29,7 +29,7 @@ __all__ = ["transfer_from_graph"]
 
 import math
 
-from lsst.daf.butler import Butler, CollectionType, QuantumBackedButler, Registry
+from lsst.daf.butler import Butler, CollectionType, QuantumBackedButler
 from lsst.daf.butler.registry import MissingCollectionError
 from lsst.pipe.base import QuantumGraph
 from lsst.utils.iteration import chunk_iterable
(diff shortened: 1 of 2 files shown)


python/lsst/pipe/base/script/transfer_from_graph.py
```diff
@@ -124,23 +124,34 @@ def transfer_from_graph(
...
+    try:
+        _ = butler.collections.get_info(output_run)
+    except MissingCollectionError:
+        _LOG.verbose(
+            "output RUN collection (%s) does not exist.  Skipping updating the output chain.",
+            output_run,
+        )
+        return
+
+    registry = butler.registry
```
timj, line 154: I think the code below needs to be updated as well. It's out of date. This is the old way that has a race condition. There is now an actual API for it of `butler.collections.prepend_chain` that is an atomic operation.

python/lsst/pipe/base/script/transfer_from_graph.py
```diff
@@ -124,45 +123,53 @@ def transfer_from_graph(
...
+    Parameters
+    ----------
+    butler : `lsst.daf.butler.Butler`
+        Butler where to update the collection chain.
+    output_chain : `str`
+        Name of the output CHAINED collection.
+    output_run : `str`
+        Name of the output RUN collection.
+    inputs : `list` [`str`] | None
+
```
timj, line 152:
```suggestion
        All the input collections to be included in the chain if the chain is created.
```

python/lsst/pipe/base/script/transfer_from_graph.py
```diff
@@ -124,45 +123,53 @@ def transfer_from_graph(
...
+
     """
+    # Do not need to update chain if output_run does not already exists.
```
timj, line 155:
```suggestion
    # Do not need to update chain if output_run does not already exist.
```

python/lsst/pipe/base/script/transfer_from_graph.py
```diff
@@ -124,45 +123,53 @@ def transfer_from_graph(
...
-                return
-            else:
-                del chain_definition[index]
-        except ValueError:
-            pass
-
-        chain_definition.insert(0, output_run)
-        registry.setCollectionChain(output_chain, chain_definition)
+        _LOG.verbose(
+            "output RUN collection (%s) does not exist.  Skipping updating the output chain.",
```
timj, line 160:
```suggestion
            "Output RUN collection (%s) does not exist.  Skipping updating the output chain.",
```

python/lsst/pipe/base/script/transfer_from_graph.py
```diff
@@ -124,45 +123,53 @@ def transfer_from_graph(
...
+        )
+        return
+
+    # Make chain collection if doesn't exist before calling prepend_chain.
+    created_now = butler.collections.register(output_chain, CollectionType.CHAINED)
+    if created_now:
+        _LOG.verbose("Registered chain collection: %s", output_chain)
+        if inputs:
+            # Add input collections to chain collection just made.
+            butler.collections.prepend_chain(output_chain, inputs)
```
timj, line 171: I think if we are racing with another caller, we probably want to use extend_chain here so that the inputs end up on the end rather than risk them jumping in front of a RUN output collection.

timj (approved):
Looks great. Lots of tiny comments but I love how much code we deleted (and that code was unsafe as well)

PR #508, opened 2025-07-28
DM-50124: Add rename to dict to ImportIR

taranu:
## Checklist

- [ ] ran Jenkins


python/lsst/pipe/base/pipelineIR.py
```diff
@@ -461,6 +461,8 @@ class ImportIR:
     """list of tasks that should be excluded when inheriting this pipeline.
     Either the include or exclude attributes may be specified, but not both.
     """
+    rename: dict[str, str] | None = None
```
TallJimbo, line 464: I think you could clean up the logic later by defining this as
```
rename: dict[str, str] = dataclasses.field(default_factory=dict)
```
since in this case (unlike `include` and `exclude`) `None` behaves like an empty `dict`.

PR #561, opened 2026-05-12
DM-54879: Support reprocessing when upstream outputs are selectively retained

hsinfang:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/pipe/base/quantum_graph_builder.py
+++ b/python/lsst/pipe/base/quantum_graph_builder.py
@@ -39,6 +39,7 @@ __all__ = (
     "QuantumGraphBuilderError",
 )
 
+import collections
 import dataclasses
 import operator
 from abc import ABC, abstractmethod
@@ -61,6 +62,7 @@ from lsst.daf.butler import (
 from lsst.daf.butler._rubin import generate_uuidv7
 from lsst.daf.butler.datastore.record_data import DatastoreRecordData
 from lsst.daf.butler.registry import MissingCollectionError, MissingDatasetTypeError
+from lsst.daf.butler.utils import globToRegex
 from lsst.utils.logging import LsstLogAdapter, getLogger
 from lsst.utils.timer import timeMethod
 
@@ -128,6 +130,14 @@ class QuantumGraphBuilder(ABC):
     skip_existing_in : `~collections.abc.Sequence` [ `str` ], optional
         Collections to search for outputs that already exist for the purpose of
         skipping quanta that have already been run.
+    retained_dataset_types : `~collections.abc.Sequence` [ `str` ], optional
+        Dataset type names or glob-style wildcard patterns for dataset types
+        that should exist in ``skip_existing_in`` when the producing task ran
+        successfully.  When a quantum should run, the builder propagates the
+        must-run signal backward through non-retained input datasets, forcing
+        the upstream quanta that need to regenerate those intermediates to also
+        run.  Has no effect without ``skip_existing_in``.  ``["*"]`` means
+        retaining all datasets, equivalent to not providing this option.
     clobber : `bool`, optional
         Whether to raise if predicted outputs already exist in ``output_run``
         (not including those quanta that would be skipped because they've
@@ -171,6 +181,7 @@ class QuantumGraphBuilder(ABC):
         input_collections: Sequence[str] | None = None,
         output_run: str | None = None,
         skip_existing_in: Sequence[str] = (),
+        retained_dataset_types: Sequence[str] | None = None,
         clobber: bool = False,
     ):
         self.log = getLogger(__name__)
@@ -188,6 +199,11 @@ class QuantumGraphBuilder(ABC):
         self.butler = butler.clone(collections=input_collections)
         self.output_run = output_run
         self.skip_existing_in = skip_existing_in
+        self._retained_dataset_type_patterns: list[str] | None = (
+            list(retained_dataset_types) if retained_dataset_types is not None else None
+        )
+        if self._retained_dataset_type_patterns is not None and not skip_existing_in:
+            raise ValueError("retained_dataset_types has no effect without skip_existing_in.")
         self.empty_data_id = DataCoordinate.make_empty(butler.dimensions)
         self.clobber = clobber
         # See whether the output run already exists.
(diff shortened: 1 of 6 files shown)


python/lsst/pipe/base/quantum_graph/aggregator/_writer.py
```diff
@@ -163,7 +163,7 @@ def make_compression_dictionary(self) -> zstandard.ZstdCompressionDict:
             self.comms.log.info("Making compressor with no dictionary.")
             return zstandard.ZstdCompressionDict(b"")
         self.comms.log.info("Training compression dictionary.")
-        training_inputs: list[bytes] = []
+        training_inputs: list[bytes | bytearray | memoryview[int]] = []
```
TallJimbo: I'm curious where this is coming from; AFAIK we don't use `bytearray` or `memoryview[int]` for any of these.

tests/test_graphBuilder.py
```diff
@@ -228,6 +231,75 @@ def test_datastore_records(self):
...
 
 
+class SkipExistingInTestCase(unittest.TestCase):
+    """Tests for the skip_existing_in behavior of QuantumGraphBuilder."""
+
+    def setUp(self):
+        repodir = tempfile.TemporaryDirectory()
+        self.addCleanup(tempfile.TemporaryDirectory.cleanup, repodir)
+        pipeline = simpleQGraph.makeSimplePipeline(nQuanta=1)
+        butler, _ = simpleQGraph.makeSimpleQGraph(root=repodir.name, pipeline=pipeline, nQuanta=1)
```
TallJimbo: I don't think this is worth changing at this point, but for new tests I'd generally recommend `lsst.pipe.base.tests.mocks.InMemoryRepo` or `lsst.pipe.base.tests.mocks.DirectButlerRepo` (see `test_predicted_qg.py` for an example).  That also preps a butler repo with input datasets and helps you build a pipeline, and while it's just about as concise for simple cases, it's also extensible to more complex graphs.

python/lsst/pipe/base/quantum_graph_builder.py
```diff
@@ -128,6 +128,15 @@ class QuantumGraphBuilder(ABC):
     skip_existing_in : `~collections.abc.Sequence` [ `str` ], optional
         Collections to search for outputs that already exist for the purpose of
         skipping quanta that have already been run.
+    ignore_metadata_for : `~collections.abc.Iterable` [ `str` ], optional
+        Task labels for which the task metadata dataset is not used as the
+        completion signal when ``skip_existing_in`` is provided.  For these
+        tasks a quantum is skipped only when all of its non-metadata outputs
+        are present in ``skip_existing_in``; it is not skipped when any
+        non-metadata output is absent.  This is useful for pipelines
+        where some upstream tasks do not retain all of their outputs, so that
```
TallJimbo: Reword to "where upstream processing does not retain all task outputs" (it's not the tasks that are retaining or not).

python/lsst/pipe/base/quantum_graph_builder.py
```diff
@@ -722,41 +733,64 @@ def _skip_quantum_if_metadata_exists(
...
-        causes the quantum node to be removed from the graph.  Dataset nodes
+        For tasks not listed in `ignore_metadata_for`, a quantum is skipped
+        when its metadata dataset exists in ``skip_existing_in``.
+
+        For tasks listed in `ignore_metadata_for`, the metadata dataset is
+        not used as the completion signal.  Instead, a quantum is skipped only
+        when all of its science outputs are present in ``skip_existing_in``.
+        If any such output is absent the quantum is not skipped, so the task
+        can regenerate the missing outputs.  This supports pipelines where
+        upstream tasks do not retain all of their outputs.
```
TallJimbo: Reword as in the other thread.

python/lsst/pipe/base/quantum_graph_builder.py
```diff
@@ -188,6 +198,7 @@ def __init__(
         self.butler = butler.clone(collections=input_collections)
         self.output_run = output_run
         self.skip_existing_in = skip_existing_in
+        self.ignore_metadata_for: frozenset[str] = frozenset(ignore_metadata_for)
```
TallJimbo: We probably want to raise here if `ignore_metadata_for` is not empty but `skip_existing_in`  empty (since then `ignore_metadata_for` would have no effect).

TallJimbo (approved):

python/lsst/pipe/base/quantum_graph/aggregator/_writer.py
hsinfang: I took it from mypy's suggestion directly, but this has now been fixed by https://github.com/lsst/pipe_base/commit/ac0bcb1872388a8005a4d26097dc0fb0bb6dc524

tests/test_graphBuilder.py
hsinfang: Thanks for the suggestion. The new tests are changed to use `InMemoryRepo` instead.

timj:
@hsinfang are you imminently merging this or is it going to be a few days? I am trying to sync with the v30 release so wondered what your plan was.

hsinfang:
@timj  this won't be imminent, and can wait longer too if that makes other things easier.

python/lsst/pipe/base/quantum_graph_skeleton.py
```diff
@@ -308,6 +308,21 @@ def iter_inputs_of(
...
+        dataset_key : `DatasetKey`
+            Dataset to look up producers for.
+
+        Returns
+        -------
+        quanta : `~collections.abc.Iterator` of `QuantumKey` or `TaskInitKey`
+            Quanta that produce the given dataset.
+        """
+        return self._xgraph.predecessors(dataset_key)
+
```
TallJimbo, line 326: It shouldn't be possible for more than one quantum to produce a dataset, so this method could be a `get_producer_of` that returns a `QuantumKey | TaskInitKey | None`.

TallJimbo (approved):

python/lsst/pipe/base/quantum_graph_skeleton.py
hsinfang: Good point. Changed. Thank you!

PR #565, opened 2026-06-17
DM-55232: fix zstd failure in aggregate-graph due to huge logs

TallJimbo:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/pipe/base/quantum_graph/aggregator/_writer.py
+++ b/python/lsst/pipe/base/quantum_graph/aggregator/_writer.py
@@ -171,15 +171,34 @@ class Writer:
         # sample from just the first N, since they're ordered by UUID.  We
         # chop out the datastore records since those don't appear in the
         # provenance graph.
+        n_pred_quanta: int = 0
         for predicted_quantum in self.predicted.quantum_datasets.values():
             if len(training_inputs) == self.comms.config.zstd_dict_n_inputs:
                 break
             predicted_quantum.datastore_records.clear()
             training_inputs.append(predicted_quantum.model_dump_json().encode())
+            n_pred_quanta += 1
         # Add the provenance quanta, metadata, and logs we've accumulated.
         for write_request in self.pending_compression_training:
             assert not write_request.is_compressed, "We can't compress without the compression dictionary."
             training_inputs.append(write_request.quantum)
             training_inputs.append(write_request.metadata)
             training_inputs.append(write_request.logs)
-        return zstandard.train_dictionary(self.comms.config.zstd_dict_size, training_inputs)
+        try:
+            return zstandard.train_dictionary(self.comms.config.zstd_dict_size, training_inputs)
+        except zstandard.ZstdError as err:
+            n_write_requests = len(self.pending_compression_training)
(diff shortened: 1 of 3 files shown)


python/lsst/pipe/base/quantum_graph/aggregator/_writer.py
```diff
@@ -171,15 +171,52 @@ def make_compression_dictionary(self) -> zstandard.ZstdCompressionDict:
...
         # chop out the datastore records since those don't appear in the
         # provenance graph.
+        n_pred_quanta: int = 0
+        training_sample_size: int = 0
         for predicted_quantum in self.predicted.quantum_datasets.values():
             if len(training_inputs) == self.comms.config.zstd_dict_n_inputs:
                 break
+            if training_sample_size >= self.comms.config.zstd_dict_input_max_bytes:
+                self.comms.log.warning(
+                    "Truncating compression dict training sample after %d predicted quanta.",
```
MichelleGower: I don't know if it would be quicker to understand if instead of saying "Truncating" saying something like "Reached compression dict training sample limit after..."

MichelleGower: Also, do you want a similar warning with the above break for counts?

python/lsst/pipe/base/quantum_graph/aggregator/_writer.py
```diff
@@ -171,15 +171,52 @@ def make_compression_dictionary(self) -> zstandard.ZstdCompressionDict:
...
-        return zstandard.train_dictionary(self.comms.config.zstd_dict_size, training_inputs)
+            training_sample_size += len(write_request.logs)
+        try:
+            return zstandard.train_dictionary(self.comms.config.zstd_dict_size, training_inputs)
+        except zstandard.ZstdError as err:
+            n_write_requests = len(self.pending_compression_training)
+            pred_quanta_sum = sum(len(block) for block in training_inputs[:n_pred_quanta])
+            prov_quanta_sum = sum(len(block) for block in training_inputs[n_pred_quanta::3])
+            metadata_sum = sum(len(block) for block in training_inputs[n_pred_quanta + 1 :: 3])
+            logs_sum = sum(len(block) for block in training_inputs[n_pred_quanta + 2 :: 3])
```
MichelleGower, line 213: I possibly would set the step (3) in a variable to make sure future changes get them all.  But the lines are all here together so...

If it is easy to write a unit test to catch if we add/remove something to/from the training_inputs but forget to change the step, that could avoid trying to track down a bug in the exception handling code.

MichelleGower (approved):
A couple comments.  Merge approved.

python/lsst/pipe/base/quantum_graph/aggregator/_writer.py
TallJimbo: I've also reworked the logic break before exceeding the limit, rather than just after.

python/lsst/pipe/base/quantum_graph/aggregator/_writer.py
TallJimbo: This case should be rare enough and the test looks tricky enough that I'm going to skip it.

PR #569, opened 2026-07-02
DM-55320: Add a feature to prune unanchored quanta in qgraph building

hsinfang:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/pipe/base/quantum_graph_builder.py
+++ b/python/lsst/pipe/base/quantum_graph_builder.py
@@ -138,6 +138,12 @@ class QuantumGraphBuilder(ABC):
         the upstream quanta that need to regenerate those intermediates to also
         run.  Has no effect without ``skip_existing_in``.  ``["*"]`` means
         retaining all datasets, equivalent to not providing this option.
+    prune_unanchored_quanta : `tuple` [ `str`, `str` ], optional
+        A ``(source_label, anchor_label)`` pair of task labels triggering
+        unanchored quanta pruning after the skeleton is assembled.  A
+        ``source_label`` quantum is removed along with its entire downstream
+        chain if no ``anchor_label`` quantum is reachable from it along
+        directed graph edges.
     clobber : `bool`, optional
         Whether to raise if predicted outputs already exist in ``output_run``
         (not including those quanta that would be skipped because they've
@@ -182,6 +188,7 @@ class QuantumGraphBuilder(ABC):
         output_run: str | None = None,
         skip_existing_in: Sequence[str] = (),
         retained_dataset_types: Sequence[str] | None = None,
+        prune_unanchored_quanta: tuple[str, str] | None = None,
         clobber: bool = False,
     ):
         self.log = getLogger(__name__)
@@ -204,6 +211,7 @@ class QuantumGraphBuilder(ABC):
         )
         if self._retained_dataset_type_patterns is not None and not skip_existing_in:
             raise ValueError("retained_dataset_types has no effect without skip_existing_in.")
+        self._prune_unanchored_quanta = prune_unanchored_quanta
         self.empty_data_id = DataCoordinate.make_empty(butler.dimensions)
         self.clobber = clobber
         # See whether the output run already exists.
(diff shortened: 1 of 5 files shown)


python/lsst/pipe/base/quantum_graph_skeleton.py
```diff
@@ -562,6 +562,67 @@ def remove_orphan_datasets(self) -> None:
             if not orphan.is_task and orphan not in self._global_init_outputs:
                 self._xgraph.remove_node(orphan)
 
+    def remove_unanchored_quanta(self, source_label: str, anchor_label: str) -> tuple[int, int]:
+        """Remove unanchored source quanta and their entire downstream chain.
+
+        A source quantum is considered unanchored if no quantum with
+        ``anchor_label`` is reachable along directed edges from it.  Unanchored
+        source quanta and every descendant reachable from them are removed.
```
TallJimbo: The second sentence here is redundant with the next paragraph.

python/lsst/pipe/base/quantum_graph_skeleton.py
```diff
@@ -562,6 +562,67 @@ def remove_orphan_datasets(self) -> None:
...
+
+        Any descendant of an unanchored source quantum is removed
+        unconditionally, regardless of what else may reach it. If downstream
+        subgraphs of distinct source quanta share nodes, the nodes are removed
+        as well.
```
TallJimbo: I don't think the second sentence here is adding any value; I think it's just saying that it's okay if multiple unanchored source quantum have a common downstream node, but there's nothing else we might plausibly do with such a node besides delete it.

python/lsst/pipe/base/quantum_graph_builder.py
```diff
@@ -470,6 +493,19 @@ def _build_skeleton(self, attach_datastore_records: bool = True) -> QuantumGraph
...
+                n_source, n_downstream = full_skeleton.remove_unanchored_quanta(source_label, anchor_label)
+                if n_source:
+                    self.log.info(
+                        "Pruned %d unanchored %r quanta and %d downstream quanta (%d total) based on %r.",
+                        n_source,
+                        source_label,
+                        n_downstream,
+                        n_source + n_downstream,
+                        anchor_label,
+                    )
```
TallJimbo: I think we'd appreciate it in the future if this log message turned into one log message per task, given that existing logging about skipping and pruning reports per-task counts.  Maybe just have `remove_unanchored_quanta` return a dict of `task_label: n_removed`?

TallJimbo (approved):
Looks good; only minor comments.

Can you confirm my understanding of how this relates to our earlier discussion on this?

- When we talking about this at pair-programming, I was thinking that we would have to iterate between this and the existing skip/prune logic to ensure a self-consistent graph that satisfies all `adjustQuantum` hooks.

- That's not actually needed because the logic here just removes _all_ downstream quanta entirely, and hence remaining quanta's `adjustQuantum` inputs would be unaffected.

- There is a hypothetical alternate version of this feature where you just prune the unanchored source quantum, and then let `adjustQuantum` pruning act more selectively downstream.  But what's here is actually a better fit for your use case.

- We would also need some kind of iteration if we accepted multiple (anchor, source) pairs, because those might interact.  But you don't need that either.

Finally, one thing just to be aware of: like the existing pruning, this happens *after* the dimension records for the QG are fetched, and hence the QG will end up storing dimension records even for pruned quanta.  This is usually a minor storage/memory inefficiency because the pruned fraction is small (and pruning the dimension records at the end is a denormalization problem that's usually a waste of time).  But we've hit pathological QG builds in the past where the original graph is enormous and the pruned graph is small, and that could happen here too.  But the best fix for this case it to constrain the original QG build better (this can really only happen with `--dataset-query-constraint off` and a weak `--data-query`).

hsinfang:
I agree with all you said. My recollection from our pair-programming (though that was before I looked at `adjustQuantum`) was that we were thinking whether the pruning could be at an earlier stage of graph building. But `_adjust_task_quanta` is done one task at a time in topological order. The anchor task `associateApdb` needs to settle first, and only after that source-anchor checking makes sense. 

Accepting multiple (anchor, source) pairs may still be a thing we want to do, but I feel reluctant to implement it before we have a concrete use case.  Whether iterations are needed may depend on what kinds of task pairs we have in mind.

python/lsst/pipe/base/quantum_graph_skeleton.py
hsinfang: Originally I wanted to say something like "this feature is with the AP use case in mind where no source-anchor subgraphs overlap with each other" and but that seems weird to say here.   I will just remove the second sentence.

hsinfang: Removed the second sentence and combined the first sentence to the previous paragraph.

python/lsst/pipe/base/quantum_graph_skeleton.py
hsinfang: Mildly re-written and the next paragraph is removed.

python/lsst/pipe/base/quantum_graph_builder.py
hsinfang: Good idea. Changed.

PR #574, opened 2026-07-20
DM-55521: Allow regular mutable mappings for metadata failure annnotations.

TallJimbo:
This is motivated by lsst.images types, which have a regular dict for their metadata instead of a PropertySet-like.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/pipe/base/_status.py
+++ b/python/lsst/pipe/base/_status.py
@@ -44,6 +44,7 @@ import abc
 import enum
 import logging
 import sys
+from collections.abc import MutableMapping
 from typing import TYPE_CHECKING, Any, ClassVar, Protocol
 
 import pydantic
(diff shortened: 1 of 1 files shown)


python/lsst/pipe/base/_status.py
```diff
@@ -548,7 +550,10 @@ def runQuantum(self, butlerQC, inputRefs, outputRefs):
             # Some outputs may not exist, so we cannot set metadata on them.
             if item is None:
                 continue
-            item.metadata.set_dict("failure", failure_info)  # type: ignore
+            if hasattr(item.metadata, "set_dict"):
+                item.metadata.set_dict("failure", failure_info)  # type: ignore
+            elif isinstance(item.metadata, MutableMapping):
+                item.metadata["failure"] = failure_info
```
timj, line 556: Are we asserting that we do not need an `else` because mypy knows we've tested all the variants?

timj (approved):

python/lsst/pipe/base/_status.py
TallJimbo: Good catch; I think this probably is worth an assert since a lot of the calling code is not type-checked.

PR #578, opened 2026-08-12
DM-55525: Improve guards on compression-dict failures.

TallJimbo:
Mostly AI-authored, with human rewritten docstrings, comments, and log messages.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/pipe/base/quantum_graph/aggregator/_writer.py
+++ b/python/lsst/pipe/base/quantum_graph/aggregator/_writer.py
@@ -178,7 +178,6 @@ class Writer:
                 break
             predicted_quantum.datastore_records.clear()
             block = predicted_quantum.model_dump_json().encode()
-            training_sample_size += len(block)
             if training_sample_size >= self.comms.config.zstd_dict_input_max_bytes:
                 self.comms.log.warning(
                     "Reached compression dict training sample size limit at %d predicted quanta.",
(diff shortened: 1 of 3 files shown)


python/lsst/pipe/base/quantum_graph/aggregator/_writer.py
```diff
@@ -203,6 +202,17 @@ def make_compression_dictionary(self) -> zstandard.ZstdCompressionDict:
             training_inputs.append(write_request.quantum)
             training_inputs.append(write_request.metadata)
             training_inputs.append(write_request.logs)
+        if len(training_inputs) < 7:
+            # zstandard raises a totally opaque exception when there aren't
+            # enough training samples or their total size is too big.  An
+            # AI-assisted scan of the source code says it needs >=7 samples
+            # and < 4GB total.
```
MichelleGower: Did you happen to test exactly 7?  When I was searched Google for a  constant we could use instead of hardcoding this, it said one doesn't exist and recommended adding a "< 8" check.

tests/test_aggregator.py
```diff
@@ -1379,6 +1381,66 @@ def test_bad_metadata_readable(self) -> None:
         self.assertTrue(np.isnan(prov_md.attempts[0]["calibrateImage:psf_measure_psf"]["spatialFitChi2"]))
 
 
+class CompressionDictTestCase(unittest.TestCase):
+    """Test `Writer.make_compression_dictionary` guards against zstandard's
+    undocumented requirments.
```
MichelleGower: Typo: requirments.

MichelleGower (approved):
Looks good.  A typo and a question.  Merge approved

python/lsst/pipe/base/quantum_graph/aggregator/_writer.py
TallJimbo: ~The analysis I ran said that <8 would work if and only if we also relaxed some other parameter from the defaults, and I didn't think that was worth the complexity.~

Oops, that gets the sign wrong, and clearly I'm not remembering that analysis very well (it was a couple of weeks ago now).  I'm just going to set it to 10 to be safe.

PR #579, opened 2026-08-13
DM-55803 : Provenance Report should support JSON Output

tcjennings:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,7 +23,7 @@ classifiers = [
 ]
 keywords=["lsst"]
 dependencies = [
-    "lsst-resources [s3]",
+    "lsst-resources[s3]",
     "lsst-utils",
     "lsst-daf-butler",
     "lsst-pex-config",
@@ -39,6 +39,19 @@ dependencies = [
 
 dynamic = ["version"]
 
+[dependency-groups]
+dev = [
+    "documenteer[guide] >2.0,<3.0",
+    "lsst-sphinxutils",
+    "mypy",
+    "prek",
+    "pytest",
+    "pytest-xdist",
+    "pytest-cov",
+    "types-PyYAML",
+    "types-Deprecated",
+]
+
 [project.entry-points.'butler.cli']
 pipe_base = "lsst.pipe.base.cli:get_cli_subcommands"
 
@@ -50,6 +63,13 @@ pipe_base = "lsst.pipe.base.cli:get_cli_subcommands"
 pandas = ["pandas"]
 test = ["pytest >= 3.2"]
 mermaid = ["mermaid-py >= 0.7.1"]
+
+[tool.uv.sources]
+lsst-resources = { git = "https://github.com/lsst/resources", branch = "main" }
+lsst-utils = { git = "https://github.com/lsst/utils", branch = "main" }
+lsst-daf-butler = { git = "https://github.com/lsst/daf_butler", branch = "main" }
+lsst-pex-config = { git = "https://github.com/lsst/pex_config", branch = "main" }
+lsst-sphinxutils = { git = "https://github.com/lsst/sphinxutils", branch = "main" }
 
 [tool.setuptools.packages.find]
 where = ["python"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,6 +47,7 @@ pipe_base = "lsst.pipe.base.cli:get_cli_subcommands"
 "Source" = "https://github.com/lsst/pipe_base"
 
 [project.optional-dependencies]
+pandas = ["pandas"]
 test = ["pytest >= 3.2"]
 mermaid = ["mermaid-py >= 0.7.1"]
 
--- a/python/lsst/pipe/base/quantum_graph/_provenance.py
+++ b/python/lsst/pipe/base/quantum_graph/_provenance.py
@@ -1309,7 +1309,7 @@ class ProvenanceQuantumGraph(BaseQuantumGraph):
         return self._bipartite_xgraph.copy(as_view=True)
 
     def make_quantum_table(
-        self, drop_unused_columns: bool = True, expand_caveats: bool = False, as_table: bool = True
+        self, *, drop_unused_columns: bool = True, expand_caveats: bool = False, as_table: bool = True
     ) -> astropy.table.Table | list[dict]:
         """Construct an `astropy.table.Table` with a tabular summary of the
         quanta.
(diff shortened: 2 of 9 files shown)


pyproject.toml
```diff
@@ -23,11 +23,12 @@ classifiers = [
 ]
 keywords=["lsst"]
 dependencies = [
-    "lsst-resources [s3]",
+    "lsst-resources[s3]",
     "lsst-utils",
     "lsst-daf-butler",
     "lsst-pex-config",
     "astropy",
+    "pandas",
```
timj: We don't want middleware to depend on pandas. Is there an alternative?

tcjennings: Yes, it'll be a little more fiddly but I can get there.

pyproject.toml
```diff
@@ -39,6 +39,19 @@ dependencies = [
 
 dynamic = ["version"]
 
+[dependency-groups]
+dev = [
+    "documenteer[guide] >2.0,<3.0",
```
TallJimbo, line 44: Please check that this is the right documenteer variant for `package-docs` (or maybe @timj knows off the top of his head).  I'm not sure if package docs are different from user-guide docs or not.

pyproject.toml
```diff
@@ -51,6 +64,13 @@ pandas = ["pandas"]
 test = ["pytest >= 3.2"]
 mermaid = ["mermaid-py >= 0.7.1"]
 
+[tool.uv.sources]
+lsst-resources = { git = "https://github.com/lsst/resources", branch = "main" }
+lsst-utils = { git = "https://github.com/lsst/utils", branch = "main" }
+lsst-daf-butler = { git = "https://github.com/lsst/daf_butler", branch = "main" }
+lsst-pex-config = { git = "https://github.com/lsst/pex_config", branch = "main" }
+lsst-sphinxutils = { git = "https://github.com/lsst/sphinxutils", branch = "main" }
```
TallJimbo, line 72: I'm a little surprised `uv` doesn't just read the regular dependencies.  I thought it was mostly a drop-in `pip` replacement.

This is more curiosity than objection, unless this means that we'd need an override in two places to get GitHub Actions to build against a branch of a dependency.

python/lsst/pipe/base/quantum_graph/_provenance.py
```diff
@@ -1307,7 +1308,9 @@ def bipartite_xgraph(self) -> networkx.DiGraph:
         """
         return self._bipartite_xgraph.copy(as_view=True)
 
-    def make_quantum_table(self, drop_unused_columns: bool = True) -> astropy.table.Table:
+    def make_quantum_table(
+        self, drop_unused_columns: bool = True, expand_caveats: bool = False, as_table: bool = True
```
TallJimbo: Let's make all of these keyword-only to force more readable calls (including the current one; I don't think there's any backwards-compatibility issue with that in practice).

python/lsst/pipe/base/quantum_graph/_provenance.py
```diff
@@ -1365,23 +1380,31 @@ def make_quantum_table(self, drop_unused_columns: bool = True) -> astropy.table.
...
+        Parameters
+        ----------
+        as_table : `bool`, optional
+            Whether to return an `astropy.table.Table` or a list of rows.
+
         Returns
         -------
-        table : `astropy.table.Table`
-            A table with columns for task label, exception type, and counts.
+        table : `astropy.table.Table` | `list[dict]`
```
TallJimbo: I think it needs to be spelled
```
`list` [`dict`]
```
to get Sphinx to parse it properly.

But you're also welcome to drop the `Returns` section and let the Sphinx type-annotation plugin handle it by moving the actual information up to the main docstring section or notes (unfortunately we can't get the type-annotation sphinx plugin to fire for returns any other way).

TallJimbo (changes requested):
Looks good; mostly only minor comments.  But please put this on a new PR with a `tickets/DM-55803` branch and actually merge from that.

pyproject.toml
tcjennings: That's how it's set in `docs/requirements.txt` but I'm happy to change it there too if there's a better variant.

TallJimbo: Ah, probably just something correct that I was unfamiliar with, then.

pyproject.toml
tcjennings: As a `uv` enthusiast, it is much more than a `pip` replacement but I'm not trying to boil the ocean here (the github actions are a MESS). This is more directing towards a smooth devx where one can run `uv sync` and have a fully-hydrated venv out of the gate instead of having to source dependencies from multiple places. The GHA build mechanics are unchanged and you only need to change, e.g., the `requirements.txt` to affect that.

python/lsst/pipe/base/quantum_graph/_provenance.py
tcjennings: Good point. This is a reminder of why linters are good but can be a crutch. The `FBT` ruff ruleset would have caught this.

pyproject.toml
timj: I think @TallJimbo is saying that we currently edit requirements.txt branches to let the build find other dev branches in PRs and we have CI gates to make sure such edits aren't merged to main. He is asking if we would now need to edit two branch specifiers (one in requirements.txt and one in pyproject.toml) now? Would this let us remove requirements.txt (but mypy action is currently requiring that exists).

tcjennings: If we go all-in on `uv` we can get rid of `requirements.txt`, `types.txt`, `docs/requirements.txt`, (and other incidental cruft), simplify the GH actions and make `pyproject.toml` the only source of truth.

The VCS details from `requirements.txt` aren't declared in the published package anyway, and keeping them separate in `pyproject.toml` via `[uv.tool.sources]` keeps it that way. (i.e., there's a distinction between "I need this dependency checkout/ref/commit when installing" and "You need this dependency to install my package")

tcjennings:
Closing in favor of a new PR with a corrected branch name.

pyproject.toml
timj: Yes, going all in on `uv` does simplify but we first need to make sure that things like the mypy action will work without a types.txt/requirements.txt file and we should probably start them migrating bottom up in the middleware stack rather than in the middle. I assume "pip install" still works and you don't require "uv pip install" (we have non-Rubin users of this package).

tcjennings: Right, adding plumbing for `uv` at this level without removing those other files doesn't change how anyone using `pip` or `uv pip` interacts with the package, whether they `git clone` it or install it from pypi. I'll take a look at the shared mypy action and see what it needs.

tcjennings:
> Looks good; mostly only minor comments. But please put this on a new PR with a `tickets/DM-55803` branch and actually merge from that.

See: https://github.com/lsst/pipe_base/pull/581

PR #583, opened 2026-08-19
Revert VCS Sources from pyproject.toml

tcjennings:
fix: redact `tool.uv.sources` from pyproject.toml


timj (approved):
Thanks. Can you add a note as to why in the commit message?

